PR #2, opened 2025-08-04
Add some JIT filters to the methods

solangegbv:
in pcs.py and planar_pcs.py to improve performance.

+Fix one debugging comment in pcs.py to use vmap correctly.


mstoelzle (approved):
Looks good

PR #4, opened 2025-08-05
Change the default reference configurations

solangegbv:
to align every rod with the x-axis by default.

This change includes:
- Adjusting the default reference configurations
- Changing the documentation to reflect the new default behavior
- Changing the default stiffness matrix to match the new reference configurations

All previous "equilibrium" and "rest" are now called "reference".


src/soromox/symbolic_derivation/planar_hsa.py
```diff
@@ -544,9 +544,9 @@ def symbolically_derive_planar_hsa_model(
             "rout": rout_syms,
             "rin": rin_syms,
             "roff": roff_syms,
-            "kappa_b_eq": kappa_b_eq_syms,
```
mstoelzle, line 547: If you do these changes to the symbolic derivation, you will also have to generate the symbolic expressions and push them in this PR.

tests/test_planar_pcs_num.py
```diff
@@ -72,7 +72,7 @@ def test_planar_cs_num(
     # activate all strains (i.e. bending, shear, and axial)
     strain_selector = jnp.ones((3,), dtype=bool)
 
-    xi_eq = jnp.array([0.0, 0.0, 1.0])
+    xi_ref = jnp.array([0.0, 0.0, 1.0])
```
mstoelzle: This `xi_ref` should probably be changed to `jnp.array([0.0, 1.0, 0.0])`?

src/soromox/systems/planar_pcs.py
```diff
@@ -344,46 +355,101 @@ def update_params(self, params: Dict[str, Array]) -> "PlanarPCS":
...
+                - "D": List/Array of (num_segments x num_segments) floats
+                    Damping matrix of each segment [Pa*s]
         """
         # Apply updates sequentially
         updated_self = self
         
         if "th0" in params:
-            updated_self = eqx.tree_at(lambda m: m.th0, updated_self, jnp.asarray(params["th0"], dtype=jnp.float64))
+            th0 = params["th0"]
+            if not (isinstance(th0, (float, int, jnp.ndarray))):
```
mstoelzle, line 381: Having conditions in the `update_params` method is potentially problematic, as this function is designed to be used in jitted settings. We can try though.

examples/simulate_planar_pcs.py
```diff
@@ -153,7 +152,7 @@ def update_plot(frame_idx):
         "L": 1e-1 * jnp.ones((num_segments,)),
         "r": 2e-2 * jnp.ones((num_segments,)),
         "rho": rho,
-        "g": jnp.array([0.0, -9.81]),
+        "g": jnp.array([0.0, 9.81]), # gravity vector [m/s^2] UP!
```
mstoelzle, line 155: So I assume that here gravity is pointing upwards?

src/soromox/symbolic_expressions/pendulum_nl-1.dill
mstoelzle, line 1: We need to keep the pendulum symbolic expressions I think, no? Or are you planning to also do a numeric implementation of the dynamics there?

src/soromox/symbolic_expressions/planar_hsa_ns-1_nrs-2.dill
mstoelzle, line 1: We need to keep the planar HSA symbolic expressions I think, no? Or are you planning to also do a numeric implementation of the dynamics there?

mstoelzle (changes requested):

examples/simulate_planar_pcs.py
solangegbv: Actually I think it was the case on your previous examples so I putted that same way. I'm not sure why, but I just imitated your previous example.

src/soromox/symbolic_expressions/pendulum_nl-1.dill
solangegbv: No, for the moment I'm keeping them, I'll see if I eventually have the time.

src/soromox/symbolic_expressions/planar_hsa_ns-1_nrs-2.dill
solangegbv: No, for the moment I'm keeping them, I'll see if I eventually have the time.

mstoelzle (approved):

PR #6, opened 2025-08-08
Class for HSA

solangegbv:
Changed Planar HSA system to add class structure


examples/simulate_pneumatically_actuated_planar_pcs.py
```diff
@@ -30,26 +30,31 @@
     "l": 1e-1 * jnp.ones((num_segments,)),
```
mstoelzle: The `th0` should be set here to `jnp.pi / 2` as well, right? Can you please double-check for all examples?

examples/simulate_pneumatically_actuated_planar_pcs.py
```diff
@@ -30,26 +30,31 @@
...
     "G": 1e3 * jnp.ones((num_segments,)),  # Shear modulus [Pa]
     "r_cham_in": 5e-3 * jnp.ones((num_segments,)),
     "r_cham_out": 2e-2 - 2e-3 * jnp.ones((num_segments,)),
-    "varphi_cham": jnp.pi/2 * jnp.ones((num_segments,)),
+    "varphi_cham": jnp.pi / 2 * jnp.ones((num_segments,)),
 }
 params["D"] = 5e-4 * jnp.diag(
-    (jnp.repeat(
-        jnp.array([[1e0, 1e3, 1e3]]), num_segments, axis=0
-    ) * params["l"][:, None]).flatten()
```
mstoelzle, line 43: Please update to `params["L"]`

examples/simulate_planar_hsa.py
```diff
@@ -1,128 +1,399 @@
...
+    linewidth=jnp.inf,
+    formatter={"float_kind": lambda x: "0" if x == 0 else f"{x:.2e}"},
 )
 
-# activate all strains (i.e. bending, shear, and axial)
-strain_selector = jnp.ones((3 * num_segments,), dtype=bool)
-consider_hysteresis = True
 
-params = PARAMS_FPU_HYSTERESIS_CONTROL if consider_hysteresis else PARAMS_FPU_CONTROL
+def draw_robot(
```
mstoelzle: Please migrate this example to the `draw_robot` function that is included here: https://github.com/tud-phi/soromox/blob/main/src/soromox/rendering/planar_hsa/opencv_renderer.py

src/soromox/systems/planar_hsa.py
```diff
@@ -10,237 +10,717 @@
...
+                    for param in params_vals.flatten():
+                        params_for_lambdify.append(param)
+        except KeyError:
+            return KeyError(
+                f"Symbolic expressions file does not contain the required parameters. Please generate the symbolic expressions first."
+            )
+        self.params_for_lambdify = params_for_lambdify
+
+        try:
+            L = params["l"]
```
mstoelzle: This should probably be changed to `params["L"]` for consistency

src/soromox/systems/planar_hsa.py
```diff
@@ -10,237 +10,717 @@
...
+        self, params: Dict[str, Array], consider_hysteresis: bool, num_dofs: int
+    ) -> None:
         """
-        Select the parameters for lambdify
+        Set the parameters of the PlanarHSA model.
+
         Args:
-            params: Dictionary of robot parameters
+            params (Dict[str, Array]):
+                Dictionary containing the robot parameters to update:
```
mstoelzle, line 381: Can you please resolve these TODO's? If the meaning of some of them is unclear, I can help out.

src/soromox/systems/planar_hsa.py
```diff
@@ -10,237 +10,717 @@
...
+        Args:
+            q: generalized coordinates of shape (num_dofs, )
+        Returns:
+            xi: strains of the virtual backbone of shape (num_dofs, )
+        """
+        # rest strains of the virtual backbone
+        xi_star = self.rest_strains_fn()
+
+        # map the configuration to the strains
+        xi = self.B_xi @ q + xi_star
```
mstoelzle: Probably `xi_star` should be renamed to `xi_ref`?

src/soromox/systems/planar_hsa.py
```diff
@@ -10,237 +10,717 @@
...
-
-        # map the configuration to the strains
-        xi = xi_ref + B_xi @ q
-
-        return xi
-
-    @jit
-    def apply_eps_to_bend_strains_fn(xi: Array, _eps: float) -> Array:
+        # rest strains of the physical rods
+        pxi_star = jnp.zeros((self.num_segments, self.num_rods_per_segment, 3))
```
mstoelzle: I think I saw something similar also in the `PlanarPCS` and the `PCS` system. I think we should actually change this as we saw in the past that these kind of `.at[]` in-place changes are quite slow. Instead, we can replace it with code that is based on `jnp.repeat` (https://docs.jax.dev/en/latest/_autosummary/jax.numpy.repeat.html). Please let me know if you have any questions about it and I can help out.

src/soromox/systems/planar_hsa.py
```diff
@@ -264,223 +744,180 @@ def apply_eps_to_bend_strains_fn(xi: Array, _eps: float) -> Array:
...
         # add a small number to the bending strain to avoid singularities
-        xi_epsed = apply_eps_to_bend_strains_fn(xi, eps)
-
-        # convert the dictionary of parameters to a list, which we can pass to the lambda function
-        params_for_lambdify = select_params_for_lambdify_fn(params)
+        xi_epsed = self.apply_eps_to_bend_strains_fn(xi)
 
         chip = lax.switch(
-            segment_idx, chip_lambda_sms, *params_for_lambdify, *xi_epsed
+            segment_idx, self.chip_lambda_sms, *self.params_for_lambdify, *xi_epsed
```
mstoelzle, line 876: So did you switch around the order from `[p_x, p_y, theta]` to `[theta, p_x, p_y]` in the symbolic functions? Or where is this switch happening? Please add a code comment about this.

src/soromox/systems/planar_hsa.py
```diff
@@ -544,120 +981,304 @@ def inverse_kinematics_end_effector_fn(
...
+            K_full (Array): Full stiffness vector of shape (num_dofs_max, ).
+        """
+        xi = self.strain(q)
+
+        K_full = self.K_lambda(*self.params_for_lambdify, *xi).squeeze()
+
+        return K_full
+
+    @eqx.filter_jit
+    def stiffness_vector(self, q: Array) -> Array:
```
mstoelzle: This should be `elastic_force`, right?

src/soromox/systems/planar_hsa.py
```diff
@@ -544,120 +981,304 @@ def inverse_kinematics_end_effector_fn(
...
-        params: Dict[str, Array],
-        q: Array,
-        qd: Array,
-        B: Array,
-        C: Array,
-        eps: float = 1e4 * global_eps,
+    @eqx.filter_jit
+    def Shat(self) -> Array:
+        """
+        TODO
```
mstoelzle: `Shat` is the nominal stiffness of the system. It should return a stiffness matrix of shape `(ndof, ndof)`

src/soromox/systems/planar_hsa.py
```diff
@@ -671,123 +1292,161 @@ def operational_space_dynamical_matrices_fn(
...
+
         else:
-            B, C, G, K, D, _ = dynamical_matrices_fn(
-                params, q, qd, z=z, phi=jnp.zeros((num_rods,))
-            )
-            tau_q = u
+            B = self.inertia_matrix(q)
+            C = self.coriolis_matrix(q, qd)
+            G = self.gravitational_force(q)
+            K = self.stiffness_vector(q)
```
mstoelzle: I think this should be renamed to `tau_el` instead of `K`

src/soromox/systems/planar_hsa.py
```diff
@@ -671,123 +1292,161 @@ def operational_space_dynamical_matrices_fn(
...
 
         if consider_underactuation_model is True:
             phi = u
-            B, C, G, K, D, tau_q = dynamical_matrices_fn(params, q, qd, z=z, phi=phi)
+            B = self.inertia_matrix(q)
+            C = self.coriolis_matrix(q, qd)
+            G = self.gravitational_force(q)
+            K = self.stiffness_vector(q)
+            D = self.damping_matrix()
+            alpha = self.actuation_matrix(q, phi)
```
mstoelzle: Two things:
- Please double check that the `self.actuation_matrix` method actually exists.
- Please rename it to `self.actuation_force` as it does not return an actuation matrix, but instead directly the actuation forces acting on the generalized coordinates.

mstoelzle (changes requested):

examples/simulate_pneumatically_actuated_planar_pcs.py
solangegbv: For the moment pneumatically and tendon actuated systems are still like the old system and not changed at all. I will wait until I've made them classes to bring the modification and not get lost if that's okay with you.

src/soromox/systems/planar_hsa.py
solangegbv: No, for the moment the switch is done only on the numerical functions using jnp.roll(..., 1)

solangegbv: Each definition has the comment "chi...: 
                Consists of [theta, p_x, p_y]
                where theta is the planar orientation with respect to the x-axis,
                p_x is the x-position, p_y is the y-position,"

src/soromox/systems/planar_hsa.py
solangegbv: I don't understand your first point

src/soromox/systems/planar_hsa.py
mstoelzle: I see! Please add a code comment about the `jnp.roll(..., 1)` that this is the function that maps from the convention used in the symbolic functions `[p_x, p_y, theta]` to `[theta, p_x, p_y]`

examples/simulate_pneumatically_actuated_planar_pcs.py
mstoelzle: Sounds good!

src/soromox/systems/planar_hsa.py
mstoelzle: Please disregard my first point. I just didn't see it on the PR initially but it is indeed here!

src/soromox/systems/factory_pneumatically_actuated_planar_pcs.py
mstoelzle, line 1: Are you planning to delete this file later?

src/soromox/systems/planar_pcs_sym.py
mstoelzle, line 1: I think this file can be deleted, right?

mstoelzle (changes requested):

src/soromox/systems/factory_pneumatically_actuated_planar_pcs.py
solangegbv: Yes, I'm just using it to compare while I'm building the (unfinished) class.

src/soromox/systems/planar_pcs_sym.py
solangegbv: Yes, same as the factory ones, I will delete it when every numerical classes will be done.

src/soromox/systems/pneumatically_actuated_planar_pcs.py
```diff
@@ -1,239 +1,415 @@
...
+from .planar_pcs import PlanarPCS
+
+
+class PneumaticallyActuatedPlanarPCS(PlanarPCS):
+    r_chamber_in: Array  # inner radius of each segment's chamber, shape (num_segments,)
+    r_chamber_out: (
+        Array  # outer radius of each segment's chamber, shape (num_segments,)
+    )
+    phi_chamber: Array  # sector angle of each segment's chamber, shape (num_segments,)
+    num_chambers: int = eqx.field(
```
mstoelzle: Small detail: Maybe this should be `num_chambers_per_segment`

src/soromox/systems/pneumatically_actuated_planar_pcs.py
```diff
@@ -1,239 +1,415 @@
...
+        if segment_actuation_selector is None:
+            segment_actuation_selector = jnp.ones(num_segments, dtype=bool)
+
+        # self.segment_indices_to_actuate = jnp.array(
+        #     [i for i, act in enumerate(segment_actuation_selector) if act]
+        # )
+
+        self.num_actuators = (
+            int(jnp.sum(segment_actuation_selector)) * 2
+        )  # each segment has two tendons
```
mstoelzle: The pneumatic soft robot has two tendons?

src/soromox/systems/pneumatically_actuated_planar_pcs.py
```diff
@@ -1,239 +1,415 @@
...
-        B_xi: Array,
-        xi_ref: Array,
-        q: Array,
-    ) -> Array:
+        order_gauss: int = 5,
+        strain_selector: Optional[Array] = None,
+        xi_ref: Optional[Array] = None,
+        segment_actuation_selector: Optional[Array] = None,
+        simplified_actuation_mapping: bool = False,
+    ):
```
mstoelzle, line 99: Can you add a docstring please?

mstoelzle (changes requested):

src/soromox/systems/pcs.py
```diff
@@ -1127,15 +1172,42 @@ def gravitational_force(self, q: Array) -> Array:
...
+    @eqx.filter_jit
+    def _local_stiffness_matrix(self, i: int) -> Array:
+        """
+        Compute the local stiffness matrix of a planar system for a rod aligned along the x-axis.
+
+        Args:
+            i (int): index of the segment
+
+        Returns:
+            S_i (Array): Local stiffness matrix of shape (3, 3) for the i-th segment.
```
mstoelzle: The dimension here is wrong

src/soromox/systems/pcs.py
```diff
@@ -1359,7 +1433,12 @@ def total_energy(self, q: Array, qd: Array) -> Array:
         return E
 
     @eqx.filter_jit
-    def operational_space_dynamical_matrices(self, q: Array, qd: Array, s: Array, operational_space_selector: Tuple = (True, True, True),
+    def operational_space_dynamical_matrices(
```
mstoelzle, line 1436: The operational space dynamical matrices function still has the bug about using s_local

mstoelzle (changes requested):

mstoelzle (approved):

PR #8, opened 2025-08-21
Implement tendon actuation in pcs

Michele-Martini:
This pull request introduces several new features and improvements to the tendon and pneumatic actuated continuum robot simulation codebase. The main highlights include the addition of a comprehensive simulation and visualization example for tendon-actuated robots, new actuation routing functions, and codebase consistency improvements for class variable definitions and typing. These changes enhance usability for interactive analysis, improve code maintainability, and add new capabilities for modeling tendon routing.

**New simulation and visualization example:**

* Added a new example script `examples/simulate_tendon_actuated_pcs.py` that demonstrates simulation and interactive visualization (with animation and slider options) for tendon-actuated continuum robots, including energy and trajectory plotting.

**Tendon actuation modeling:**

* Introduced new functions `linear_routing` and `linear_routing_derivative` in `src/soromox/actuation/tendon_actuation.py` to model linear tendon routing and its spatial derivative, supporting more flexible actuation modeling.

**Code consistency and maintainability:**

* Updated `global_eps` to be a `ClassVar` rather than a dataclass field in both `PCS` and `PlanarPCS` classes, preventing default/non-default ordering issues in subclasses and improving code clarity. [[1]](diffhunk://#diff-19b0b4a682b4e94d1538d4dd92005bf5c097e7cabb76cb4ddf8c61e55fc851c0L84-R86) [[2]](diffhunk://#diff-4397bec6ccb864d3bfb2d4a8e482b3312663630e95789bf9d8ada63ffae30b5cL81-R83)
* Standardized typing for segment index arguments in several methods of `PlanarPCS` to use `Array` instead of `int`, supporting JAX transformations and vectorization. [[1]](diffhunk://#diff-4397bec6ccb864d3bfb2d4a8e482b3312663630e95789bf9d8ada63ffae30b5cL518-R520) [[2]](diffhunk://#diff-4397bec6ccb864d3bfb2d4a8e482b3312663630e95789bf9d8ada63ffae30b5cL618-R620) [[3]](diffhunk://#diff-4397bec6ccb864d3bfb2d4a8e482b3312663630e95789bf9d8ada63ffae30b5cL936-R943) [[4]](diffhunk://#diff-4397bec6ccb864d3bfb2d4a8e482b3312663630e95789bf9d8ada63ffae30b5cL951-R958) [[5]](diffhunk://#diff-4397bec6ccb864d3bfb2d4a8e482b3312663630e95789bf9d8ada63ffae30b5cL966-R973) [[6]](diffhunk://#diff-4397bec6ccb864d3bfb2d4a8e482b3312663630e95789bf9d8ada63ffae30b5cL1170-R1177)

**General improvements:**

* Added `__all__` definitions to several modules (`pcs.py`, `planar_pcs.py`, `planar_hsa.py`, `pneumatic_actuated_planar_pcs.py`) for better module export control. [[1]](diffhunk://#diff-19b0b4a682b4e94d1538d4dd92005bf5c097e7cabb76cb4ddf8c61e55fc851c0R1-R20) [[2]](diffhunk://#diff-4397bec6ccb864d3bfb2d4a8e482b3312663630e95789bf9d8ada63ffae30b5cR1-R7) [[3]](diffhunk://#diff-e85995f693caba6f25e2e68ea7ab533b5f9b3079021d7fe97c506300dc7a3713R1)
* Refactored the `PneumaticActuatedPlanarPCS` class for clarity by cleaning up field definitions and constructor arguments, and ensuring proper handling of chamber parameters. [[1]](diffhunk://#diff-e85995f693caba6f25e2e68ea7ab533b5f9b3079021d7fe97c506300dc7a3713L79-L100) [[2]](diffhunk://#diff-e85995f693caba6f25e2e68ea7ab533b5f9b3079021d7fe97c506300dc7a3713L126-L128)


examples/simulate_tendon_actuated_pcs.py
```diff
@@ -0,0 +1,291 @@
...
+
+def draw_robot_curve(
+    batched_forward_kinematics: Callable,
+    L_max: float,
+    q: Array,
+    num_points: int = 50,
+):
+    s_ps = jnp.linspace(0, L_max, num_points)
+    g_ps = batched_forward_kinematics(q, s_ps)[:, :3, 3]
+
```
mstoelzle, line 36: @Michele-Martini Can you please visualize the tendons here for debugging purposes?

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,336 @@
...
+                    Density of each segment [kg/m^3]
+                - "g": List/Array of 2 floats [gx, gy]
+                    Gravitational acceleration vector [m/s^2]
+                - "E": List/Array of num_segments floats
+                    Elastic modulus of each segment [Pa]
+                - "G": List/Array of num_segments floats
+                    Shear modulus of each segment [Pa]
+                - "D": List/Array of (num_segments x num_segments) floats
+                    Damping matrix of each segment [Pa*s]
+                - "d": List/Array of num_tendon floats
```
mstoelzle: This should be removed as the distance will not be a constant (anymore).

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,336 @@
...
+
+    @eqx.filter_jit
+    def _actuation_matrix(self, q: Array, s: Array) -> Array:
+        """
+        """
+
+        def Phi_a_k(k, j):
+            idx = 6*j
+            column = jnp.array([
+            (f[k]*s*self.my[k] + self.ry[k])*(f[k]*self.mz[k] + xi[idx]*(f[k]*s*self.my[k] + self.ry[k]) + xi[idx+5]) + \
```
mstoelzle: As we discussed, the next step is to move to a numerical implementation and integration.

mstoelzle (changes requested):

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: For now, I implemented a general computation for the actuation matrix that takes user-specified tendon_routing functions for the tendon routings and their derivative along s, so that only the actuation matrix is numerically integrated along s.

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: Done.

examples/simulate_tendon_actuated_pcs.py
```diff
@@ -137,6 +137,25 @@ def update_plot(frame_idx):
         )  # Slider cannot be converted to HTML
 
 
+def linear_routing(tendon_params, s):
```
mstoelzle: Please put them later directly into the `tendon_actuated_pcs.py` file as a default initialization if no routing functions are provided by the user.

mstoelzle: Also, later, please add docstrings and code comments everywhere

examples/simulate_tendon_actuated_pcs.py
```diff
@@ -137,6 +137,25 @@ def update_plot(frame_idx):
...
+    ry      =   tendon_params['ry']
+    my      =   tendon_params['my']
+    rz      =   tendon_params['rz']
+    mz      =   tendon_params['mz']
+    f       =   tendon_params['f']
+    d_s     =   jnp.array([0.0, ry + my*s*f, rz + mz*s*f, 1.0])
+    return d_s
+
+
+def d_linear_routing_ds(tendon_params, s):
```
mstoelzle: Maybe you could renaming to `linear_routing_derivative` and explain in the docstring what kind of derivative this exactly is.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -47,14 +47,14 @@ class TendonActuatedPCS(PCS):
...
-        Distances of the straight tendons from the segment's backbone.
-    alpha: Array
-        Angles of the longitudinal planes where the tendons lie, 0 if aligned to the z-axis
-    x_n: Array
-        Depths of the attachment points of the tendons, along the x-axis
-    segment_indices_to_actuate : Array
-        Indices of the segments that are actuated.
+    tendon_params : Dict[str, Array]
+        Dictionary of arrays of length n_actuators representing the tendon parameters.
+    d_s : Callable
```
mstoelzle, line 53: - This doesn't match the actual parameters of the function.
- Also, I would suggest to first ask as arguments for the actuation basis and only then for the tendon params.
- Finally, why is one called `actuation_basis` and then other `tendon_params`? Seems a bit inconsistent. Either, I would go with `tendon_routing_basis` and `tendon_routing_params` or with `actuation_basis` and `actuation_params`.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -97,213 +95,85 @@ def __init__(
...
+                """
+                (3,3)
+                """
+                return jnp.array([[0, -vec[2], vec[1]], [vec[2], 0, -vec[0]], [-vec[1], vec[0], 0]])
+
+            def hat(vec):
+                """
+                (4,4)
+                """
+                return jnp.vstack([jnp.column_stack([tilde(vec[:3]), vec[3:]]), jnp.zeros(4)])
```
mstoelzle: Please check if you can re-use here some existing functions. For example, in `utils/lie_algebra`. If they don't already exist, I would suggest placing these helper math functions in `math_utils.py` for now.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -317,20 +187,32 @@ def A_j(j):
...
+        """
+        Compute the actuation matrix of the robot.
+
+        Args:
+            q (Array): generalized coordinates of shape (num_active_strains,).
+
+        Returns:
+            A (Array): Actuation matrix of shape (num_active_strains, num_actuators).
+        """
+        A = self.compute_actuation_matrix(q)
```
mstoelzle: You could put the code from `compute_actuation_matrix` directly in here. Why is there is additional abstraction?

mstoelzle (changes requested):

examples/simulate_tendon_actuated_pcs.py
Michele-Martini: Added.

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: This is an override of the parent class PCS, which has a method called actuation_matrix that is used during resolve_upon_time

mstoelzle: I think we are misunderstanding each other. What I am saying is you could just put the content that is currently in the `compute_actuation_matrix` method instead directly into the `actuation_matrix` method.

Michele-Martini: Oh sure, sorry for the confusion. Done.

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: Accepted.

examples/simulate_tendon_actuated_pcs.py
Michele-Martini: Accepted. With also defaults tendon_params = empty arrays.

examples/simulate_tendon_actuated_pcs.py
Michele-Martini: What do you mean kind of derivative? The geometrical interpretation?

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: - indeed. I found more intuitive for the user to provide the depth of the attachment point lt and then compute what I need: f = lt / L .
- How can I make it? Like setting the parameters only with a manual call of a _set_tendon_params method?
- I agree. I would go for tendon_routing_basis and tendon_routing_params

mstoelzle: > How can I make it? Like setting the parameters only with a manual call of a _set_tendon_params method?

No, I mean just changing the order of arguments in the method call. So something in the direction of

```python
def __init__(
        self,
        num_segments: int,
        params: Dict[str, Array],
        tendon_routing_basis: Dict[str, Callable],
        tendon_routing_params: Dict[str, Array],
        order_gauss: int = 5,
        strain_selector: Optional[Array] = None,
        xi_ref: Optional[Array] = None,
    ):
```

examples/simulate_tendon_actuated_pcs.py
mstoelzle: Specify, that this a derivative of the tendon routing with respect to `s` (and not for example time or something else).

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: Now the tendon_routing_params accepts lt as an array of integers, representing the index of the segment they attach to. The corresponding tendon_routing_basis are not function of lt anymore, as the effect of the attachment is considered in the actuation matrix (specifically, as a simple condition in the _actuation_matrix_s method).

src/soromox/systems/tendon_actuated_pcs.py
mstoelzle, line 1: @Michele-Martini Please run `ruff format src/soromox/tendon_actuated_pcs.py` to format the code nicely.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,337 @@
...
+        Reference strain (reference configuration) of the robot.
+    num_gauss_points : int
+        Number of points used for numerical integration.
+        Corresponds to the order of Gauss-Legendre quadrature + 2 (for the endpoints).
+    Xs, Ws : Array
+        Gauss-Legendre quadrature nodes and weights for numerical integration.
+    tendon_routing_params : Dict[str, Array]
+        Dictionary of arrays of length n_actuators representing the tendon parameters.
+    d_s : Callable
+        Function that returns the homogeneous vector (4D) of the distance of the tendons
```
mstoelzle: Can you please what convention you follow for the homogenous coordinates? I.e., what is the order of entries and what is the meaning of each entry?

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,337 @@
...
+
+        # update all fields
+        updated_self = eqx.tree_at(lambda x: x.tendon_routing_params, updated_self, tendon_routing_params)
+        updated_self = eqx.tree_at(lambda x: x.num_actuators, updated_self, num_actuators)
+
+        return updated_self
+
+    
+    @eqx.filter_jit
+    def _actuation_matrix_s1(self, q: Array, s: Array) -> Array:
```
mstoelzle: Why do we have `_actuation_matrix_s1` and `_actuation_matrix_s `? What is the difference?

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,337 @@
...
+        
+        Phi_a = Phi_a.reshape((6*self.num_segments, self.num_actuators), order='C')     # (num_segments*6, num_actuators)
+
+        A_local = self.B_xi.T @ Phi_a                                                   # (num_segments*6, num_actuators)
+
+        return A_local
+
+    
+    @eqx.filter_jit
+    def _actuation_matrix_s(self, q: Array, s: Array) -> Array:
```
mstoelzle: To follow the convention that Solange used, maybe this could be named `_local_actuation_matrix`?

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,337 @@
...
+        # # For debugging purposes, you can uncomment the following line to see the step-by-step computation
+        # A_blocks_tot = jnp.stack([A_i(i) for i in range(self.num_segments)], axis=0)
+
+        A_full = jnp.sum(A_blocks_tot, axis=(0, 1))  # Sum over segments and Gauss points
+        
+        return A_full
+    
+
+    @eqx.filter_jit
+    def tendon_kinematics(self, q: Array, s: Array) -> Array:
```
mstoelzle: I think it might be best to call it `forward_kinematics_tendons`? We also do similar stuff in other places, such as `forward_kinematics_joints`, `forward_kinematics_end_effector`, etc.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,337 @@
...
+            t_s = t_s[:-1]                                      # (3,)
+            
+            lt = self.L_cum[tendon_routing_params["lt"] + 1]    # ()
+            mask = s_val <= lt                                  # ()
+
+            t_s = jnp.where(mask, t_s, jnp.nan)                 # (3,)
+
+            return t_s
+        
+        t = vmap(
```
mstoelzle: Maybe it would make sense to instead just have a function `forward_kinematics_tendon` where the user provides a tendon idx as an argument. Then, the user can apply `vmap` themselves to get the forward kinematics for each tendon.

Currently, when I only want to know the forward kinematics of a single tendon, it is a bit annoying as I am wasting computation.

mstoelzle (changes requested):

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: Sorry that was the previous function that I forgot to delete after I tested the new one. I deleted _actuation_matrix_s1 now.

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: Of course

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: I agree, as the other functions. Thank you for the aligment.

examples/simulate_tendon_actuated_pcs.py
```diff
@@ -0,0 +1,409 @@
...
+        update_plot(0)  # Initial frame
+
+        if show:
+            plt.show()
+        plt.close(fig)
+
+        return HTML("Slider animation not implemented in HTML format. Use matplotlib directly to view the slider.")
+
+
+def linear_routing(tendon_routing_params, s):
```
mstoelzle: @Michele-Martini Why do we need the `linear_routing` and `linear_routing_derivative` functions here? Because, they are also defined in the `tendon_actuated_pcs.py` file.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,376 @@
...
+    Args:
+            tendon_routing_params (Dict[Array]): parameters of the linear tendons (6,)
+                ry: y-coordinate of the pulling point of the tendons [m]
+                my: slope coefficient in the x-y plane of the tendons [-]
+                rz: z-coordinate of the pulling point of the tendons [m]
+                mz: slope coefficient in the x-z plane of the tendons [-]
+            s (Array): abscissa points (num_gauss_points,)
+
+        Returns:
+            d_s (Array): position of the tendon at s wrt centerline, in homogeneous coordinates (4,)
```
mstoelzle: - Please add more information about the order of the homogeneous coordinates?
- Does it provide a lot of advantages of using  homogeneous coordinates? Could the "1" at the end of the vector not be easily added in the main code, instead of requiring to be present in the routing basis functions?

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,376 @@
...
+
+        Args:
+            q (Array): generalized coordinates of shape (num_active_strains,)
+            s (Array): abscissa points (num_gauss_points,)
+
+        Returns:
+            A_s (Array): Actuation matrix of shape at s (num_active_strains, num_actuators)
+        """
+
+        def Phi_a_kj(tendon_routing_params, j):
```
mstoelzle: Please add a docstring to the function

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,376 @@
...
+        Compute the actuation matrix of the robot.
+
+        Args:
+            q (Array): generalized coordinates of shape (num_active_strains,).
+
+        Returns:
+            A (Array): Actuation matrix of shape (num_active_strains, num_actuators).
+        """
+
+        def A_i(i):
```
mstoelzle: Please add a docstring to the function

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,376 @@
...
+                )
+
+            strain_mask = build_block(
+                6 * self.num_segments, # numer of rows
+                self.num_actuators, # number of columns
+                6 * i, # starting index
+                6, # number of rows set to 1 starting from starting index
+            )
+
+            def A_j(j):
```
mstoelzle: Please add a docstring to the function

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,376 @@
...
+        Returns:
+            A (Array): Actuation matrix of shape (num_active_strains, num_actuators).
+        """
+
+        def A_i(i):
+            Xs_scaled, Ws_scaled = scale_gaussian_quadrature(
+                self.Xs, self.Ws, self.L_cum[i], self.L_cum[i + 1]
+            )
+
+            def build_block(n1, n2, j, step):
```
mstoelzle: Please add a docstring to the function

src/soromox/systems/tendon_actuated_pcs.py
mstoelzle: Please specify the type of `i` as an `Array`

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,376 @@
...
+        Args:
+            q (Array): generalized coordinates of shape (num_active_strains,)
+            s (Array): abscissa points (num_gauss_points,)
+
+        Returns:
+            A_s (Array): Actuation matrix of shape at s (num_active_strains, num_actuators)
+        """
+
+        def Phi_a_kj(tendon_routing_params, j):
+            def get_block(xi, j, block_size=6):
```
mstoelzle: Please specify the type of inputs as arrays and integers, respectively.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,376 @@
...
+    @eqx.filter_jit
+    def forward_kinematics_tendons(
+        self, tendon_routing_params: Dict[str, float], q: Array, s: Array
+    ) -> Array:
+        """
+        Compute the forward kinematics of the robot at a point s along the robot.
+
+        Args:
+            q (Array): generalized coordinates of shape (num_active_strains,).
+            s (Array): point coordinate along the robot in the interval [0, L]. (n_s,)
```
mstoelzle: I think that this docstring is likely out of date. Because `s` here is just a single coordinate, instead of a vector of coordinates, right?

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,376 @@
...
+
+        Returns:
+            t (Array): 3D position of the tendons at s, shape (n_tendons, n_s, 3)
```
mstoelzle: I think that this docstring is likely out of date. Because it is probably returning an array of shape `(n_tendons, 3)`

mstoelzle (changes requested):

examples/simulate_tendon_actuated_pcs.py
Michele-Martini: To explicitly show to the user that the tendon routing basis can be customizable, as this is an example. Or maybe we could make a library of different routing basis (like Sorosim) and then let the user pick from them? So I can delete the repetition here. What do you think?

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: You are right, we can do it inside the actuation_matrix method.

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: Even if in Phi_a_kj I set it to Array?

examples/simulate_tendon_actuated_pcs.py
mstoelzle: You could put these routing functions into `soromox/actuation/tendon_actuation.py`. Then, we can create docs pages for them, which should be sufficient example/documentation.

src/soromox/systems/tendon_actuated_pcs.py
mstoelzle: I am referring here to the `get_block` function, whose types are probably `xi: Array, j: Array, block_size: int = 6`

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: It returns the vector of one tendon at s, so (3,)

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,376 @@
...
+
+        A_full = jnp.sum(
+            A_blocks_tot, axis=(0, 1)
+        )  # Sum over segments and Gauss points
+
+        return A_full
+
+    @eqx.filter_jit
+    def forward_kinematics_tendons(
+        self, tendon_routing_params: Dict[str, float], q: Array, s: Array
```
mstoelzle: The `tendon_routing_params` are already stored in `self`. Why do they need to be provided again?

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,376 @@
...
+    ) -> Array:
+        """
+        Compute the forward kinematics of the robot at a point s along the robot.
+
+        Args:
```
mstoelzle, line 446: The `tendon_routing_params` are missing from the docstring "Args"

mstoelzle (changes requested):

src/soromox/systems/tendon_actuated_pcs.py
mstoelzle: Ok, but then, I don't understand how I can choose which tendon I want to evaluate the forward kinematics for. Should I have to provide something like a `tendon_idx`?

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: True. Then I can just return the values for all the tendons like you said (n_actuators, 3).

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: We can remove the tendon_routing_params altogether as per the previous comment, so to return (n_actuators, 3).

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: Now it returns the vector (n_tendons, 3), so the user can index themself.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,392 @@
...
+        Returns:
+            A_s (Array): actuation matrix of shape at s (num_active_strains, num_actuators)
+        """
+
+        def Phi_a_kj(tendon_routing_params: Dict[str, Array], j: Array):
+            """
+            Compute the the actuation basis of the tendon at the segment j of the robot.
+
+            Args:
+                tendon_routing_params (Dict[str, Array]): parameters of the linear tendons (6,)
```
mstoelzle: - This is different from the usual `tendon_routing_params` as that it only corresponds to the parameters of a *single* instead of all tendons, right? If this interpretation is correct, this should be reflected both in (a) the argument name, and (b) the docstring description.
- Maybe don't mention/overspecify to "linear tendons" here, because this should be a general description.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,392 @@
...
+
+            def A_j(j: Array):
+                """
+                Compute the actuation matrix at the abscissa point corresponding to the gaussian point j.
+
+                Args:
+                    j (Array): index of the gaussian point ()
+
+                Returns:
+                    A_j (Array): local actuation matricx of shape (num_active_strains, num_actuators).
```
mstoelzle: Typo in "matrix"

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,392 @@
...
+                    j (int): index of the first row to set to 1 ()
+                    step (int): number of rows to set to 1 starting from index j ()
+
+                Returns:
+                    A_i (Array): stack of actuation matrices of shape (num_gauss_points, num_active_strains, num_actuators).
+                """
+                a = jnp.zeros((n1, n2), dtype=jnp.int32)
+                block = jnp.ones((step, n2), dtype=jnp.int32)
+                # j_clipped = jnp.clip(j, 0, n2 + step)
+                return lax.dynamic_update_slice(
```
mstoelzle: We found that such in-memory operations, such as `lax.dynamic_update_slic` can be a performance bottleneck. Instead of first intializing a default matrix and then updating some parts, its usually better to, for example, concatenate and stack different matrix blocks.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,392 @@
...
+        """
+
+        def forward_kinematics_tendon_k(
+            tendon_routing_params: Dict[str, float], q: Array, s: Array
+        ) -> Array:
+            """
+            Compute the forward kinematics of the specified tendon actuator at a point s along the robot.
+
+            Args:
+                tendon_routing_params (Dict[str, float]): parameters of the tendon (6,)
```
mstoelzle: Similar to another comment, it should be reflected in the argument name and in the docstring that the entries of this dictionary follow a different data structure than `self.tendon_routing_params`

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,392 @@
...
+        Args:
+            q (Array): generalized coordinates of shape (num_active_strains,)
+            s (Array): point coordinate along the robot ()
+
+        Returns:
+            t_s (Array): cartesian position of the tendons at s, shape (n_actuators, 3)
+        """
+
+        def forward_kinematics_tendon_k(
+            tendon_routing_params: Dict[str, float], q: Array, s: Array
```
mstoelzle: This typehint is very likely wrong. Instead of `Dict[str, float]`, it should be `Dict[str, Array]`. Please fix everywhere.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,392 @@
...
+                t_k_s (Array): cartesian position of the tendons at s, shape (n_actuators, 3)
+            """
+            g_s = self.forward_kinematics(q, s)  # (4,4)
+            t_k_s = g_s @ jnp.append(self.d_s(tendon_routing_params, s), 1.0)  # (4,)
+            t_k_s = t_k_s[:-1]  # (3,)
+
+            lt = self.L_cum[tendon_routing_params["lt"] + 1]  # ()
+            mask = s <= lt  # ()
+
+            t_k_s = jnp.where(mask, t_k_s, jnp.nan)  # (3,)
```
mstoelzle: This looks very dangerous - particularly, for applications that require gradients, as just all gradients might become `NaN`. Why are you setting these entries to `NaN` instead of to, for example, zero? Also, could you add a few more code comments what this mask is doing?

src/soromox/actuation/tendon_actuation.py
```diff
@@ -0,0 +1,48 @@
...
+    Args:
+            tendon_routing_params (Dict[Array]): parameters of the linear tendons (6,)
+                ry: y-coordinate of the pulling point of the tendons [m]
+                my: slope coefficient in the x-y plane of the tendons [-]
+                rz: z-coordinate of the pulling point of the tendons [m]
+                mz: slope coefficient in the x-z plane of the tendons [-]
+            s (Array): abscissa points (num_gauss_points,)
+
+        Returns:
+            d_s (Array): position of the tendon at s wrt the centerline (3,)
```
mstoelzle: Please specify the order of entries of `d_s` here. Also, it is probably good to add a comment here that the x-entry should usually be zero, as the robot is aligned with the local x-axis.

src/soromox/actuation/tendon_actuation.py
```diff
@@ -0,0 +1,48 @@
...
+    Specifically, dd_s_ds is the derivative of d_s over s, at s.
+
+    Args:
+            tendon_routing_params (Dict[Array]): parameters of the linear tendons (6,)
+                my: slope coefficient in the x-y plane of the tendons [-]
+                mz: slope coefficient in the x-z plane of the tendons [-]
+            s (Array): abscissa points (num_gauss_points,)
+
+        Returns:
+            dd_s_ds (Array): path of the tendon at s wrt the centerline (3,)
```
mstoelzle: Similar considerations as I have left for `linear_routing` also apply here.

mstoelzle (changes requested):
Needs a few more finishing touches

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: For consistency with the current variable names, I called it "single_tendon_routing_params". Do you like it?

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: I understand, for the gradient can surely be problematic if non-treated. I am setting to NaN because it thought it was semantically correct, meaning that if you ask for a tendon position at an s that is higher that the tendon's length it replies that there is not such a thing. Zero can be interpreted as the origin. What do you think?

mstoelzle: There are two aspects that need to be separated:

1. The user might provide tendon parameters that should be invalid. However, these should be filtered out at parameter-update time and *not* at runtime. So, you should add `assert` and `raise ValueError` statements to the `_set_tendon_params` and `update_tendon_params` methods to ensure that the parameters stored are actually valid.
2. You might create masks at runtime to filter out contributions to the actuation matrix of segments that you don't want. But here, you should only work with zeros. (I am not sure if this aspect applies to this piece of code).

Michele-Martini: Sorry, I am getting confused. I can add the assert in the set method, but it is allowed to have a tendon that is attached, say, at the end of the first segment and it ends there (meaning lt = 0), even if the robot has 2 segments with total length L = 2 * Ls. So s \in [0, L], but if the user asks for the tendon position at s = L, this should return a zeros(3), correct?

examples/simulate_tendon_actuated_pcs.py
```diff
@@ -0,0 +1,372 @@
...
+        * jnp.array([0.0, 0.0]),  # z-coordinate of the pulling point of the tendons [m]
+        "my": jnp.array(
+            [0.0, 0.0]
+        ),  # slope coefficient in the x-y plane of the tendons [-]
+        "mz": jnp.array(
+            [0.0, 0.0]
+        ),  # slope coefficient in the x-z plane of the tendons [-]
+        "lt": jnp.array(
+            [1, 0]
+        ),  # length of the tendons = x-coordinate of the attachment points [m]
```
mstoelzle, line 249: This code comment is outdated, because now, you are specifying here the segment indices, right?

Also, would it make sense to rename the parameter `lt` to something else? Because, currently it is implied that it contains the length of the tendon.

src/soromox/systems/tendon_actuated_pcs.py
mstoelzle: I thought about this a bit more. I guess there are two options that are sensible in this case:

- Returning NaNs are you are currently doing. Sufficient for visualization, but bad if these functions are used in a computational fashion for optimization applications
- Returninging the forward kinematics for the highest `s` that is valid. So, for example, if `lt = 0` and `s = 1.5 * Ls`, then return the forward kinematics for `s = Ls`. This should also work decently for visualization purposes as it will end up  with a lot of "dots" at the same position (i.e., for all `s` that are past the tendon end point).

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: I am having issues stacking since the index j is dynamically tracked and it appears that it brings issue to the instantiation of a matrix with shape determined by a dynamically tracked value. Thus, I made a masking operation yet again, but it seems to be comparable in performance to the dynamic_slice

mstoelzle: A few more alternative options:

1. You could pre-compute the mask matrices at initialization time, store them, and access the suitable mask matrix at runtime
2. You could create the mask on-the-fly by comparing the index with an arange-based array. Something along the lines of

```python
@jax.jit
def compute_strains_mask(n1: int,
                         n2: int,
                         j: Union[int, Array],
                         step: Union[int, Array],
                         *,
                         dtype=jnp.int32) -> Array:
    """
    Return an (n1, n2) mask of 0/1 where rows in [j, j+step) are 1 and the rest 0.

    Notes:
    - No scatters or dynamic updates; built from comparisons + broadcasting.
    - j, step can be traced scalars (shape ()), and `n1`, `n2` should be static per compiled instance.
    - Out-of-range indices are safely clipped into [0, n1], negative or zero step yields all zeros.
    """
    j = jnp.asarray(j, jnp.int32)
    step = jnp.asarray(step, jnp.int32)
    step = jnp.maximum(step, 0)  # guard against negative lengths

    rows = jnp.arange(n1, dtype=jnp.int32)  # (n1,)

    # Clip the active interval into bounds to match "write a block of ones" semantics.
    start = jnp.clip(j, 0, n1)
    stop  = jnp.clip(j + step, 0, n1)

    row_mask = (rows >= start) & (rows < stop)      # (n1,)
    mask = jnp.broadcast_to(row_mask[:, None], (n1, n2))  # (n1, n2)
    return mask.astype(dtype)
```

mstoelzle: The simplified 1d variant to motivate how it works is: `mask = (jnp.arange(size) == idx).astype(dtype)`

Michele-Martini: > A few more alternative options:
> 
> 1. You could pre-compute the mask matrices at initialization time, store them, and access the suitable mask matrix at runtime
> 2. You could create the mask on-the-fly by comparing the index with an arange-based array. Something along the lines of
> 
> ```python
> @jax.jit
> def compute_strains_mask(n1: int,
>                          n2: int,
>                          j: Union[int, Array],
>                          step: Union[int, Array],
>                          *,
>                          dtype=jnp.int32) -> Array:
>     """
>     Return an (n1, n2) mask of 0/1 where rows in [j, j+step) are 1 and the rest 0.
> 
>     Notes:
>     - No scatters or dynamic updates; built from comparisons + broadcasting.
>     - j, step can be traced scalars (shape ()), and `n1`, `n2` should be static per compiled instance.
>     - Out-of-range indices are safely clipped into [0, n1], negative or zero step yields all zeros.
>     """
>     j = jnp.asarray(j, jnp.int32)
>     step = jnp.asarray(step, jnp.int32)
>     step = jnp.maximum(step, 0)  # guard against negative lengths
> 
>     rows = jnp.arange(n1, dtype=jnp.int32)  # (n1,)
> 
>     # Clip the active interval into bounds to match "write a block of ones" semantics.
>     start = jnp.clip(j, 0, n1)
>     stop  = jnp.clip(j + step, 0, n1)
> 
>     row_mask = (rows >= start) & (rows < stop)      # (n1,)
>     mask = jnp.broadcast_to(row_mask[:, None], (n1, n2))  # (n1, n2)
>     return mask.astype(dtype)
> ```

This is very similar to what I have implemented in the form of masking. Maybe I can try to compare this with the stored version if I notice any performance change

examples/simulate_tendon_actuated_pcs.py
Michele-Martini: What about "idx_a" for index of attachment?

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: I implemented the first option, computing at initialization the matrices and then accessing them at runtime.

examples/simulate_tendon_actuated_pcs.py
mstoelzle: Maybe `idx_seg_att`?

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,393 @@
...
+        self._set_B_xi_segments()
+
+    def _set_B_xi_segments(self):
+        strains_masks = []
+        for i in range(self.num_segments):
+            idx = 6*i
+            strais_mask_i = jnp.zeros_like(self.B_xi, dtype=self.B_xi.dtype)
+            strais_mask_i = strais_mask_i.at[idx:idx+6, idx:idx+6].set(self.B_xi[idx:idx+6, idx:idx+6])
+            strains_masks.append(strais_mask_i)
+        self.B_xi_segments = jnp.array(strains_masks)
```
mstoelzle: This should be jnp.stack as `strais_mask_i` is already an `Array`.

examples/simulate_tendon_actuated_pcs.py
Michele-Martini: Clear > verbose, so let's go idx_seg_att :)

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,393 @@
...
+            }
+        self._set_tendon_routing_basis(tendon_routing_basis)
+        self._set_tendon_routing_params(tendon_routing_params)
+        self._set_B_xi_segments()
+
+    def _set_B_xi_segments(self):
+        strains_masks = []
+        for i in range(self.num_segments):
+            idx = 6*i
+            strais_mask_i = jnp.zeros_like(self.B_xi, dtype=self.B_xi.dtype)
```
mstoelzle: Typo. Should this be `strain_mask_i`?

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,393 @@
...
+            strais_mask_i = strais_mask_i.at[idx:idx+6, idx:idx+6].set(self.B_xi[idx:idx+6, idx:idx+6])
+            strains_masks.append(strais_mask_i)
+        self.B_xi_segments = jnp.array(strains_masks)
+
+    def _set_tendon_routing_params(self, tendon_routing_params: Dict[str, Array]):
+        if not isinstance(tendon_routing_params, (dict)):
```
mstoelzle, line 212: You could add some varlidation of the params here. For example, that the segment attachment indices are in the feasible range (i.e., smaller than `num_segments`)

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,393 @@
...
+            raise TypeError(
+                "The parameter 'tendon_routing_params' must be a dictionary of jnp.ndarrays."
```
mstoelzle: This should be

```python
"The parameter 'tendon_routing_params' must be a dictionary of jax.Array."
```

mstoelzle (changes requested):

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: Rewrote the function with vmap and remove loop with list.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,412 @@
...
+        updated_self = eqx.tree_at(
+            lambda x: x.tendon_routing_params, updated_self, tendon_routing_params
+        )
+        updated_self = eqx.tree_at(
+            lambda x: x.num_actuators, updated_self, num_actuators
+        )
+
+        return updated_self
+    
+    def check_tendon_routings(self):
```
mstoelzle: This should be marked as an internal function. Therefore, the name should be `_check_tendon_routings`

mstoelzle: Please add a docstring describing what exactly the function is checking.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,412 @@
...
+        if tendon_routing_basis is None:
+            tendon_routing_basis = {
+                "d_s": act.linear_routing,
+                "dd_s_ds": act.linear_routing_derivative,
+            }
+        self._set_tendon_routing_basis(tendon_routing_basis)
+        self._set_tendon_routing_params(tendon_routing_params)
+        self._set_B_xi_segments()
+
+    def _set_B_xi_segments(self):
```
mstoelzle: Please add a docstring

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -0,0 +1,412 @@
...
+    Xs, Ws : Array
+        Gauss-Legendre quadrature nodes and weights for numerical integration.
+    tendon_routing_params : Dict[str, Array]
+        Dictionary of arrays of length n_actuators representing the tendon parameters.
+    d_s : Callable
+        Function that returns the vector [d_x, d_y, d_z] of the tendon
+        position w.r.t. the central backbone within the cross-sectional plane at a given
+        abscissa point s. The three entries represent the coordinate of the tendon
+        position with respect to the local cross-sectional frame at s. For this reason,
+        d_x is always equal to 0 (as the backbone is pointing in the local x-direction)
```
mstoelzle, line 58: Please add the docstring about `self.B_xi_segments`

mstoelzle (changes requested):

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: I haven't added them yet before you agreed on the structure.

mstoelzle: I agree on the structure!

Michele-Martini: Ahaha good. Do you want me to add the doctrings also to the other _set methods?

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -160,27 +159,55 @@ def __init__(
...
-            idx = 6*i
+        """
+        Compute the strains basis by segments. While the full strains basis B_xi
+        is a block diagonal matrix composed by the 6x6 strain basis of each segment,
+        this function extracts each diagonal block and returns it as a matrix of shape
+        (num_segments, num_active_strains, num_active_strains). This is used for the
+        computation of the actuation matrix.
+
+        Returns:
+            B_xi_segments (Array): strain basis of each segments (num_segments, num_active_strains, num_active_strains)
```
mstoelzle, line 181: Slight inaccuracy: `B_xi_segments` is not returned, but instead written to the attributes. Alternatively, you could name the function `_B_xi_segments`, return `B_xi_segments` and write in the `__init__` method: `self.B_xi_segments = _B_xi_segments()`

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -194,28 +221,86 @@ def _set_tendon_routing_params(self, tendon_routing_params: Dict[str, Array]):
...
+        This internal function stores as attributes of the class the basis functions of
+        the tendon routings specified by the user.
+
+        Args:
+            tendon_routing_basis (Dict[str, Callable]): basis functions of the tendons
+        """
         self.d_s = tendon_routing_basis["d_s"]
         self.dd_s_ds = tendon_routing_basis["dd_s_ds"]
 
+    def _check_tendon_routings(self, tendon_routing_params):
```
mstoelzle: I imagine that there are also other tendon routing checks possible (e.g., checking if they are intersecting in an infeasible fashion, etc.). Maybe it would make sense to specifically point out in the method name that this method is checking if the tendons are routed inside the body? Something like: `_check_tendon_routing_in_body`

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -194,28 +221,86 @@ def _set_tendon_routing_params(self, tendon_routing_params: Dict[str, Array]):
...
+            idx = jnp.sum(cond) - 1
+            return self.r[idx]
+
+        s = jnp.linspace(0.0, self.L_cum[-1], 75)
+        t = vmap(
+            vmap(self.d_s, in_axes=(None, 0), out_axes=0), in_axes=(0, None), out_axes=0
+        )(tendon_routing_params, s)  # (num_tendons, N, 3)
+        d = t[:, :, 1:]  # (num_tendons, N, 2)
+        radii = jnp.linalg.norm(d, axis=2)  # (num_tendons, N)
+        r = vmap(r_s)(s)  # (N,)
```
mstoelzle: Rename to `r_body` for better semantic understanding.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -194,28 +221,86 @@ def _set_tendon_routing_params(self, tendon_routing_params: Dict[str, Array]):
...
+            cond = self.L_cum <= s
+            idx = jnp.sum(cond) - 1
+            return self.r[idx]
+
+        s = jnp.linspace(0.0, self.L_cum[-1], 75)
+        t = vmap(
+            vmap(self.d_s, in_axes=(None, 0), out_axes=0), in_axes=(0, None), out_axes=0
+        )(tendon_routing_params, s)  # (num_tendons, N, 3)
+        d = t[:, :, 1:]  # (num_tendons, N, 2)
+        radii = jnp.linalg.norm(d, axis=2)  # (num_tendons, N)
```
mstoelzle: Rename to `r_tendons` for better semantic understanding.

mstoelzle (changes requested):

tests/test_tendon_actuated_pcs.py
```diff
@@ -0,0 +1,166 @@
...
+            params=params,
+            order_gauss=5,
+            # strain_selector=strain_selector,
+            # xi_ref=xi_ref,
+            tendon_routing_params=tendon_params,
+        )
+        
+        return robot, jnp.zeros(6*num_segments)
+    
+    def A_special_case(spec_params, tendon_params):
```
mstoelzle: Please specify in the function name what special case this is to improve the semantic understanding. I guess in this case that it is for a straight backbone / zero configuration.

tests/test_tendon_actuated_pcs.py
```diff
@@ -0,0 +1,166 @@
...
+
+        robot = TendonActuatedPCS(
+            num_segments=num_segments,
+            params=params,
+            order_gauss=5,
+            # strain_selector=strain_selector,
+            # xi_ref=xi_ref,
+            tendon_routing_params=tendon_params,
+        )
+
```
mstoelzle: Add a line here `q = jnp.zeros(6*num_segments)` to make the code easier to understand.

tests/test_tendon_actuated_pcs.py
```diff
@@ -0,0 +1,166 @@
...
+            -l**2*my/2 - l*ry,
+            l,
+            l*my,
+            l*mz
+        ])
+
+        return A / jnp.sqrt(my**2 + mz**2 + 1)
+    
+    def actuation_basis(tendon_params, xi, s):
+        ry, rz, my, mz = tendon_params["ry"][0], tendon_params["rz"][0], tendon_params["my"][0], tendon_params["mz"][0]
```
mstoelzle: Please add a docstring to explain a bit better what his `actuation_basis` is. Also, you could rename to `reference_actuation_basis` or `target_actuation_basis`

tests/test_tendon_actuated_pcs.py
mstoelzle: Also, you could specify in the function name that this is the reference/target.

mstoelzle: Also, please explain in the docstring what special setting/configuration you are looking at here and for what reason.

tests/test_tendon_actuated_pcs.py
```diff
@@ -0,0 +1,166 @@
...
+        (
+            jnp.zeros_like(xi_ref),
+            robot.L_cum[-1]*0.25,
+            actuation_basis(pars, xi_ref, robot.L_cum[-1]*0.25)
+        ),
+
+        (
+            jnp.array([0., jnp.pi*3, jnp.pi*2, 0.1, 0.2, 0.]),
+            robot.L_cum[-1]*0.25,
+            actuation_basis(pars, jnp.array([0., jnp.pi*3, jnp.pi*2, 0.1, 0.2, 0.]) + xi_ref, robot.L_cum[-1]*0.25)
```
mstoelzle: This is not great. Because you need to specify the configuration in two spots. Why not move the computation of the expected value into the for loop? Then, you can access the `q` from the for loop iteration.

tests/test_tendon_actuated_pcs.py
```diff
@@ -0,0 +1,166 @@
...
+            {"L": jnp.array([0.5, 0.6])},
+            {"ry": jnp.array([0.15]), "rz": jnp.array([0.]), "my": jnp.array([0.01]), "mz": jnp.array([0.01]), 'idx_seg_att': jnp.array([1])},
+            A_special_case({"L": jnp.array([0.5, 0.6])},
+                           {"ry": jnp.array([0.15]), "rz": jnp.array([0.]), "my": jnp.array([0.01]), "mz": jnp.array([0.01]), 'idx_seg_att': jnp.array([1])})
+        ),
+        (
+            3,
+            {"L": jnp.array([0.5, 0.6, 0.7])},
+            {"ry": jnp.array([0.1]), "rz": jnp.array([-0.05]), "my": jnp.array([0.01]), "mz": jnp.array([-0.05]), 'idx_seg_att': jnp.array([2])},
+            A_special_case({"L": jnp.array([0.5, 0.6, 0.7])},
```
mstoelzle: You need to specify the robot params here twice, which is not great for maintainability. Instead, you could move the computation of the expected value into the for loop, where you can directly access the robot and tendon routing params.

mstoelzle (changes requested):

mstoelzle (approved):

PR #10, opened 2025-08-22
Fix mass matrix calculation by using polar moment of inertia for twist strain

mstoelzle:


solangegbv (approved):
I agree with the modification, in fact that's what I've done in gvs.py

PR #14, opened 2025-09-07
 Tendon Actuated Pendulum

Michele-Martini:
This work aims to introduce the class and the simulation example of a tendon actuated pendulum, with both active inelastic tendons and passive ones connected to spring-dampers.


Michele-Martini:
Work in progress. The following need to be fixed/decided:

- assert for _check_tendon_feasibility: conflict with previous check on the rank of matrices Ra and Rp, as they become mutually exclusive in the default case (meaning without any tendon).
- is it viable to make the introduced routing parameters static (they are constants) in the hope of improving performance?
- the video with the 2 segments pendulum and one active tendon seems off... why the tip swings if there is no elasticity at the joints?

examples/simulate_tendon_actuated_pendulum.py
```diff
@@ -0,0 +1,170 @@
...
+import numpy as onp
+from pathlib import Path
+from typing import Callable, Dict
+
+import soromox
+from soromox.systems.tendon_actuated_pendulum import TendonActuatedPendulum
+
+num_links = 2
+
+sym_exp_filepath = (
```
mstoelzle: This can be deleted as the symbolic expressions are not needed anymore.

src/soromox/systems/tendon_actuated_pendulum.py
```diff
@@ -0,0 +1,308 @@
...
+
+import soromox.utils.lie_algebra as lie
+import soromox.actuation.tendon_actuation as act
+
+from .pendulum import Pendulum
+
+
+class TendonActuatedPendulum(Pendulum):
+    """
+    Planar N-link serial pendulum (revolute chain) with closed-form JAX dynamics.
```
mstoelzle: Add some more information here about the specific tendon-actuation. Currently, this is just the default information about an N-link serial pendulum

src/soromox/systems/tendon_actuated_pendulum.py
```diff
@@ -0,0 +1,308 @@
...
+        self.Ra = jnp.asarray(tendon_params.get("Ra", jnp.identity(N)))
+        self.A = self.Ra.T
+
+        # Actuation matrix of passive tendons
+        self.Rp = jnp.asarray(tendon_params.get("Rp", jnp.identity(N)))
+        self.Ap = self.Rp.T
+        Np = len(self.Rp)
+
+        # Stiffness matrix of the passive tendons
+        kp = jnp.asarray(tendon_params.get("Kp", jnp.zeros(Np)))
```
mstoelzle: There should be a warning when passive tendons are defined, but `Kp` is not provided.

mstoelzle: If the user should only provide the diagonal stiffness values, then it would probably better to name the parameter in smaller case (i.e., `kp`). Also, I wonder if `kp` is semantically sufficiently informative. Maybe `k_pt` or `kpt` instead (for k passive tendon)?

src/soromox/systems/tendon_actuated_pendulum.py
```diff
@@ -0,0 +1,308 @@
...
+        """
+        super().__init__(
+            params
+        )
+
+        N = self.num_links
+
+        # Actuation matrix of active tendons
+        self.Ra = jnp.asarray(tendon_params.get("Ra", jnp.identity(N)))
+        self.A = self.Ra.T
```
mstoelzle: Maybe for consistency, you could name it `Aat` or `A_at`?

src/soromox/systems/tendon_actuated_pendulum.py
```diff
@@ -0,0 +1,308 @@
...
+        Compute the generalized elastic force from joint stiffness plus the elasticity from the
+        springs of the passive tendons.
+
+        Args:
+            q (Array): Joint angles, shape (N,) [rad]
+
+        Returns:
+            tau_el_tot (Array): Total elastic force vector, shape (N,) [N⋅m]
+        """
+        Lp = self.passive_tendon_length(q)
```
mstoelzle: Please instead overwrite the `stiffness_matrix` method and don't modify the `elastic_force` method.

mstoelzle (changes requested):
Good first draft.

Let's agree on a convention for naming the parameters and then apply this convention to all parameters and attributes.

src/soromox/systems/tendon_actuated_pendulum.py
Michele-Martini: About the naming conventions, since the class is called TendonActuatedPendulum, I find redundant to specify A_pt, maybe just A_a and A_p. What do you think?

mstoelzle: From this perspective I agree. However, I am thinking of a setting where now people actually just use the `soromox` package in their own codebase/repo. Then, they might specify/modify this `kp` parameter in the `params` dictionary and it is likely not very obvious to the one reading the code what is the meaning without first checking out the documentation of the soromox library.

src/soromox/systems/tendon_actuated_pendulum.py
Michele-Martini: In this case, how do you suggest to add the initial force due to the pre-stretch of the passive tendons (L_p0)?

mstoelzle: Ok, then I stand partially corrected:
- You implement the stiffness modification due the passive tendon into the `stiffness_matrix` method
- You implement the constant force due to the pre-stretch of the passive tendon into the `elastic_force` method

Michele-Martini:
> * is it viable to make the introduced routing parameters static (they are constants) in the hope of improving performance?

Would be worth it if we made all the constant matrices as onp.array and so fixed and static? Would it be compatible and would actually boost performance?

mstoelzle:
> > * is it viable to make the introduced routing parameters static (they are constants) in the hope of improving performance?
> 
> Would be worth it if we made all the constant matrices as onp.array and so fixed and static? Would it be compatible and would actually boost performance?

I am not sure to be honest. But intuitively, I wouln't expect it to make a crazy big difference. At the end, the matrix multiplications still need to be executed.

- Either we keep it as currently implemented as see how the performance is. I expect the performance to be pretty good.
- Or you set up some benchmarking script where you compare the performances of the two.

As this is not super crucial to your project, maybe I recommend to go with the 1st option for now.

mstoelzle:
@Michele-Martini Please ask Copilot to create the necessary modifications to the docs to show an entry for this `TendonActuatedPendulum` systems in the API documentation.

src/soromox/systems/tendon_actuated_pendulum.py
```diff
@@ -0,0 +1,354 @@
...
+
+        Args:
+            q (Array): Joint angles, shape (N,) [rad] (unused)
+
+        Returns:
+            A (Array): Actuation matrix, shape (N, Na)
+        """
+        return self.A_at
+
+    def actuation_force(self, q: Array, u: Array) -> Array:
```
mstoelzle: Sorry, this is actually a deficiency in the `Pendulum` class. Please implement the necessary modifications (e.g., evaluating the `actuation_matrix`) to the `actuation_force` method in the `Pendulum` class and remove the method from the `TendonActuatedPendulum` class

mstoelzle (approved):

PR #18, opened 2025-09-22
Refactorization DynamicalSystem.resolve_upon_time

Michele-Martini:
Refactored resolve_upon_time to be jax friendly and be used within jax.scan.


Michele-Martini:
Also, about the update_params and update_tendon_params methods, I think that we should re-do the checks with the asserts after each update... but with the current implementation is kind of weird... Maybe could we make a function that asserts everything so we can call it at our liking? @mstoelzle what do you think?

mstoelzle:
> Also, about the update_params and update_tendon_params methods, I think that we should re-do the checks with the asserts after each update... but with the current implementation is kind of weird... Maybe could we make a function that asserts everything so we can call it at our liking? @mstoelzle what do you think?

I think checking this after every update using asserts would likely destroy the jit-compilation of the optimizatoin loop. If you you only compile the inner loop (i.e., the simulation), then it might be fine. 

But because of these considerations, I wouldn't put the asserts in the update function of the SoRoMoX library. You can put it into your own code if you want.

Also, if it is a problem that your optimization (frequently) leads to unfeasible values, we need to think of solutions, such as projected gradient descent.

mstoelzle (approved):

PR #22, opened 2025-10-29
Fix wrong usage of strain basis in tendon-actuated pcs system

mstoelzle:
This pull request refactors the handling of strain basis matrices in the `TendonActuatedPCS` class and updates the actuation matrix computation to improve clarity and correctness. The main changes include replacing the segment-wise dense strain basis with a list of per-segment strain bases, simplifying the actuation matrix computation, and ensuring correct projection into the active strains space.

**Strain basis matrix handling:**
* Replaced the `B_xi_segments` attribute from a single dense array to a `List[Array]`, where each entry represents the strain basis for an individual segment (`B_xi_segments`), and introduced a new dense stack `B_xi_segments_dense` for all segments. Updated documentation accordingly.
* Removed the `_B_xi_segments` method and its usage, as the new approach no longer requires extracting diagonal blocks from a dense matrix. [[1]](diffhunk://#diff-b0093925a979f4790787ab7cb6ea560470a8b9f434306ea38565c8a3afc83fa9L168-L203) [[2]](diffhunk://#diff-b0093925a979f4790787ab7cb6ea560470a8b9f434306ea38565c8a3afc83fa9L78)

**Actuation matrix computation updates:**
* Updated the `_local_actuation_basis` and related actuation matrix computation methods to use the full strain space shape `(6 * num_segments, ...)` instead of the previous active strains shape, and clarified documentation for returned shapes. [[1]](diffhunk://#diff-b0093925a979f4790787ab7cb6ea560470a8b9f434306ea38565c8a3afc83fa9L326-R294) [[2]](diffhunk://#diff-b0093925a979f4790787ab7cb6ea560470a8b9f434306ea38565c8a3afc83fa9L392-R360) [[3]](diffhunk://#diff-b0093925a979f4790787ab7cb6ea560470a8b9f434306ea38565c8a3afc83fa9L403-R386)
* Changed the computation of the local actuation matrix to skip the segment-specific strain basis multiplication, instead applying the projection into the active strains space only at the end of the computation. [[1]](diffhunk://#diff-b0093925a979f4790787ab7cb6ea560470a8b9f434306ea38565c8a3afc83fa9L403-R386) [[2]](diffhunk://#diff-b0093925a979f4790787ab7cb6ea560470a8b9f434306ea38565c8a3afc83fa9L431-R407)

**Type annotation improvements:**
* Updated type annotations to reflect the new use of `List[Array]` for segment-wise strain bases.

### DIFF
--- a/src/soromox/systems/tendon_actuated_pcs.py
+++ b/src/soromox/systems/tendon_actuated_pcs.py
@@ -38,7 +38,7 @@ class TendonActuatedPCS(PCS):
     num_strains : int
         Total number of strain components (6 * num_segments).
     B_xi : Array
-        Basis matrix for projecting active strains (num_active_strains, num_active_strains).
+        Basis matrix for projecting active strains (6 * num_segments, num_active_strains).
     xi_ref : Array
         Reference strain (reference configuration) of the robot.
     num_gauss_points : int
@@ -213,7 +213,8 @@ class TendonActuatedPCS(PCS):
         true.
 
         Args:
-            tendon_routing_params (Dict[str, Array]): parameters of the tendons (6,)
+            tendon_routing_params : Dict[str, Array]
+                Dictionary of arrays of length n_actuators representing the tendon parameters.
 
         Returns:
             flag (bool): True if any of the tendons is out of the body
@@ -409,7 +410,8 @@ class TendonActuatedPCS(PCS):
             position of tendon (i.e., its attachement point).
 
             Args:
-                single_tendon_routing_params (Dict[str, Array]): parameters of one tendon (6,)
+                tendon_routing_params : Dict[str, Array]
+                    Dictionary of arrays of length n_actuators representing the tendon parameter for tendon k.
                 q (Array): generalized coordinates of shape (num_active_strains,)
                 s (Array): point coordinate along the robot ()
 
--- a/src/soromox/systems/tendon_actuated_pcs.py
+++ b/src/soromox/systems/tendon_actuated_pcs.py
@@ -278,62 +278,6 @@ class TendonActuatedPCS(PCS):
         return updated_self
 
     @eqx.filter_jit
-    def _local_actuation_basis(self, q: Array, s: Array) -> Array:
-        """
-        Compute the local actuation basis at current abscissa s.
-
-        Args:
-            q (Array): generalized coordinates of shape (num_active_strains,)
-            s (Array): abscissa points (num_gauss_points,)
-
-        Returns:
-            Phi_a_s (Array): actuation basis at s of shape (6 * num_segments, num_actuators)
-        """
-
-        def Phi_a_ki(single_tendon_routing_params: Dict[str, Array], i: Array):
-            """
-            Compute the actuation basis of one tendon at the segment i of the robot.
-
-            Args:
-                single_tendon_routing_params (Dict[str, Array]): parameters of one tendon (6,)
-                i (Array): index of the segment ()
-
-            Returns:
-                Phi_a_ki (Array): actuation basis of one tendon at the i-th segment of shape (6,)
-            """
-            attachment_segment_idx = single_tendon_routing_params["idx_seg_att"]  # ()
-            cond = attachment_segment_idx >= i  # ()
-
-            xi_j = jnp.reshape(xi, (6, self.num_segments), order="F")[
-                :, i
-            ]  # strains of segment j (6,)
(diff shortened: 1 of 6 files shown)


src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -39,8 +39,10 @@ class TendonActuatedPCS(PCS):
         Total number of strain components (6 * num_segments).
     B_xi : Array
         Basis matrix for projecting active strains (num_active_strains, num_active_strains).
-    B_xi_segments : Array
-        Basis matrix for projecting active strains, by segments (num_segments, num_active_strains, num_active_strains).
+    B_xi_segments : List[Array]
+        Segment-wise strain bases. Each entry has shape (6, num_active_strains_segment_i).
+    B_xi_segments_dense : Array
```
Michele-Martini: Is "B_xi_segments_dense" used somewhere? I cannot see it in this script. Is it intended to be called by the user on demand for inspection purposes?

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -39,8 +39,10 @@ class TendonActuatedPCS(PCS):
         Total number of strain components (6 * num_segments).
     B_xi : Array
         Basis matrix for projecting active strains (num_active_strains, num_active_strains).
-    B_xi_segments : Array
-        Basis matrix for projecting active strains, by segments (num_segments, num_active_strains, num_active_strains).
+    B_xi_segments : List[Array]
```
Michele-Martini: Also this I cannot see it being used. It is not stored as an attribute anymore either.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -400,24 +368,22 @@ def A_point_j(j: Array):
...
 
                 Returns:
-                    A_j (Array): local actuation matrix of shape (num_active_strains, num_actuators).
+                    A_j (Array): local actuation matrix of shape (6 * num_segments, num_actuators).
                 """
                 Xs_j = Xs_scaled[j]
                 Ws_j = Ws_scaled[j]
                 Phi_a_j = self._local_actuation_basis(q, Xs_j)
-                A_j = B_xi_i.T @ Phi_a_j  # A_s = B_xi.T @ Phi_a
+                A_j = Phi_a_j  # A_s = B_xi.T @ Phi_a
```
Michele-Martini: This mean that the Phi_a_j acts on all the strains of the robot (6 * num_segments), as the previous mask B_xi_i was 0 for the strains that are not in the current segment i. This could lead to sum the same j-th gaussian points multiple times when below we do A = jnp.sum(A_blocks, axis=(0, 1)).

src/soromox/systems/tendon_actuated_pcs.py
mstoelzle: You are right - thanks for spotting!

src/soromox/systems/tendon_actuated_pcs.py
mstoelzle: Sorry, I don't quite understand your doubt yet. 

- Do you mean that this code would also count contributions of strains that are not active in a specific segment?
- Why would the same Gaussian point contribute twice for the same segment with this new setup while it didn't before?

Could you please clarify and explain a bit more?

Michele-Martini: Of course, sorry for being unclear. Probably I am missing  something, but the reason why we had the B_xi_i before was to "mask out" the strains that are not in the current i-th segment. Basically, the function A_segment_i returns A_i, which is computed by vmapping A_point_j over the gaussian points of segment i. However, A_j is a matrix (6 * num_segments, num_actuators) as it derives from _local_actuation_matrix that computes the whole Phi_a_s. But if a tendon is attached to the tip of a robot with 2 segments, a value of s that is in the second segment (s > L/2 for example, if the 2 segments are equally long) will still have the contribution of the part of the tendon that is in the first segment. This is propagated in A_j and then in A_i and then in the sum A_blocks. That's why we multiplied in A_j for the B_xi_i that masked out the values of other segments != i. I understand that it is not a clear code structure, but this choice was made to avoid vertical concatenation of Phi_a of shape (6, num_actuators).

mstoelzle:
@Michele-Martini Can you please review this revised version of the PR ASAP?

Michele-Martini (approved):

PR #23, opened 2025-11-04
Statics

Michele-Martini:
Tentative approach to solve the statics of the system by solving the root finding problem of the statics equation with Newton-Raphson solver.

### DIFF
--- a/src/soromox/systems/dynamical_system.py
+++ b/src/soromox/systems/dynamical_system.py
@@ -1,6 +1,6 @@
 __all__ = ["DynamicalSystem"]
 import equinox as eqx
-from jax import Array, jit, lax
+from jax import Array, jit, lax, vmap
 from jax import numpy as jnp
 from pathlib import Path
 from typing import Callable, Dict, List, Tuple, Type, Union, Optional
@@ -15,6 +15,9 @@ from diffrax import (
     ConstantStepSize,
     AbstractSolver,
 )
+
+# Optimistix (for root finding)
+import optimistix as optx
 
 
 class DynamicalSystem(eqx.Module):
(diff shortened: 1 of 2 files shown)


src/soromox/systems/pendulum.py
```diff
@@ -692,6 +692,50 @@ def forward_dynamics(
         qdd = jnp.linalg.solve(B, rhs)
         return jnp.concatenate([qd, qdd])
 
+    @eqx.filter_jit
+    def statics_residual(
```
mstoelzle, line 696: Maybe for consistency, this could be renamed to `forward_statics`? Is this an expression that is sometimes used?

src/soromox/systems/pendulum.py
```diff
@@ -692,6 +692,50 @@ def forward_dynamics(
...
+
+        if u is None:
+            u = jnp.zeros((self.num_actuators,))
+        if tau_ext is None:
+            tau_ext = jnp.zeros((self.num_links,))
+
+        G = self.gravitational_force(q)
+        tau_el = self.elastic_force(q)
+        tau_u = self.actuation_force(q, u)
+        residual = tau_el + G - tau_u - tau_ext
```
mstoelzle, line 735: I think it would make sense to switch the sign here to match the convention of the forward dynamics.

src/soromox/systems/pendulum.py
```diff
@@ -692,6 +692,50 @@ def forward_dynamics(
...
+            raise ValueError("actuation_args must be None, (u,), or (u, tau_ext)")
+
+        if u is None:
+            u = jnp.zeros((self.num_actuators,))
+        if tau_ext is None:
+            tau_ext = jnp.zeros((self.num_links,))
+
+        G = self.gravitational_force(q)
+        tau_el = self.elastic_force(q)
+        tau_u = self.actuation_force(q, u)
```
mstoelzle, line 734: Do you think it would make sense to add a `potential_force` method to `DynamicalSystem` that sums the elastic and the gravitational force?

src/soromox/systems/dynamical_system.py
```diff
@@ -99,3 +102,36 @@ def resolve_upon_time(
         qs, qds = jnp.split(y_out, 2, axis=1)
 
         return ts, qs, qds
+
+    def solve_statics(
```
mstoelzle, line 106: jit missing

src/soromox/systems/dynamical_system.py
```diff
@@ -99,3 +102,36 @@ def resolve_upon_time(
...
+        Returns:
+            sol (optx.Solution): Solution object of the root finder.
+        """
+        if u is None:
+            u = jnp.zeros((self.num_actuators,))
+        if tau_ext is None:
+            tau_ext = jnp.zeros((self.num_links,))
+        
+        sol = vmap(
+            lambda fn, s, y0, args: optx.root_find(fn, s, y0, args, max_steps=256, throw=False),
```
mstoelzle, line 133: The `max_steps` needs to be a parameter of the function.

mstoelzle (changes requested):
The reasons why I am currently a bit sceptical of including a statics solver in the SoRoMoX main library are the following:

- As seen by the discussion we are having on Teams, you can formulate the problem in quite a few different ways. You can assume `u` to be given and identify `q` and vice-versa. Also, you can add the mapping into task space, and, for example, assume the task space position to be given. Finally, you can also define the task space in infinite different ways.
- There are quite a lot of different ways (e.g., gradient descent, nonlinear least squares, closed-form solutions in some cases, forward rollouts, etc.) how you can solve this problem from an optimization point of view and the same strategy might not work well for a wide class of systems and system parameters. 

Alternatively, we could just add an example for solving the statics.

mstoelzle:
Closing this PR for now as solving the statics is a very complex problems settings with not a single solution, but instead many possible approaches, each with various advantages and disadvantages. Therefore, it is better in my opinion to include the statics in examples instead of the core package.

PR #24, opened 2025-11-05
Bugfixes for `resolve_upon_time` and `TendonActuatedPendulum`

Michele-Martini:
This pull request introduces several improvements and corrections to the tendon-actuated pendulum simulation and its underlying system models. The main changes enhance the tendon length/displacement calculations, improve simulation and video output settings, and clarify documentation and code structure. Below are the most important changes grouped by theme.

### Tendon Length and Displacement Calculations

* Added new methods to `TendonActuatedPendulum` for computing active/passive tendon displacement and total length, with clear separation between displacement and length calculations. The initial displacement is now handled separately from total length, and the logic for calculating these values is clarified and made more robust. (`src/soromox/systems/tendon_actuated_pendulum.py`)
* Updated docstrings and variable naming throughout the tendon model to consistently refer to "displacement" rather than "length" for initial tendon configuration, improving clarity and correctness. (`src/soromox/systems/tendon_actuated_pendulum.py`) [[1]](diffhunk://#diff-445b713c2772619ff10413abb52d037f02afe5a1a15ed07374f0e32461722f7fL102-R102) [[2]](diffhunk://#diff-445b713c2772619ff10413abb52d037f02afe5a1a15ed07374f0e32461722f7fL169-R169)

### Simulation and Video Output Improvements

* Refined video rendering in the pendulum simulation: increased resolution, improved color schemes, added options for drawing target positions and tendons, and fixed video path handling. Also corrected video frame rate calculation and improved drawing logic for tendons. (`examples/simulation/pendulum/simulate_tendon_actuated_pendulum.py`) [[1]](diffhunk://#diff-ff6cd0a8d34fe1660673839597db33ec1ef8a8c0f7b8b90de61543cb40770ddeL38-R169) [[2]](diffhunk://#diff-ff6cd0a8d34fe1660673839597db33ec1ef8a8c0f7b8b90de61543cb40770ddeL192-R254) [[3]](diffhunk://#diff-ff6cd0a8d34fe1660673839597db33ec1ef8a8c0f7b8b90de61543cb40770ddeL206-R268)
* Added print statements to output the computed tendon lengths/displacements at the initial configuration for better debugging and transparency during simulation runs. (`examples/simulation/pendulum/simulate_tendon_actuated_pendulum.py`)

### Simulation Parameter Handling

* Improved time step and save interval logic for simulation: clarified documentation on `t0`, `t1`, and `saveat_ts`, and adjusted the simulation end time to ensure inclusion of the final time step. (`src/soromox/systems/dynamical_system.py`) [[1]](diffhunk://#diff-01e0052383ed2a1884d67c784069cc19a09b5253ab5ae23fd9e3ed1556bc4522L47-R53) [[2]](diffhunk://#diff-01e0052383ed2a1884d67c784069cc19a09b5253ab5ae23fd9e3ed1556bc4522L87-R87)

### Jacobian and Dynamics Corrections

* Corrected the shape and computation of Jacobian matrices in actuation/conformation conversions to ensure proper matrix dimensions and clarify linearity assumptions. Also fixed a typo in a comment about Jacobian inversion. (`src/soromox/systems/tendon_actuated_pendulum.py`) [[1]](diffhunk://#diff-445b713c2772619ff10413abb52d037f02afe5a1a15ed07374f0e32461722f7fL481-R537) [[2]](diffhunk://#diff-445b713c2772619ff10413abb52d037f02afe5a1a15ed07374f0e32461722f7fL494-R550) [[3]](diffhunk://#diff-445b713c2772619ff10413abb52d037f02afe5a1a15ed07374f0e32461722f7fL522-R578)
* Fixed the computation of the actuation space inertia matrix to remove an unnecessary matrix inversion, ensuring correct dynamics calculations. (`src/soromox/systems/tendon_actuated_pendulum.py`)

These changes collectively improve the accuracy, usability, and maintainability of the tendon-actuated pendulum simulation and its supporting system models.


src/soromox/systems/dynamical_system.py
```diff
@@ -84,7 +84,7 @@ def resolve_upon_time(
             terms=term,
             solver=solver,
             t0=t0,
-            t1=t1,
+            t1=t1 + dt,
```
mstoelzle: Does this generally hold? Or is it more an issue of the `saveat_ts` that you provide? To me, it looks like a modification that shouldn't be necessary if `save_ts` is properly constructed and `diffeqsolve` is working as expected.

examples/simulate_tendon_actuated_pendulum.py
mstoelzle, line 1: Please put the file into the correct examples folder. Maybe, you first need to merge `origin/main` into your local branch.

src/soromox/systems/tendon_actuated_pendulum.py
```diff
@@ -300,6 +300,20 @@ def update_tendon_params(
...
+        """
+        Compute the length of the active tendons.
+
+        Args:
+            q (Array): Joint angles, shape (N,) [rad]
+
+        Returns:
+            l_a (Array): length of the active tendons, shape (Na,) [m]
+        """
+        l_a = self.R_at @ q
```
mstoelzle: Would it make sense to also add a constant offset here, that would be specified as a system parameter? I am thinking here of initializing the tendon length by default using some kind of heuristic, such as the length of the pendulum to the attachment point of the tendon. I think when people read-out the tendon length, they could be slightly confused and think that there must be a bug.

mstoelzle (changes requested):

src/soromox/systems/dynamical_system.py
Michele-Martini: It holds when save_dt is specified instead of the save_ts array. But you are right, it may not hold when save_ts is specified instead. In that case, we would need to remove the + save_dt from saveat_ts = jnp.arange(t0, t1 + save_dt, save_dt), as this leads the saveat_ts array to finish at t1 while the diffeqsolve finishes at t1 - dt

src/soromox/systems/tendon_actuated_pendulum.py
Michele-Martini: Oh you mean like length and displacement? Actually what I have described is the displacement of the tendon from the rest position, so the length is that + the initial length?

src/soromox/systems/dynamical_system.py
mstoelzle: I don't quite follow your thoughts yet. Is the current behavior the following (before your code changes)?

- The `save_ts` array contains as the last value `t1`.
- Diffeqsolve only integrates to `t1-save_ts`, but contains the same number of timesteps as `saveat_ts`.

Or am I missing something? Also, does currently the shape of `save_ts` match the shape of `qs`? This is super important to me that the array shapes are coherent.

src/soromox/systems/tendon_actuated_pendulum.py
mstoelzle: Yeah, exactly. Add the initial length of the tendon.

Michele-Martini: in this case it would mean that the thing about the geometry of the cable actually matters... I am not sure about putting it as system parameter, as the tendon length depends on our code rather than the user's specifications (even if they specify the routing R_at and R_pt). Can I suggest to rename it to displacement instead of length?

Michele-Martini: Maybe the length is easier to compute in the PCS case given that the user specifies also d', which we can integrate directly.

src/soromox/systems/tendon_actuated_pendulum.py
mstoelzle, line 174: @Michele-Martini I think the passive tendon length should not be initialized as zero by default. But instead, as the length of the pendulum until the attachment point of the tendon.

Michele-Martini: I see, but also based on my comment above, I think that this can be misleading, as the initial length is not the pendulum length at the attachment point, but rather depends on the radii of the pulleys and configuration q0. In any case, this parameter is intended to encode the pre-stretch of the spring, so it should be renamed accordingly, and at most we should add another parameter for the length. What do you think?

src/soromox/systems/tendon_actuated_pendulum.py
mstoelzle: Would computing the tendon length as "link length sum plus tendon displacement" not be a decent approximation of the actual tendon length?

Michele-Martini: aaa

I think it depends on the pulleys' radius. Maybe if they are small yes, it can be an approximation, right?

mstoelzle: Let's go with this approximation for now, and then we can see :)

src/soromox/systems/dynamical_system.py
Michele-Martini: I see, maybe I rushed too much on this. Before the changes, by only passing save_dt to the resolve_upon_time method, I got an error that saveat.ts was not between t0 and t1. By looking at the code, I saw that the arange included t1 by adding the last save_dt (only in the case when you specified save_dt).
By adding a + dt to t1 in diffeqsolve it solves this, and this leads qs to have the same shape of saveat_ts when constructed via the save_dt parameter (before the changes)

Michele-Martini: Sorry if I am not replying directly to your questions, but because I could not retrieve those information. 
By specifiyng "save_dt", the original code constructs "saveat_ts" from t0 to t1 + save_dt, making t1 to be included (so the number of samples is ( (t1 - t0) / save_dt ) + 1.
In the original code however, specifying t0 and t1 in diffeqsolve leads to a number of samples = (t1 - t0) / dt, where t1 is not included (this was leading to the error).
I think we just need to agree if we want t1 to be included or not both in the saveat_ts and in the whole sequence in diffeqsolve.

src/soromox/systems/tendon_actuated_pendulum.py
mstoelzle: Let's implement the same logic as for the active tendon. In addition, the passive tendon should have a (separate) parameter relating to the prestretch/initial length of the tendon that can be used to compute the elastic forces of the passive tendon.

src/soromox/systems/dynamical_system.py
mstoelzle: Ok, thanks!

What is your opinion on if `t1` should be the final time of the simulation (i.e., it should be inclusive) or if the simulation should stop before `t1` (i.e., exclusive)?

Michele-Martini: Personally, when specifying an interval I tend to prefer being inclusive. However, the Python approach to lists and intervals is always exclusive, so maybe I think it is more important to stick to the most common conventions. What do you think?

mstoelzle: I asked ChatGPT about common conventions: https://chatgpt.com/share/690b6225-7810-8011-ae6f-14d7836b17a6

Based on this, I would propose to use the following convention in SoRoMoX:

- Let `save_ts` be given and let it contain `t1` as the last entry. Then, the simulation should return a state sequence of the same length as `save_ts` and the last state should correspond to the state at `t1` (i.e., inclusive).
- The first point implies that a user-provided `t1` should be overwritten by the last time step contained in a user-provided `save_ts`.
- If `save_ts` is not provided, create a `save_ts`, but using a semi-exclusive convention (like in the other step-based simulators): this should ensure that the last time step contained in the created`save_ts` should be smaller or equal to `t1`, where `save_ts` is obviously respecting the `save_dt`. 

Can you please take care of the following things:
- Document this convention in the docstring
- Implement the necessary code changes that ensure that the code *robustly* follows this convention.
- If `save_ts` is provided by the user, not the user-provided `t1` should be used, but instead a t1 should be set by the code that ensures that the `diffeqsolve` runs long enough to return all time steps contained in `save_ts`.

examples/simulate_tendon_actuated_pendulum.py
mstoelzle: Actually, I am seeing that this example already exists in the `main` branch.

Michele-Martini: Yes, I deleted that one, I'll push everything soon :)

src/soromox/systems/dynamical_system.py
```diff
@@ -44,13 +44,13 @@ def resolve_upon_time(
             u (Array, optional): Actuation/control input.
                 Default is None (no actuation).
             tau_ext (Array, optional): External forces/torques applied to the system.
-            t0 (float, optional): Initial time.
+            t0 (float, optional): Initial time of the simuation, included.
                 Default is 0.0.
-            t1 (float, optional): Final time.
+            t1 (float, optional): Final time of the simulation, included.
```
mstoelzle, line 49: Can you please expand on the explanation of the chosen convention a bit more. Otherwise, we are confused again in a few months :P

src/soromox/systems/dynamical_system.py
mstoelzle, line 81: I am missing the assert that the last entry of `save_ts` must be smaller-equal than t1

mstoelzle (changes requested):

src/soromox/systems/dynamical_system.py
Michele-Martini: Sorry I forgot to comment about this. I added but then there were issue during runtime. Basically, we cannot use the "assert" statement onto a variable which is traced dynamically, it raises an error. When assessing saveat_ts[-1] <= t1, t1 in the simulation is defined by t + ctrl_dt. t is traced during the lax.scan. There is no easy fix of this, to my (and Gemini's) knowledge. Since this is internally checked by diffeqsolve, I decided that we could skip and let diffrax to handle it. Do you agree or do you prefer alternative solutions?

mstoelzle: Ok, I see! Then, let's address it via a more extensive docstring that clearly states this constraint.

src/soromox/systems/pendulum.py
```diff
@@ -127,6 +131,7 @@ def __init__(
 
         self.K = jnp.asarray(params.get("K", jnp.zeros((n_q, n_q))))
         self.D = jnp.asarray(params.get("D", jnp.zeros((n_q, n_q))))
+        self.q_rest_k = jnp.asarray(params.get("q_rest_k", jnp.zeros((n_q,))))
```
mstoelzle: Please rename to `q_ref_k` to be consistent with the definition of reference strains for the continuum soft robots.

src/soromox/systems/tendon_actuated_pendulum.py
```diff
@@ -110,8 +110,10 @@ class TendonActuatedPendulum(Pendulum):
 
     R_at: Array
     A_at: Array
+    q_rest_at: Array
```
mstoelzle: Please rename to `q_ref_at` and `q_ref_pt` for consistency.

mstoelzle (changes requested):

mstoelzle (approved):

PR #30, opened 2025-11-18
Passive tendons in PCS

Michele-Martini:
Adding the feature of passive tendons in the PCS model. Each tendon is attached to a spring-damper unit.

### DIFF
--- a/src/soromox/systems/pcs.py
+++ b/src/soromox/systems/pcs.py
@@ -1739,9 +1739,12 @@ class PCS(DynamicalSystem):
         return D_full
 
     @eqx.filter_jit
-    def damping_matrix(self) -> Array:
+    def damping_matrix(self, q: Array) -> Array:
         """
         Compute the damping matrix of the robot.
+
+        Args:
+            q (Array): generalized coordinates of shape (num_active_strains,).
 
         Returns:
             D (Array): Damping matrix of shape (num_active_strains, num_active_strains).
@@ -2097,7 +2100,7 @@ class PCS(DynamicalSystem):
         B = self.B_xi.T @ B_full @ self.B_xi
         C = self.B_xi.T @ C_full @ self.B_xi
         G = self.B_xi.T @ G_full
-        D = self.damping_matrix()
+        D = self.damping_matrix(q)
         tau_el = self.elastic_force(q)
         tau_u = self.actuation_force(q, u)
 
--- a/src/soromox/systems/tendon_actuated_pcs.py
+++ b/src/soromox/systems/tendon_actuated_pcs.py
@@ -26,6 +26,8 @@ class TendonActuatedPCS(PCS):
         Number of segments (constant strain sections) along the robot.
     num_actuators : int
         Number of actuators (control inputs) for the robot.
+    n_p : int
+        Number of passive tendons in the robot.
     g0 : Array
         Initial pose of the robot base as an SE(3) transformation matrix.
     g : Array
@@ -62,6 +64,12 @@ class TendonActuatedPCS(PCS):
         As active_d_s for the passive tendons.
     passive_dd_s_ds : Callable
         As active_dd_s_ds for the passive tendons.
+    K_pt: Array
+        Stiffness matrix of the passive tendons (n_p, n_p).
+    D_pt: Array
+        Damping matrix of the passive tendons (n_p, n_p).
+    l_pt0: Array
+        Vector of the initial displacement of the passive tendons (n_p,).
 
     Notes:
     -----
@@ -83,13 +91,16 @@ class TendonActuatedPCS(PCS):
     - [Tendon lengths] Pustina, P., Della Santina, C., Boyer, F., De Luca, A., & Renda, F. (2024). Input decoupling of lagrangian systems via coordinate transformation: General characterization and its application to soft robotics. IEEE transactions on robotics, 40, 2098-2110.
     """
 
+    n_p: int
     active_tendon_routing_params: Dict[str, Array]
     active_d_s: Callable
     active_dd_s_ds: Callable
     passive_tendon_routing_params: Dict[str, Array]
     passive_d_s: Callable
     passive_dd_s_ds: Callable
-    n_p: int
+    K_pt: Array
+    D_pt: Array
+    l_pt0: Array
 
     def __init__(
         self,
@@ -100,6 +111,7 @@ class TendonActuatedPCS(PCS):
         active_tendon_routing_params: Dict[str, Array],
         passive_tendon_routing_basis: Optional[Dict[str, Callable]] = None,
         passive_tendon_routing_params: Dict[str, Array],
+        passive_tendon_params: Dict[str, Array],
         **kwargs,
     ):
         """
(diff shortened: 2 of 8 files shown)


Michele-Martini:
Not all features are implemented (like the spring and damping values). Still this is my proposal for how to approach this additional feature from the code factorization point of view.

src/soromox/systems/pcs.py
```diff
@@ -1739,10 +1739,13 @@ def _damping_full_matrix(self) -> Array:
         return D_full
 
     @eqx.filter_jit
-    def damping_matrix(self) -> Array:
+    def damping_matrix(self, q: Array) -> Array:
```
mstoelzle, line 1742: This is an issue as it breaks compatibility with the other systems. We have two options:

- Add `q` as a parameter to all `damping_matrix` methods of all systems
- Add the contribution of the passive tendon damping somewhere else.

Please let me know you opinion and thoughts which option would be better.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -313,9 +453,17 @@ def _local_actuation_basis(self, i: Array, xi_i: Array, s: Array, tendon_routing
         Phi_a_k = is_tendon_active * jnp.hstack([lie.tilde_SE3(d_s[:-1]) @ t, t])  # (6,)
 
         return Phi_a_k
+    
+    @eqx.filter_jit
+    def actuation_matrix(self, q: Array):
+        return self._actuation_matrix(q, self.active_tendon_routing_params, self.active_d_s, self.active_dd_s_ds)
+    
+    @eqx.filter_jit
+    def coupling_matrix(self, q: Array):
```
mstoelzle: This is not sufficiently verbose. I don't think its clear what you mean here with `coupling_matrix`.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -383,24 +534,32 @@ def A_point_j(j: Array):
...
     
     @eqx.filter_jit
-    def tendon_length(self, q: Array) -> Array:
+    def active_tendon_length(self, q: Array) -> Array:
+        return self._tendon_length(q, self.active_tendon_routing_params, self.active_d_s, self.active_dd_s_ds)
+    
+    @eqx.filter_jit
+    def passive_tendon_length(self, q: Array) -> Array:
+        return self._tendon_length(q, self.passive_tendon_routing_params, self.passive_d_s, self.passive_dd_s_ds)
+
```
mstoelzle, line 569: Please add an alias for the `tendon_length` method that routes to the `active_tendon_length` method:

It is sufficient to simply add the line

```python
tendon_length = active_tendon_length
```

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -533,15 +700,93 @@ def forward_kinematics_tendon_k(
...
+        # Stiffness of the springs attached to the passive tendons
+        A_pt = self.coupling_matrix(q)
+
+        D_full = D_body + A_pt @ self.D_pt @ A_pt.T
+        D = self.B_xi.T @ D_full @ self.B_xi
+        return D
+    
+    @eqx.filter_jit
+    def elastic_energy(self, q: Array) -> Array:
```
mstoelzle, line 774: Can you call the parent elastic_energy method for the first component that is the same as for a "normal" PCS system?

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -533,15 +700,93 @@ def forward_kinematics_tendon_k(
...
+        Compute the elastic forces of the robot.
+
+        Args:
+            q (Array): generalized coordinates of shape (num_active_strains,).
+
+        Returns:
+            tau_el (Array): Elastic force of shape (num_active_strains,).
+        """
+        # Stiffness of the body
+        K = self.stiffness_matrix()
```
mstoelzle: Instead of computing everything from scratch, I would suggest calling the parent method for the base terms.

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -533,15 +700,93 @@ def forward_kinematics_tendon_k(
...
+        Compute the elastic forces of the robot due to the springs attached
+        to the passive tendons.
+
+        Args:
+            q (Array): generalized coordinates of shape (num_active_strains,).
+
+        Returns:
+            tau_el_pt (Array): Elastic force of shape (num_active_strains,).
+        """
+        A_pt = self.coupling_matrix(q)
```
mstoelzle: Wrong indentation

src/soromox/systems/tendon_actuated_pcs.py
mstoelzle, line 1: You also need to edit the forward_dynamics method if I remember correctly. This is currently missing I think. Please double check.

mstoelzle (changes requested):
The general implementation strategy looks good. The most important question is with respect to the integration of the damping of the passive tendons / the different interface of the damping_matrix method.

src/soromox/systems/pcs.py
Michele-Martini: Yes this is one of the maing things. Personally, I prefer to avoid changing the method for all systems, but this means that we must change the forward dynamics method. The implemented approach now was to preserve the forward dynamics as much as possible. Alternatively, we should add a damping_matrix_passive_tendons here and add the two matrices. Maybe it would prefer this latter one. Do you agree?

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: It is actually a placeholder :) Usually you have the best ideas (in general) but especially about terminology. We could go for actuation_matrix_passive_tendons but this would not be symmetric wrt to the current actuation matrix, which I wouldnt change for consistency with the other systems. Coupling matrix is what is called in the murray book

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: I don't see where the wrong indentation is. Nonetheless, this method is redundant now, I am removing it.

src/soromox/systems/tendon_actuated_pcs.py
mstoelzle, line 21: @Michele-Martini Please add an extended discussion of the idea of using passive tendons. It will not be obvious to most users. Also, you could cite/reference the physical control perspective paper.

src/soromox/systems/tendon_actuated_pcs.py
mstoelzle: What about `passive_tendon_jacobian` or `jacobian_passive_tendon` and then return the transpose?

src/soromox/systems/tendon_actuated_pcs.py
```diff
@@ -533,15 +711,82 @@ def forward_kinematics_tendon_k(
...
+
+        Returns:
+            D_tot (Array): Total damping matrix of shape (num_active_strains, num_active_strains).
+        """
+        # Damping of the body
+        D = super().damping_matrix(q)
+
+        # Stiffness of the springs attached to the passive tendons
+        A_pt = self.jacobian_passive_tendon(q).T
+        D_pt_full = A_pt @ self.D_pt @ A_pt.T
```
mstoelzle: You could get rid of this "double-transpose" and instead work directly with the Jacobian. But more of a detail.

mstoelzle (approved):

src/soromox/systems/tendon_actuated_pcs.py
Michele-Martini: Oh, you are right!

mstoelzle (approved):

PR #40, opened 2025-12-09
New features

KWWnoob:
Added plot file and cbf case in \newfeature


mstoelzle (changes requested):
@KWWnoob 
- Please remove the `open3d_vis_V3.py` from the PR
- Please move theCBF example to `examples/control/safety_filter`
- Please add `cbfpy` as an `examples` optional dependency to the `pyproject.toml`
- Please import the operational-space impedance controller from the package (after #46 is merged) as the nominal controller.
- Please migrate to use `robot.rollout_closed_loop_to` instead of initializing your own `diffeqsolve`
- Please migrate to the `Open3DRenderer` after #45 is merged.

PR #50, opened 2025-12-29
Refactor Docs

mstoelzle:
Significant refactoring of the docs.


src/soromox/systems/pcs/tendon_actuated_pcs.py
```diff
@@ -28,55 +29,33 @@ class TendonActuatedPCS(PCS):
     below [Physical control]).
 
     Attributes:
-    ----------
-    num_segments : int
```
Michele-Martini, line 32: I thought it was useful to have the type and shape clearly stated. Why do you prefer not to have them?

mstoelzle: You see below that I still kept the docstrings. However, for proper formatting in the docs, a specific docstring style is required. Therefore, I changed the docstrings in quite a lot of system classes.

Michele-Martini (approved):

Danielfeliutalegon (approved):

PR #52, opened 2026-01-15
Rollout compatibility with jax.grad

Michele-Martini:
This PR addresses the issue of a killing error in the case that the "rollout_closed_loop_to" is wrapped within jax.grad.  The issue arises from calling "_compute_save_times" in the rollout method. In particular, the value t0 is traced when called from "jax.grad()" but the function "arange" requires the start value to be static. It is possible to solve this in different ways:
- make the start value float by casting directly "float()" on it and leaving "arange"
- [tentatively implemented now] make the start value float by casting directly "float()" on it and calling "linspace" instead of arange (linspace is more robust), by computing the number of samples
- leaving "_compute_save_times" as is and force the "float()" in the "rollout_closed_loop_to" method. Also, it should be declared in the argument types that only float values are allowed


mstoelzle (changes requested):
@Michele-Martini Thank you!

I am not sure about this change, though for two reasons:

- If I remember correctly, `jnp.linspace` can lead to some strange floating-point inprecisions, and `jnp.arange` leads to more consistent time steps.
- The reason that I removed the `float()` casting a few weeks ago is that `t0` and `t1` might be arrays instead of scalars (e.g., different starting times for each element of the batch array). When you apply the `float()` casting, you forbid such scenarios.

What I did in a similar scenario, where I had to apply vmap to the rollout methods for batched processing (https://github.com/tud-phi/soromox/blob/main/examples/simulation/pcs/simulate_batched_tendon_actuated_pcs.py), is to compute `save_ts` outside the `vmap`/`grad` code block. Would this not also work for your case?

Michele-Martini:
Your solution works! My intention was only to make things working in any case, for robustness. I am unsure about linspace leading to floating point imprecision, but if arange works is perfectly fine.
Also, how casting float() prevent those scenarios? Since we are not using lax.scan(), it seems that casting float() should not break things, as the cast only occurs when building the time vector. t0 and t1 can still be Arrays and vary but when those values are traced they must be static when passed to arange.

mstoelzle:
> Also, how casting float() prevent those scenarios?

Assume that `t0` and `t1` are array of shape `(B,)` instead of being scalars of shape `()`. Then, casting these variables to floats will clearly not work, as the `float(t0)` operation is only defined for scalars.

mstoelzle:
Closing this PR as it would cause other UX regressions (see points mentioned above).

PR #67, opened 2026-05-04
Added missing plot file

Michele-Martini:
Uploaded in the examples/optimization the missing plot file.


mstoelzle (changes requested):
I don't think I could run this plotting file as the data is missing. Please also add the data.

PR #83, opened 2026-06-13
Add McKibben-actuated UMArm model

mstoelzle:
## Summary

- Add cached robot parameter assets and the `McKibbenActuatedUMArm` articulated system.
- Add UMArm rendering support, simulation example, and API documentation.
- Update parameter-loading examples/tests for the cached-parameter layout.
- Implement closed-form McKibben actuation matrix evaluation with regression coverage against the potential-gradient expression.

## Validation

- `uv run pytest tests/systems/test_mckibben_umarm.py tests/systems/test_mckibben_umarm_reference.py`
- `uv run python -m py_compile src/soromox/systems/articulated/mckibben_actuated_umarm.py`

## Notes

The following local untracked files were left out of this PR: `force_goal_distance_plot.pdf`, `force_goal_distance_plot.png`, and `tests/systems/test_mckibben_umarm_reference.py`.


zuorunze (approved):
**Approve** — reviewed the McKibben-actuated UMArm model in depth and ran the test suite locally.

### Verified
- **Actuation physics is correct.** The pressure→joint-torque map `A(q)` is the gradient of the pressure-normalized McKibben potential `V = l(B²−l²)/(4πN²)`, and the closed-form length Jacobian (`_drot_x`/`_drot_y`) matches the actuator geometry exactly. `test_mckibben_actuation_matrix_matches_potential_gradient` confirms the closed form equals `jax.jacobian(_potential_per_pressure)` to 1e-10, and forward dynamics matches the dense equation of motion.
- **Full test suite passes locally:** `1003 passed, 1 skipped` (core + rendering + examples deps installed in an isolated worktree).
- **Refactor is safe:** the `hsa_params.py` deletion leaves no dangling imports (replaced by cached `.npz` assets + a `from_npz` loader with legacy-name fallbacks), and the shared `camera_config.py` change is backward-compatible (new `reference_transform` defaults to `None`).

### Minor, non-blocking notes
1. **CI `test` job was skipped** (0s) on the latest run — the automated tests never executed on CI; only `build` passed. Worth confirming the workflow trigger so tests actually run before merge. (I ran them locally in the meantime.)
2. **Validation command references a missing file.** The PR notes say `tests/systems/test_mckibben_umarm_reference.py` was intentionally left out, but the "Validation" section still lists it — so that command fails as written. The potential-gradient regression is already covered by the included `test_mckibben_umarm.py`, so coverage is fine; just update the description.
3. **`trimesh` is an undeclared dependency** — the viser guard requires it for `VISER_AVAILABLE=True`, but it isn't in any `pyproject.toml` extra. Pre-existing on `main` (not introduced here), but flagging since it makes the new `test_base_renderer.py` viser tests silently skip / fail depending on the environment.

Nice work — happy to approve.

PR #85, opened 2026-06-15
Compare HOCLF-only and HOCLF-HOCBF PCS controllers

KWWnoob:
## Summary
- Refactor the PCS CLF-CBF example into a CLI-driven comparison between HOCLF+HOCBF and HOCLF-only controllers.
- Use `robot.rollout_closed_loop_to` with `SystemState` for closed-loop simulation instead of direct Diffrax integration.
- Construct `PCSParams` directly from explicit parameter values and ignore generated force/distance plot artifacts.

## Testing
- `python3 -m py_compile examples/control/operational_space/control_pcs_with_cf_cbf_clf.py`


mstoelzle:
@KWWnoob Please cleanup and resolve the merge conflicts.

I had previously cleaned-up the script so it would be great if you can stay close to that. 

Also, maybe the controller type could be just a command line argument so we only need one script instead of copying code?

mstoelzle:
@KWWnoob Why are you adding two scripts? Shouldn't it be an update of the existing script (including the command line option which controller should be executed?)

Also, there seem to be some redundancies in the `.gitignore`

mstoelzle (changes requested):
@KWWnoob Now, I am missing the rendering & recording of the videos

Before you add this, please make sure to pull because I have cleaned-up the script (e.g., you reverted to the diffeqsolve integration instead the soromox integration again).

mstoelzle (approved):

mstoelzle (approved):

PR #90, opened 2026-06-16
Add Parallel RL Case Study

Hhannzch:
Hi Maxi, I have added the example code for parallel RL training, including the environment itself, the training file, the test file, and the saved models. Also, the dependencies are changed accordingly. Please check the code and let me know if anything needs to be changed!


examples/control/rl/checkpoints/ppo_model.zip
mstoelzle, line 1: What does this zip contain?

examples/control/rl/_repo_path.py
mstoelzle, line 1: You can remove this. Its fair to assume that the user has installed soromox via `pip install soromox` or `pip install -e .` (editable install) so the soromox package should be available (i.e., `from soromox import ..` should work)

examples/control/rl/checkpoints/env_vecnormalize.pkl
mstoelzle, line 1: Consider renaming the `soromox_models` folder. For example, `checkpoints` or `state_dicts` or something similiar.

examples/parallel_rl/parallel_soromox_env.py
```diff
@@ -0,0 +1,739 @@
+"""Parallel SoRoMoX tracking environment for Stable-Baselines3.
+
+This module exposes a JAX-vectorized VecEnv for tendon-actuated SoRoMoX arm
+tracking tasks. Each vectorized environment receives an independent target-ball
+trajectory at reset time, while all environments are stepped in one batched JAX
+call for efficient reinforcement-learning training and evaluation.
+"""
+
+from __future__ import annotations
```
mstoelzle: Is this necessary?

examples/control/rl/test.py
```diff
@@ -0,0 +1,202 @@
...
+    from ._repo_path import PARALLEL_RL_DIR, ensure_repo_src_on_path
+else:
+    from _repo_path import PARALLEL_RL_DIR, ensure_repo_src_on_path
+
+
+ensure_repo_src_on_path()
+
+import jax
+import numpy as np
+if __package__:
```
mstoelzle, line 20: Please double-check and ensure that this imports works independently from the user's workdir (i.e., independently where the user has `cd`-ed).

examples/control/rl/checkpoints/ppo_model.zip
mstoelzle: The file size doesn't seem very large. Does it contain many files?

mstoelzle (changes requested):
- Please put the code in `examples/control/rl`
- I am missing the script(s) to create the plot and the renderings

examples/control/rl/checkpoints/ppo_model.zip
Hhannzch: Hi @mstoelzle , it is the trained PPO model and the corresponding environment vectorization model. I renamed them to make them more clear

.gitignore
```diff
@@ -139,6 +139,7 @@ dmypy.json
 .DS_Store
 *.mp4
 *.gif
+!examples/control/rl/visualizations/parallel_track_video.mp4
```
mstoelzle: If the video is not too large, you don't need to git-ignore it but you can commit it instead

Hhannzch: Thank you @mstoelzle . I re-change this part

mstoelzle (approved):

PR #93, opened 2026-06-19
Add Standardization Folder

Michele-Martini:
Added a dedicated folder to store and process data for all the case studies presented in the paper.


paper_reproducibility/README.txt
```diff
@@ -0,0 +1,16 @@
+This folder aims to be a self-contained storage for the data and plots shown in the SoRoMoX paper. The suggested programming language is Python. The proposed structure is the following:
+
+- example_casestudy1/
+	-> code/    -->	 contains the .py files that reads from the data/ folder to process and generate the plots
+	-> data/    -->  contains the raw data in .csv, .npy, or .pkl format that are processed to generate the plots
+	-> images/  -->  contains the plots of the current case study in .pdf format preferably
```
mstoelzle: I would rename `images` to `outputs` as they could also contain videos in `.mp4` format.

paper_results/README.txt
```diff
@@ -0,0 +1,16 @@
+This folder aims to be a self-contained storage for the data and plots shown in the SoRoMoX paper. The suggested programming language is Python. The proposed structure is the following:
+
```
mstoelzle, line 2: What is the parent folder that contains these case study folders?

paper_reproducibility/README.txt
```diff
@@ -0,0 +1,16 @@
...
+
+- example_casestudy2/
+	-> code/
+	-> data/
+	-> images/
+
+- ...
+
+- final_figures/    -->  contains the final plots used in the paper
```
mstoelzle: Maybe name it `paper_figures`? And how would the plots arrive here? We would copy-paste it from the example case study folders?

mstoelzle (changes requested):

paper_reproducibility/README.txt
Michele-Martini: I think so. It is simply a single final place where to find everything. But I thought it as a copy of some of the ones in the sub-folders

paper_results/README.txt
Michele-Martini: Do you mean the paper_reproducibility? It is just a place for reviewers or interested users with a self-explanatory name where to find what has been shown in the paper. It is just a proposal. You can rename it as you please :)

paper_reproducibility/README.txt
mstoelzle: Sounds good!

paper_results/README.txt
mstoelzle: I don't like the name `paper_reproducibility` that much. How about about `paper_artifacts`, `paper_results`, `examples/paper`, or `publication_artifacts`.

mstoelzle (approved):
LGTM

PR #94, opened 2026-06-22
Analytical derivatives

mohammedtarnini:
Added backward pass and obtained analytical derivatives.

- Fixed se3.py to use "coadjointbar_se3" and "pcs_SoftJointDifferentialKinematics".
- Added a serial inverse-dynamics backward pass returning `dID/dq` and `dID/dqd`.
- Added analytical forward-dynamics Jacobians:
  - `dyd/dy`
  - `dyd/du`
  - `dyd/dtau_ext`
- Added base PCS force derivative helpers for elastic, damping, and actuation terms.
- Added a validation file examples/simulation/pcs/test_analytical_derivative.py that compares analytical derivatives against JAX autodiff.

forward_dynamics_derivatives dqdd/dq:
rel error = 5.782666e-12
max abs error = 1.495151e-06

Codes can be further optimized for speed in the future, but they are giving correct values as per the test done. custom_jvp still not used and no comparison for speed against autodiff, but it should be feasible with the current state of the code.

### DIFF
--- a/examples/simulation/pcs/test_analytical_derivative.py
+++ b/examples/simulation/pcs/test_analytical_derivative.py
@@ -1,0 +1,182 @@
+import jax
+import jax.numpy as jnp
+
+jax.config.update("jax_enable_x64", True)
+
+from soromox.systems import PCS, PCSParams
+
+jnp.set_printoptions(
+    threshold=jnp.inf,
+    linewidth=jnp.inf,
+    formatter={"float_kind": lambda x: "0" if x == 0 else f"{x:.2e}"},
+)
+
+
+if __name__ == "__main__":
+    num_segments = 2
+    seed = 7212
+
+    rho = 1070 * jnp.ones((num_segments,))
+    segment_lengths = 1e-1 * jnp.ones((num_segments,))
+    damping_matrix = 1e-3 * jnp.diag(
+        (
+            jnp.repeat(
+                jnp.array([[1e0, 1e0, 1e0, 1e3, 1e3, 1e3]]), num_segments, axis=0
+            )
+            * segment_lengths[:, None]
+        ).flatten()
+    )
+    params = PCSParams(
+        base_pose=jnp.array([jnp.pi / 2, jnp.pi / 2, 0.0, 0.0, 0.0, 0.0]),
+        length=segment_lengths,
+        radius=2e-2 * jnp.ones((num_segments,)),
+        density=rho,
+        gravity=jnp.array([0.0, 0.0, 9.81]),
+        young_modulus=2e3 * jnp.ones((num_segments,)),
+        shear_modulus=1e3 * jnp.ones((num_segments,)),
+        damping_matrix=damping_matrix,
+        reference_strain=jnp.tile(
+            jnp.array([0.0, 0.0, 0.0, 1.0, 0.0, 0.0]), num_segments
+        ),
+    )
+
+    robot = PCS(params=params)
+
+    key_q, key_qd, key_u, key_tau = jax.random.split(jax.random.PRNGKey(seed), 4)
+    q = 0.5 * jax.random.normal(key_q, (robot.num_dofs,))
+    qd = 0.2 * jax.random.normal(key_qd, (robot.num_dofs,))
+    u = 1e-2 * jax.random.normal(key_u, (robot.num_actuators,))
+    tau_ext = 1e-2 * jax.random.normal(key_tau, (robot.num_dofs,))
+    y = jnp.concatenate([q, qd])
+    t = jnp.array(0.0)
+
+    yd = robot.forward_dynamics(t, y, (u, tau_ext))
+    _, qdd = jnp.split(yd, 2)
+
+    analytical_dID_dq, analytical_dID_dqd = robot.inverse_dynamics_derivatives(
+        q, qd, qdd
+    )
+    analytical_dtau_el_dq = robot.elastic_force_derivative_q(q)
+    analytical_dtau_damp_dq, analytical_dtau_damp_dqd = robot.damping_force_derivatives(
+        q, qd
+    )
+    analytical_dtau_u_dq = robot.actuation_force_derivative_q(q, u)
+    analytical_dtau_u_du = robot.actuation_force_derivative_u(q)
+    analytical_dqdd_dq, analytical_dqdd_dqd = robot.forward_dynamics_derivatives(
+        q, qd, qdd, u
+    )
+    analytical_dqdd_du, analytical_dqdd_dtau_ext = (
+        robot.forward_dynamics_input_derivatives(q)
+    )
+    analytical_dyd_dy = robot.forward_dynamics_state_jacobian(t, y, (u, tau_ext))
+    analytical_dyd_dy_full, analytical_dyd_du, analytical_dyd_dtau_ext = (
+        robot.forward_dynamics_jacobians(t, y, (u, tau_ext))
+    )
+
+    autodiff_dID_dq = jax.jacfwd(
+        lambda q_arg: robot.inverse_dynamics_force(q_arg, qd, qdd)
+    )(q)
+    autodiff_dID_dqd = jax.jacfwd(
+        lambda qd_arg: robot.inverse_dynamics_force(q, qd_arg, qdd)
+    )(qd)
+    autodiff_dtau_el_dq = jax.jacfwd(lambda q_arg: robot.elastic_force(q_arg))(q)
+    autodiff_dtau_damp_dq = jax.jacfwd(lambda q_arg: robot.damping_matrix(q_arg) @ qd)(
+        q
+    )
+    autodiff_dtau_damp_dqd = jax.jacfwd(
+        lambda qd_arg: robot.damping_matrix(q) @ qd_arg
+    )(qd)
+    autodiff_dtau_u_dq = jax.jacfwd(lambda q_arg: robot.actuation_force(q_arg, u))(q)
+    autodiff_dtau_u_du = jax.jacfwd(lambda u_arg: robot.actuation_force(q, u_arg))(u)
+    autodiff_dqdd_dq = jax.jacfwd(
+        lambda q_arg: jnp.split(
+            robot.forward_dynamics(t, jnp.concatenate([q_arg, qd]), (u, tau_ext)),
+            2,
+        )[1]
+    )(q)
+    autodiff_dqdd_dqd = jax.jacfwd(
+        lambda qd_arg: jnp.split(
+            robot.forward_dynamics(t, jnp.concatenate([q, qd_arg]), (u, tau_ext)),
+            2,
+        )[1]
+    )(qd)
+    autodiff_dqdd_du = jax.jacfwd(
+        lambda u_arg: jnp.split(robot.forward_dynamics(t, y, (u_arg, tau_ext)), 2)[1]
+    )(u)
+    autodiff_dqdd_dtau_ext = jax.jacfwd(
+        lambda tau_ext_arg: jnp.split(
+            robot.forward_dynamics(t, y, (u, tau_ext_arg)),
+            2,
+        )[1]
+    )(tau_ext)
+    autodiff_dyd_dy = jax.jacfwd(
+        lambda y_arg: robot.forward_dynamics(t, y_arg, (u, tau_ext))
+    )(y)
+    autodiff_dyd_du = jax.jacfwd(
+        lambda u_arg: robot.forward_dynamics(t, y, (u_arg, tau_ext))
+    )(u)
+    autodiff_dyd_dtau_ext = jax.jacfwd(
+        lambda tau_ext_arg: robot.forward_dynamics(t, y, (u, tau_ext_arg))
+    )(tau_ext)
+
+    comparisons = [
+        ("inverse_dynamics_derivatives dID/dq", analytical_dID_dq, autodiff_dID_dq),
+        ("inverse_dynamics_derivatives dID/dqd", analytical_dID_dqd, autodiff_dID_dqd),
+        ("elastic_force_derivative_q", analytical_dtau_el_dq, autodiff_dtau_el_dq),
+        (
+            "damping_force_derivatives dtau_damp/dq",
+            analytical_dtau_damp_dq,
+            autodiff_dtau_damp_dq,
+        ),
+        (
+            "damping_force_derivatives dtau_damp/dqd",
+            analytical_dtau_damp_dqd,
+            autodiff_dtau_damp_dqd,
+        ),
+        (
+            "actuation_force_derivative_q",
+            analytical_dtau_u_dq,
+            autodiff_dtau_u_dq,
+        ),
+        (
+            "actuation_force_derivative_u",
+            analytical_dtau_u_du,
+            autodiff_dtau_u_du,
+        ),
+        ("forward_dynamics_derivatives dqdd/dq", analytical_dqdd_dq, autodiff_dqdd_dq),
+        (
+            "forward_dynamics_derivatives dqdd/dqd",
+            analytical_dqdd_dqd,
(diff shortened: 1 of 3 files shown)


src/soromox/utils/lie_algebra/se3.py
```diff
@@ -128,149 +224,1144 @@ def _compute_V_inv_general(args):
...
+        theta <= eps,
+        lambda _: _S_series_branch(),
+        lambda _: _S_general_branch(theta),
+        operand=None,
+    )
+
+    return (Omega, Z, g, T, S, Sd, f, fd, adjOmegap, dSdq_qd, dSdq_qdd, dSddq_qd)
+
+
+def pcs_SoftJointDifferentialKinematics(
```
mstoelzle, line 1046: We recently clean-up the re-organized the lie algebra functions a bit. Specifically, we tried to separate more between core SO3 and SE3 methods and constant-strain-specific solutions. This function would likely be better suited for `constant_strain.py`?

src/soromox/utils/lie_algebra/se3.py
```diff
@@ -128,149 +224,1144 @@ def _compute_V_inv_general(args):
...
+    eps, H, Phi, xi_star, q, qd, qdd
```
mstoelzle, line 1047: Please make the argument names and argument order more coherent with the rest of the codebase. For example, `eps` usually comes last. Also, we (likely) use `strain_basis` instead of `Phi`, etc.

src/soromox/utils/lie_algebra/se3.py
```diff
@@ -128,149 +224,1144 @@ def _compute_V_inv_general(args):
...
+) -> tuple[
+    Array, Array, Array, Array, Array, Array, Array, Array, Array, Array, Array, Array
+]:
```
mstoelzle, line 1050: Please add full docstrings, including list of arguments and return values. AI agents can also help you with that.

src/soromox/utils/lie_algebra/se3.py
```diff
@@ -128,149 +224,1144 @@ def _compute_V_inv_general(args):
...
+    k = Omega[:3]
+    kd = Omegad[:3]
+    theta = _rotational_strain_magnitude(Omega, eps)
+    thetad = (kd @ k) / theta
+
+    Omegahat = hat_SE3(Omega)
+    Omegahatp2 = Omegahat @ Omegahat
+    Omegahatp3 = Omegahatp2 @ Omegahat
+
+    adjOmegap = adjOmegap.at[0:6, :].set(adjoint_se3(Omega))
```
mstoelzle, line 716: As mentioned during the last meeting, I would prefer a stacking/concatenation of array instead of first initializing and then overwriting blocks of the matrix as this is sometimes slower.

src/soromox/utils/lie_algebra/se3.py
```diff
@@ -128,149 +224,1144 @@ def _compute_V_inv_general(args):
...
+                (4 - 4 * costheta - t1) / (2 * tp2),
+                (4 * theta - 5 * sintheta + t2) / (2 * tp3),
+                (2 - 2 * costheta - t1) / (2 * tp4),
+                (2 * theta - 3 * sintheta + t2) / (2 * tp5),
+            ]
+        )
+
+        fd = jnp.array(
+            [
+                t3 / (2 * tp3),
```
mstoelzle, line 787: Most/all of these vairables are already arrays. Therefore, we shouldn't use `jnp.array` but instead `jnp.concatenate` or `jnp.stack` for constructing the _larger_ array (`fd` in this case)

src/soromox/utils/lie_algebra/se3.py
```diff
@@ -128,149 +224,1144 @@ def _compute_V_inv_general(args):
...
+
+    def _S_series_branch() -> tuple[Array, Array, Array]:
+        dSdq_qd = TZd
+        dSdq_qdd = T @ Zdd
+        dSddq_qd = Td @ Zd
+        # dSdq_qd = jnp.zeros_like(Z)
+        # dSdq_qdd = jnp.zeros_like(Z)
+        # dSddq_qd = jnp.zeros_like(Z)
+        for r in range(4):
+            for u in range(1, r + 2):
```
mstoelzle, line 860: Please benchmark this implementation against a `lax.scan` structure and see which one is faster in terms of compilation and execution time. If you manage to use `vmap` instead (for one of the loops), it would be even better.

src/soromox/utils/lie_algebra/se3.py
```diff
@@ -128,149 +224,1144 @@ def _compute_V_inv_general(args):
...
+
+    angtilde = tilde_SE3(ang)  # Tilde operator for angular part
+    lintilde = tilde_SE3(lin)  # Tilde operator for linear part
+
+    coadbar = -jnp.block([[angtilde, lintilde], [lintilde, jnp.zeros((3, 3))]])
+
+    return coadbar
+
+
+def gvs_SoftJointDifferentialKinematics_Z4(
```
mstoelzle, line 672: This function should go into a different file. I am not quite sure what to call it yet. I guess something that generalizes `constant_strain.py`. Therefore, would `variable_strain.py` make sense?

src/soromox/utils/lie_algebra/se3.py
mstoelzle, line 1: Just to double check: `pcs_SoftJointDifferentialKinematics`, `gvs_SoftJointDifferentialKinematics_Z4` are all methods relating to the kinematics (and not the dynamics), right? If not, we should consider removing them from the `lie_algebra` submodule.

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
         yd = jnp.concatenate([qd, qdd])
 
         return yd
+
+    @eqx.filter_jit
+    def _eta_etad_bodyframe(
```
mstoelzle, line 2670: Are you using this method somewhere?

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
...
+        """
+        J_local, Jd_local = self.jacobian_and_time_derivative_bodyframe(q, qd, s)
+
+        eta = J_local @ qd
+        etad = J_local @ qdd + Jd_local @ qd
+
+        return eta, etad, J_local, Jd_local
+
+    @eqx.filter_jit
+    def _pcs_soft_joint_differential_kinematics(
```
mstoelzle, line 2696: Please reflect on a better method name. For example, the class is already called `PCS` so we don't need to also include this in the method name. Also, it should be apparent that this is for a single segment. Maybe `_segment_differential_kinematics` or `_constant_strain_differential_kinematics`?

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
...
+            etad_bases[segment_indices],
+            R_B_bases[segment_indices],
+            L_B_bases[segment_indices],
+            Q_B_bases[segment_indices],
+        )
+
+        return R_B_ps, L_B_ps, Q_B_ps, Y_B_ps
+
+    @eqx.filter_jit
+    def _dSTdq_FC_pcs_step(self, i: Array, H: Array, q: Array, F_C: Array) -> Array:
```
mstoelzle, line 3454: Also, here `pcs` is not necessary in the method name. Also, let's try to identify a more semantically informative yet concise method name!

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
...
+                q,
+                qd_unit,
+                qdd_zero,
+            )
+            return F_C @ dS_dq_qd
+
+        return vmap(projected_wrench_row)(qd_basis)
+
+    @eqx.filter_jit
+    def _backward_pass_interval_kinematics(
```
mstoelzle, line 3487: Please add docstring

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
...
+            q,
+            qd,
+            qdd,
+        )
+        return dID_dq, dID_dqd
+
+    @eqx.filter_jit
+    def elastic_force_derivative_q(self, q: Array) -> Array:
+        """Return the analytical derivative ``d(tau_el) / dq``."""
+        del q
```
mstoelzle, line 3998: This `del` shouldn't be necessary

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
...
+        """Return ``dID/dq`` and ``dID/dqd`` from the PCS backward pass."""
+        dID_dq, dID_dqd, _, _, _, _, _, _ = self.inverse_dynamics_backward_pass(
+            q,
+            qd,
+            qdd,
+        )
+        return dID_dq, dID_dqd
+
+    @eqx.filter_jit
+    def elastic_force_derivative_q(self, q: Array) -> Array:
```
mstoelzle, line 3996: To be more consistent with `*_time_derivative*` and `*_spatial_derivative*`, maybe we could call these `*_configuration_derivative`

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
...
+        """Return the analytical derivative ``d(tau_el) / dq``."""
+        del q
+        return self.stiffness_matrix()
+
+    @eqx.filter_jit
+    def damping_force_derivatives(
```
mstoelzle, line 4002: I recommend `damping_force_state_derivative`

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
...
+        qd: Array,
+    ) -> tuple[Array, Array]:
+        """Return analytical derivatives of ``damping_matrix(q) @ qd``."""
+        return (
+            jnp.zeros((q.shape[0], q.shape[0]), dtype=q.dtype),
+            self.damping_matrix(q),
+        )
+
+    @eqx.filter_jit
+    def actuation_force_derivative_q(self, q: Array, u: Array) -> Array:
```
mstoelzle, line 4014: Same: `actuation_force_configuration_derivative`

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
...
+        """Return the analytical derivative ``d(tau_u) / dq``."""
+        del u
+        return jnp.zeros((q.shape[0], q.shape[0]), dtype=q.dtype)
+
+    @eqx.filter_jit
+    def actuation_force_derivative_u(self, q: Array) -> Array:
```
mstoelzle, line 4020: `actuation_force_input_derivative` or `actuation_force_actuation_derivative`

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
...
+        """Return the analytical derivative ``d(tau_u) / du``."""
+        return self.actuation_matrix(q)
+
+    @eqx.filter_jit
+    def forward_dynamics_derivatives(
```
mstoelzle, line 4025: Let's call this `forward_dynamics_state_derivative`.

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
...
+            M_C_steps,
+            N_C_steps,
+            F_C_steps,
+            P_S_steps,
+            U_S_steps,
+            V_S_steps,
+        )
+
+    @eqx.filter_jit
+    def inverse_dynamics_derivatives(
```
mstoelzle, line 3981: Let's call it `inverse_dynamics_state_derivative`

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
...
+        self,
+        q: Array,
+        qd: Array,
+        qdd: Array,
+        u: Array | None = None,
+    ) -> tuple[Array, Array]:
+        """
+        Return analytical ``dqdd/dq`` and ``dqdd/dqd``.
+
+        This is the unconstrained PCS equivalent of the MATLAB ``ODEJacobian``
```
mstoelzle, line 4035: I don't think the references to the MATLAB implementation should be here.

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
...
+    def actuation_force_derivative_u(self, q: Array) -> Array:
+        """Return the analytical derivative ``d(tau_u) / du``."""
+        return self.actuation_matrix(q)
+
+    @eqx.filter_jit
+    def forward_dynamics_derivatives(
+        self,
+        q: Array,
+        qd: Array,
+        qdd: Array,
```
mstoelzle, line 4029: Why do you require `qdd` as an input? This means that the user has to execute another evaluation of the forward dynamics first which causes a lot of redundant computation. Instead, this function could compute `qdd` in the same pass and return it as well.

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
...
+        H: Array,
+        q: Array,
+        qd: Array,
+        qdd: Array,
+    ) -> tuple[Array, Array, Array, Array, Array, Array, Array, Array, Array, Array]:
+        """
+        Return the local PCS interval terms used by the forward and backward passes.
+
+        ``eta_plus`` and ``etad_plus`` are expressed in the frame before the step,
+        matching the MATLAB recursion immediately before multiplying by
```
mstoelzle, line 2792: Remove the references to MATLAB

src/soromox/systems/pcs/pcs.py
```diff
@@ -2685,3 +2665,1489 @@ def forward_dynamics(
...
+        ``dqdd_dq = M \\ (dtau_dq - dID_dq)``
+        ``dqdd_dqd = M \\ (dtau_dqd - dID_dqd)``
+
+        For base PCS, ``dtau_dq = -K`` and ``dtau_dqd = -D`` because the
+        actuation matrix, stiffness, and damping are configuration independent.
+        """
+        if u is None:
+            u = jnp.zeros((self.num_actuators,), dtype=q.dtype)
+
+        mass_matrix = self.inertia_matrix(q)
```
mstoelzle, line 4047: Calling such functions like the `inertia_matrix` separately will be very expensive. Can you get it "for free" during the `inverse_dynamics_derivatives` pass?

examples/simulation/pcs/test_analytical_derivative.py
mstoelzle, line 1: Later, this content should go as pytest-compatible test functions into `tests/systems`.

mstoelzle (changes requested):
Thanks, @mohammedtarnini, for implementing this! This is a great step! And it is very exciting that the tests comparing against the autdiff jacobians already seem to pass! 🎉

I have some comments. Most of them are related to the goal of making the interface and style of the implementation coherent and consistent with the rest of the package.

Some overarching points for improvement:
- Please add full docstrings for all methods/functions that explains what the function does, lists all arguments and return values, including array shapes, physical units, assumptions, etc.
- Please add more code comments in general to explain the implementation and the code to make it easier to maintain and extend
- Please make the variable naming more consistent with the rest of the codebase
- Please make the argument order more consistent with the rest of the codebase

Also, please merge the current `main` branch into this branch and resolve the merge conflict with `se3.py`

Next, I would suggest adding some benchmarking code so we have an initial idea of how much faster or slower the analytical derivative is compared to an autodiff-based derivative?

PR #95, opened 2026-06-23
Consolidate examples and reproducible paper workflows

Michele-Martini:
## Summary

This PR has two related goals:

1. Establish a common publication style for figures across the paper.
2. Make the code, data, and outputs underlying paper results discoverable and reproducible.

The visual work is the primary user-facing change. Plotters for Sections IVa, IVb, Va, Vc, Vd, Ve, and Vf now share the same Matplotlib style foundation and use paper-oriented dimensions, typography, line treatment, labels, legends, and export settings.

The repository changes support that work by moving paper-specific workflows out of general examples and benchmark utilities into section-owned directories under `paper_results/`.

## Common paper figure style

This PR adds `paper_results/paper.mplstyle` and applies it across the revised paper plotters. The shared style defines the baseline serif and mathematical typography, grid appearance, legend frames, axis and tick treatment, default line width, and export bounding box.

Each plotter retains local control over figure-specific requirements such as physical size, aspect ratio, DPI, colors, and panel layout. This keeps the common layer small while aligning the visible conventions that should be consistent across the paper.

The revised plotters use:

- figure sizes specified in physical units for paper-column layouts;
- consistent serif and mathematical typography;
- explicit axis labels with physical units;
- deliberate colors and line styles for coordinates, controllers, references, and baselines;
- prominent reference trajectories where close tracking would otherwise obscure them;
- compact legends and layouts suitable for the manuscript;
- noninteractive CLI options for automated figure generation;
- paper-ready PDF outputs, with PNG or video outputs where the workflow needs them.

Section Vc receives additional plot-specific work:

- The configuration-space paper figure compares the controller trajectories directly and reports RMSE separately for the setpoint and trajectory-reference phases.
- RMSE panels identify the generalized configurations or strains represented by each value.
- The operational-space paper figure presents the partial-feedback-linearization rollout in a compact position/orientation layout.
- A separate saved-data workflow provides the full-versus-partial feedback-linearization analysis without coupling it to the canonical paper figure.
- Desired trajectories use increased line width and drawing order so they remain visible under closely matching actual trajectories.

The PR also adds composed paper outputs under `paper_results/final_outputs/` for the figures that combine or arrange section-level results.

## Paper reproduction structure

Paper workflows are organized by paper section. Each case keeps its generation or simulation code alongside its input or generated data and documented outputs. Reproduction commands and artifact mappings are provided by the root paper-results index and the section READMEs that are included:

```text
paper_results/<section>/
├── code/
├── data/
├── outputs/
└── README.md  # where the section has dedicated reproduction documentation
```

Paths are resolved from the owning script or case directory rather than the caller's working directory. Migrated CLIs expose input and output paths, create destination directories when needed, support noninteractive plotting, and protect existing outputs where appropriate.

Section data and paper outputs are stored as regular Git objects; trained Section Vf checkpoints use Git LFS. Validation runs use temporary destinations so they do not replace the committed paper artifacts.

### Section IVa: sequential CPU comparison

- Renames `paper_results/benchmarking_sequential_cpu/` to the paper-section-oriented `secIVa_benchmarking_sequential_cpu/`.
- Keeps the existing SoRoMoX, PyElastica, and SoRoSim generators and datasets with the owning paper result.
- Reworks the rollout comparison plotter to use the common paper style.
- Adds section documentation and refreshed paper outputs.

### Section IVb: parallel GPU rollouts

- Moves the batch-scaling generator and plotter from `tools/benchmarks/` into `paper_results/secIVb_parallel_rollouts_gpu/`.
- Moves the corresponding CSV data out of the top-level `benchmarks/batch_scaling/` directory.
- Adds standardized scaling, combined-scaling, and throughput figures.
- Leaves reusable system-method and derivative-path diagnostics in `tools/benchmarks/`.

### Section Va: system identification

- Moves the soft-tentacle parameter and residual-identification workflows from `examples/system_identification/` into `paper_results/secVa_system_identification/`.
- Adds the plotting workflow, supporting data, four paper figures, and reproduction documentation.

### Section Vc: model-based control

- Moves the configuration-space comparison from `examples/control/configuration_space/` into its own paper case.
- Moves the operational-space impedance example from `examples/control/operational_space/` into a separate paper case.
- Adds saved NPZ trajectories, case-specific plotting and rendering entry points, dedicated READMEs, and the committed `configuration_space_comparison.pdf`, `regulation_tracking_rmse.pdf`, and `operational_space_impedance_control.pdf` paper figures.
- Separates data generation, plotting, and rendering so figures and videos can be reproduced without rerunning simulations.

### Section Vd: control-gain optimization

- Moves the collocated and synergistic optimization workflows and their data from `examples/optimization/` into `paper_results/secVd_control_gain_optimization/`.
- Adds case-relative paths, CLI controls, documentation, and the standardized paper figure.

### Section Ve: safety-constrained control

- Replaces the single example entry point with separate generation, shared-control, plotting, and video-rendering modules under `paper_results/secVe_safety_constrained_control/`.
- Adds saved rollouts, system parameters, snapshots, and the paper plot outputs.

### Section Vf: parallel reinforcement learning

- Moves the RL environment, training, evaluation, policy rollout, checkpoints, and reward logs from `examples/control/rl/` into `paper_results/secVf_parallel_rl/`.
- Adds an offline video renderer, saved rollout and snapshot data, a standardized reward plot, and reproduction documentation.

## Examples and development benchmarks

The remaining examples stay focused on lightweight demonstrations. Their generated figures and videos are written to example-local artifact directories instead of the repository root, and `examples/README.md` documents the resulting structure.

The benchmark system registry is updated to use the current typed parameter APIs for Pendulum, articulated robot, planar PCS, PCS, and GVS systems. Tests cover each registered system and the GVS basis-order configuration. The outdated dictionary constructors are also present on the PR base, so this repairs an existing benchmark issue rather than a regression caused by the reorganization.

## Documentation and artifact handling

- Replaces `paper_results/README.txt` with a section and artifact index in `paper_results/README.md`.
- Updates the root README, examples guide, benchmark documentation, and changelog.
- Keeps trained Section Vf checkpoints in Git LFS; Section Vc data, figures, and video are regular Git objects.
- Removes generated caches and superseded benchmark images from their former locations.

## Validation

- Section Vc test suite: **70 passed**
- Focused migration and benchmark tests: **76 passed**
- Full local suite excluding the Viser live-mode shutdown case: **1191 passed, 1 deselected**
- Ruff passes for the changed Python files.
- Revised plotters were run noninteractively and checked for dimensions, panel structure, labels, legends, line visibility, and output creation.
- Migrated workflows were exercised from unrelated working directories to verify case-relative paths.
- Plotting tests use temporary schema-complete inputs and do not require Git LFS hydration in CI.
- Review follow-up tests cover automatic Open3D viewer shutdown and the disableable RL trajectory overlay.

## Reproduction limitations observed

- The UMArm Viser live-mode smoke test hangs during server-thread shutdown in the local environment.
- The Section Va neural-network residual prediction does not reproduce the committed RMSE, although the geometry and optimal-residual paths agree to numerical precision.
- The Section Vd collocated optimization develops steady-state velocity and subsequently NaNs under the committed scientific settings.
- Full GPU timing reproduction, MATLAB/SoRoSim execution, full RL retraining, and interactive video integrations require hardware or external runtimes that were not available for the complete validation run.

No scientific parameters, gains, seeds, datasets, or benchmark definitions were changed merely to force agreement.


mstoelzle:
@Michele-Martini Can you please write down here the the concept for applying the color palette throughout the figures? So, which color is used to represent "initial/unoptimized/unsafe behavior", which one "optimized/safe behavior", which one "x-coordinates", which one "y-coordinates", which one for the control reference, which one for the ground-truth, etc.? Same with the linestyle.

Michele-Martini:
```
COLORS = {
    "pre_opt_1": "#006BA6",
    "pre_opt_2": "#0496FF",
    "post_opt_1": "#D81159",
    "post_opt_2": "#8F2D56",
    "post_opt_3": "#FFBC42",
    "backbone_opt": "#7B2CBF",
    "backbone_init": "#0ead69",
    "ground_truth": "#FFBC42",
    "x_t": "#2a9d8f",
    "y_t": "#e9c46a",
    "z_t": "#e76f51",
}
AXIS = {
    "x_axis_label": r"$x$ $\mathrm{[m]}$",
    "y_axis_label": r"$y$ $\mathrm{[m]}$",
    "z_axis_label": r"$z$ $\mathrm{[m]}$",
}

def configure_matplotlib() -> None:
    """Configure a compact publication-style matplotlib theme."""
    plt.rcParams.update(
        {
            # Adjust white padding around image
            # --- Grid Configuration ---
            "axes.grid": True,
            "grid.alpha": 0.5,
            "grid.linestyle": "-",
            "grid.linewidth": 0.8,
            "grid.color": "#B0B0B0",
            # --- Font Configuration ---
            "font.family": "serif",
            "font.serif": ["Computer Modern Serif", "DejaVu Serif", "cmr10"],
            "mathtext.fontset": "cm",
            "axes.formatter.use_mathtext": True,
            "font.size": 10,
            "axes.labelsize": 10,
            # --- Legend Configuration ---
            "legend.fontsize": 10,
            "legend.frameon": True,
            "legend.facecolor": "white",
            "legend.framealpha": 0.8,
            "legend.edgecolor": "#A0A0A0",
            "legend.fancybox": False,
            "patch.linewidth": 0.7,
            # --- Axes and Spines ---
            "axes.linewidth": 0.8,
            "xtick.direction": "out",
            "ytick.direction": "out",
            "savefig.bbox": "tight",
            "figure.dpi": 150,
            "lines.linewidth": 1.8,
        }
    )

    if shutil.which("latex") is not None:
        plt.rcParams.update({"text.usetex": True})
```

paper_results/final_outputs/benchmark_cpu_system_id_residual_learning.svg
mstoelzle, line 1: I think it would make sense if the markers have the same color as the corresponding backbones.

paper_results/secVe_safety_constrained_control/outputs/force_goal_distance_plot.pdf
mstoelzle, line 1: I think it could possibly make sense to use for the lines the same colors as for the renderings (i.e., safety-unaware and safety-constrained control). Then, the contact force and the goal distance could be different line types. Let's try and see how this would look like.

mstoelzle (changes requested):
Thanks @Michele-Martini! A few questions and comments:
- Whats the point of the `use_cases` folder? How is it different from the `examples` and the `paper_results` folder?
- I was expecting that the `paper_results` folder would also contain the code for generating the data that would later be plotted to produce the paper result figures. This doesn't seem to be the case. What is your intention and concept here?
- I would suggest to remove the figure numbering for now from `paper_results/final_outputs` as we might frequently change the order of figures in the paper so it might be annoying and burdensome to keep these in-sync.
- The script `control_gain_optimization_with_collocated.py` produces NaN results (see output below) when I attempt running it. 

```
(soromox) maximilianstoelzle@mac soromox % python ./examples/optimization/control_gain_optimization_with_collocated.py 
Number of DOFs: 6
Number of actuators: 3

Finding steady-state configuration under constant actuation...
  Applied tendon tensions: [-1.2  -0.6  -0.05]
  Final velocity norm: 4.68e-01
  Warning: Steady-state may not have been reached. Consider increasing sim_duration.
  Steady-state configuration (q_des): [-1.68624440e+00 -2.55435279e+01  1.52828444e+01 -5.91276684e-01
 -2.69822352e-02 -8.04724082e-06]
  Actuation-space coordinates (y_des): [ 2.81442658e-06  4.04018592e-02  8.26159456e-02  1.68624440e+00
 -2.69822352e-02 -8.04724082e-06]
Optimization variables =  ['opt_ctr_params', 'opt_atr_params']

Starting optimization...

[INFO] Compilation + first execution time = 47.82 s

[WARNING] NaN detected at iteration 1. Stopping.
Completion: 33.3 %  |  iteration    1 of 3     |  iter time = {iter_duration:>.4f} s
Optimization Finished. Best optimization variables:
 {'opt_atr_params': {}, 'opt_ctr_params': {'Kd': Array([nan, nan, nan], dtype=float64), 'Ki': Array([nan, nan, nan], dtype=float64), 'Kp': Array([nan, nan, nan], dtype=float64)}}
/Users/maximilianstoelzle/Documents/phd/soromox/src/soromox/rendering/open3d_renderer.py:1858: RuntimeWarning: Open3D viewer assumes cross_section_geometry is configuration-independent; geometry is frozen to the first frame.
  handles = self._build_scene(
Actuation-Space Regulation (Optimized System): Space=Play/Pause  ←/→=Step  H=Home  S=Snapshot  R=ResetCam  C=CaptureCam  L=LoadCam  V=PrintCam  Q/Esc=Quit
```

paper_results/final_outputs/benchmark_cpu_system_id_residual_learning.svg
Michele-Martini: I think they will be hardly visible, but we can try

Michele-Martini:
> * I would suggest to remove the figure numbering for now from `paper_results/final_outputs` as we might frequently change the order of figures in the paper so it might be annoying and burdensome to keep these in-sync.
Sure!

> * The script `control_gain_optimization_with_collocated.py` produces NaN results (see output below) when I attempt running it.
> 
> ```
> (soromox) maximilianstoelzle@mac soromox % python ./examples/optimization/control_gain_optimization_with_collocated.py 
> Number of DOFs: 6
> Number of actuators: 3
> 
> Finding steady-state configuration under constant actuation...
>   Applied tendon tensions: [-1.2  -0.6  -0.05]
>   Final velocity norm: 4.68e-01
>   Warning: Steady-state may not have been reached. Consider increasing sim_duration.
>   Steady-state configuration (q_des): [-1.68624440e+00 -2.55435279e+01  1.52828444e+01 -5.91276684e-01
>  -2.69822352e-02 -8.04724082e-06]
>   Actuation-space coordinates (y_des): [ 2.81442658e-06  4.04018592e-02  8.26159456e-02  1.68624440e+00
>  -2.69822352e-02 -8.04724082e-06]
> Optimization variables =  ['opt_ctr_params', 'opt_atr_params']
> 
> Starting optimization...
> 
> [INFO] Compilation + first execution time = 47.82 s
> 
> [WARNING] NaN detected at iteration 1. Stopping.
> Completion: 33.3 %  |  iteration    1 of 3     |  iter time = {iter_duration:>.4f} s
> Optimization Finished. Best optimization variables:
>  {'opt_atr_params': {}, 'opt_ctr_params': {'Kd': Array([nan, nan, nan], dtype=float64), 'Ki': Array([nan, nan, nan], dtype=float64), 'Kp': Array([nan, nan, nan], dtype=float64)}}
> /Users/maximilianstoelzle/Documents/phd/soromox/src/soromox/rendering/open3d_renderer.py:1858: RuntimeWarning: Open3D viewer assumes cross_section_geometry is configuration-independent; geometry is frozen to the first frame.
>   handles = self._build_scene(
> Actuation-Space Regulation (Optimized System): Space=Play/Pause  ←/→=Step  H=Home  S=Snapshot  R=ResetCam  C=CaptureCam  L=LoadCam  V=PrintCam  Q/Esc=Quit
> ```

This is new and unfortunate. The last version of Soromox i used to run this script, it was successful (the results in the plot come from there). Maybe something in the latest version has changed things, my guess is the se3 operations, as in the backward pass both branches of the the lax.cond functions are executed. Also, if there version still includes the linal.norm() error function, that can go to nan when the error is exactly zero. I will look at it asap

mstoelzle:
> > * I would suggest to remove the figure numbering for now from `paper_results/final_outputs` as we might frequently change the order of figures in the paper so it might be annoying and burdensome to keep these in-sync.
> 
> Sure!
> 
> 
> 
> > * The script `control_gain_optimization_with_collocated.py` produces NaN results (see output below) when I attempt running it.
> 
> > 
> 
> > ```
> 
> > (soromox) maximilianstoelzle@mac soromox % python ./examples/optimization/control_gain_optimization_with_collocated.py 
> 
> > Number of DOFs: 6
> 
> > Number of actuators: 3
> 
> > 
> 
> > Finding steady-state configuration under constant actuation...
> 
> >   Applied tendon tensions: [-1.2  -0.6  -0.05]
> 
> >   Final velocity norm: 4.68e-01
> 
> >   Warning: Steady-state may not have been reached. Consider increasing sim_duration.
> 
> >   Steady-state configuration (q_des): [-1.68624440e+00 -2.55435279e+01  1.52828444e+01 -5.91276684e-01
> 
> >  -2.69822352e-02 -8.04724082e-06]
> 
> >   Actuation-space coordinates (y_des): [ 2.81442658e-06  4.04018592e-02  8.26159456e-02  1.68624440e+00
> 
> >  -2.69822352e-02 -8.04724082e-06]
> 
> > Optimization variables =  ['opt_ctr_params', 'opt_atr_params']
> 
> > 
> 
> > Starting optimization...
> 
> > 
> 
> > [INFO] Compilation + first execution time = 47.82 s
> 
> > 
> 
> > [WARNING] NaN detected at iteration 1. Stopping.
> 
> > Completion: 33.3 %  |  iteration    1 of 3     |  iter time = {iter_duration:>.4f} s
> 
> > Optimization Finished. Best optimization variables:
> 
> >  {'opt_atr_params': {}, 'opt_ctr_params': {'Kd': Array([nan, nan, nan], dtype=float64), 'Ki': Array([nan, nan, nan], dtype=float64), 'Kp': Array([nan, nan, nan], dtype=float64)}}
> 
> > /Users/maximilianstoelzle/Documents/phd/soromox/src/soromox/rendering/open3d_renderer.py:1858: RuntimeWarning: Open3D viewer assumes cross_section_geometry is configuration-independent; geometry is frozen to the first frame.
> 
> >   handles = self._build_scene(
> 
> > Actuation-Space Regulation (Optimized System): Space=Play/Pause  ←/→=Step  H=Home  S=Snapshot  R=ResetCam  C=CaptureCam  L=LoadCam  V=PrintCam  Q/Esc=Quit
> 
> > ```
> 
> 
> 
> This is new and unfortunate. The last version of Soromox i used to run this script, it was successful (the results in the plot come from there). Maybe something in the latest version has changed things, my guess is the se3 operations, as in the backward pass both branches of the the lax.cond functions are executed. Also, if there version still includes the linal.norm() error function, that can go to nan when the error is exactly zero. I will look at it asap
> 
> 

Yes, please try running it yourself with the latest SoRoMoX version. If the optimization/closes-loop rollouts are close to instability, it could even be a numerical issue where the script works on your workstation, but not on my macBook. If this is the case, we should try to make it more stable and robust.

paper_results/final_outputs/benchmark_cpu_system_id_residual_learning.svg
Michele-Martini: [3d_shape_comparison.pdf](https://github.com/user-attachments/files/29479943/3d_shape_comparison.pdf)

what do you think?

paper_results/secVe_safety_constrained_control/outputs/force_goal_distance_plot.pdf
Michele-Martini: force_goal_distance_plot

what about this?

mstoelzle: @Michele-Martini It currently doesn't agree with the renderings. Indeed, I think we meant the opposite:
- Use the colors to differentiate between the two controllers (similar idea as before and after optimization)
- Use line types to distinguish between the goal distance and normal force

paper_results/final_outputs/benchmark_cpu_system_id_residual_learning.svg
mstoelzle: I think it looks good - thanks!

paper_results/secVe_safety_constrained_control/outputs/force_goal_distance_plot.pdf
Michele-Martini: force_goal_distance_plot

Now?

mstoelzle: Looks good! Do you agree?

Michele-Martini: I think it's cool! The only thing for me is that usually the graphs used the pre_opt and post_opt colors, while these 2 colors were reserved for the backbones, but it's okay it's just a matter of style preferences :)

mstoelzle: Good point! But then, wouldn't it make sense to also render the backbones with `pre_opt` and `post_opt` colors (both for this case study, and the other case studies)?

Michele-Martini: Could be, but at the beginning i recall that we wanted some dedicated colors for the backbones. Tell me what do you prefer, but since doing the rendering images is manual and a bit laborious, let's decide a final version.

Michele-Martini: We could use pre_opt and post_opt for the backbones as well, then use the green for the targets and violet for the obstacles

mstoelzle: I think we were originally rather more referring to which colors to use to color the various segments. However, it seems that now we have converged to using the same color for the entire robot anyway.

I think we should use the `pre_opt` and `post_opt` colors also for rendering the backbone.

Maybe, we could still point out the segments: Use a slight color gradient from very solid for the most proximal segment to slightly more white for the distal segment. So, something like the Matplotlib "Sequential Colormap", but for our `pre_opt` and `post_opt` colors.

Screenshot 2026-06-30 at 18 37 35

> but since doing the rendering images is manual and a bit laborious, let decide a final version.
Two things:

- Yeah, I think it would be best if you post the rendered video and we can agree (or possibly iterate) on that before updating the paper figure.
- You can also extract image frames programtically instead of manually frame-by-frames. Your AI agent of choice can help you with that.

mstoelzle: > We could use pre_opt and post_opt for the backbones as well, then use the green for the targets and violet for the obstacles

Yeah, sounds good. Note my comment about the color gradient to signify segments above.

.gitignore
```diff
@@ -140,6 +140,21 @@ dmypy.json
...
 
+# Generated example artifacts stay next to the script that creates them.
+examples/**/figures/*
+examples/**/videos/*
+
+# Canonical paper media is versioned with Git LFS.
+!paper_results/secVc_model_based_control/**/*.mp4
+!paper_results/secVc_model_based_control/**/*.gif
+
+# Section Vc diagnostics are reproducible; only paper figures and video are canonical.
```
mstoelzle: Why do we need such specialized rules for Section Vc? Are they still up-to-date? Why are we not including them in git?

.gitattributes
```diff
@@ -0,0 +1,5 @@
+paper_results/secVc_model_based_control/**/*.npz filter=lfs diff=lfs merge=lfs -text
```
mstoelzle: Are the modle-based control artifacts (e.g., pdf, png, mp4) much larger than for other paper result sections? Why are they all in git lfs?

data/trial_1/animation_data.pkl
mstoelzle, line 1: Where is the data in `data/` used? Also, `trial_1` is a quite bad and not very informative name.

mstoelzle (changes requested):

.gitignore
mstoelzle: Addressed in df02e9c. I removed the Section Vc-specific rules from the root `.gitignore`. Each Section Vc case now owns a small `outputs/.gitignore` that ignores reproducible diagnostics while explicitly keeping its documented canonical figures/video. The canonical outputs remain committed.

.gitattributes
mstoelzle: Addressed in df02e9c. The Section Vc artifacts are not unusually large (the largest NPZ is about 5.2 MB and the video about 2.5 MB), so I removed the Section Vc LFS rules and renormalized all of those artifacts as ordinary Git objects. Their content hashes are unchanged. Git LFS is now reserved here for the trained Section Vf checkpoints.

data/trial_1/animation_data.pkl
mstoelzle: Addressed in 1348179. No tracked script or documentation consumed the two `data/trial_1` files, so both were removed from the PR instead of retaining or renaming an unowned dataset.

src/soromox/rendering/open3d_renderer.py
```diff
@@ -2092,6 +2096,13 @@ def update_frame(i: int) -> None:
                     state["playing"] = state["playing"] and loop
                 update_frame(nxt)
                 state["last_tick"] = now
+                if (
+                    record_cfg.close_when_done
+                    and record_cfg.path is not None
+                    and not loop
+                    and nxt == scene_data.num_frames - 1
+                ):
+                    break
```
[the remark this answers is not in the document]
mstoelzle: Addressed in df02e9c. `_run_viewer` now closes the video writer and destroys the Open3D window in a `finally` block, including the automatic end-of-recording path. The explicit quit callback records that it already destroyed the window to avoid a duplicate call. A regression test covers automatic shutdown after the final frame.

paper_results/secVf_parallel_rl/code/render_rl_video.py
```diff
@@ -0,0 +1,240 @@
...
+        type=str,
+        default=OUTPUT_DIR / "rl_rollout.mp4",
+        help="Output MP4 path",
+    )
+    parser.add_argument(
+        "--show-trajectory",
+        action="store_true",
+        default=True,
+        help="Plot the target ball's trajectory as a trail of small red spheres",
+    )
```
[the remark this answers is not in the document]
mstoelzle: Addressed in df02e9c. The option now uses `argparse.BooleanOptionalAction`, remains enabled by default, and can be disabled with `--no-show-trajectory`. Parser tests cover both the default and disabled states.

mstoelzle (approved):

PR #99, opened 2026-07-05
updated outer radius

Elektron97:
The model presented in the [paper](https://ieeexplore.ieee.org/abstract/document/10098800) uses this area only to scale the pressure values into force (Newtons). However, when pressure is applied to the chambers, they unroll (as shown [here](https://link.springer.com/article/10.1007/s40964-020-00158-y?utm_source=researchgate.net&utm_medium=article)), causing the effective pressure to act over the entire annular area, which is approximated by the outer circle rather than the inner one.

### DIFF
--- a/src/soromox/systems/pcs/isupport.py
+++ b/src/soromox/systems/pcs/isupport.py
@@ -363,7 +363,8 @@ class ISupport(PCS):
 
         def _actuation_matrix_one_chamber(i: Array, varphi: Array) -> Array:
             # Area of one pneumatic chamber
-            A_one_chamber = jnp.pi * self.r_chamber_in[i] ** 2
+            # A_one_chamber = jnp.pi * self.r_chamber_in[i] ** 2
+            A_one_chamber = jnp.pi * self.r_chamber_out[i] ** 2
 
             # force contribution of the chamber
             force_contrib = jnp.array(


Elektron97:
[WARNING]: The scale factor bar->Pa of 1e+5 still creates overactuation. Probably is the actuation matrix itself the problem.
In the meanwhile, the identification is done here in MATLAB in this [repo](https://github.com/Elektron97/am_isupportGVS).

mstoelzle (changes requested):
Are we sure that this change makes sense?

Would it not be more accurate if we only consider the area between the inner and outer radius? So `jnp.pi * (self.r_chamber_out[i] - self.r_chamber_in[i]) ** 2`?

Elektron97:
Yep, the air flows in all the chamber area for the unfolding process.

mstoelzle:
> Yep, the air flows in all the chamber area for the unfolding process.

But where the air flows doesn't matter. What counts is the cross-sectional area where the pressure acts in the direction of the backbone. That is why I was proposing `jnp.pi * (self.r_chamber_out[i] - self.r_chamber_in[i]) ** 2` (which assumes that the `self.r_chamber_out ` is the outer radius of the bellow, and  `self.r_chamber_in ` the inner radius of the bellow)

Elektron97:
For sure $\pi (r_o^{2} - r_i^{2})$ is not correct, as you can observe the structure of the chamber [here](https://link.springer.com/article/10.1007/s40964-020-00158-y). If we want to be more conservative, we can keep $\pi r_i^{2}$ to be consistent with the previous [paper](https://ieeexplore.ieee.org/abstract/document/10098800). Let me know.

mstoelzle:
Closing for now as (1) we are currently doing a lot of migration work on the `ISupport`, (2) we first need to settle on how to best identify the effective pneumatic chamber radius / area.

PR #105, opened 2026-07-09
Add PCS material damping coefficient

mstoelzle:
## Summary

This PR adds `material_damping_coefficient` as the preferred damping input for PCS-based systems while preserving explicit `damping_matrix` support for exact/custom damping.

The change covers:

- `PCSParams` and `PlanarPCSParams` validation for exactly one damping input.
- Scalar or per-segment material damping coefficients.
- Runtime conversion from material damping coefficients into `D_full` using model geometry.
- Inherited PCS systems, including tendon-actuated PCS variants and I-SUPPORT.
- Selected control/simulation example migration and docs updates.
- Regression tests for material damping, explicit damping matrices, parameter updates, and inherited paths.
- Documentation of local material-frame x-axis conventions, PCS circular cross-sections, and tendon routing planes across PCS, GVS, and Pendulum models.
- Corrected material-damping units and clarification of generalized damping-matrix entry units.

`young_modulus` remains the canonical modulus keyword; this PR does not add an `elastic_modulus` alias.

## API Behavior

PCS-family params now accept exactly one of:

```python
material_damping_coefficient=...
```

or:

```python
damping_matrix=...
```

`material_damping_coefficient` can be either a scalar, broadcast to all segments, or an array with shape `(num_segments,)`. It is a viscosity-like modulus with units `Pa·s` (`N·s/m²`).

The assembled damping matrix also contains segment length and cross-sectional geometry factors. Its entries therefore have units determined by the associated generalized strain-coordinate pair rather than one blanket `Pa·s` unit.

For example, a spatial PCS model can now be constructed as:

```python
import jax.numpy as jnp
from soromox.systems import PCS, PCSParams

num_segments = 2
params = PCSParams(
    base_pose=jnp.array([1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]),
    length=jnp.array([0.1, 0.12]),
    radius=jnp.array([0.02, 0.02]),
    density=1070.0 * jnp.ones((num_segments,)),
    young_modulus=2e3 * jnp.ones((num_segments,)),
    shear_modulus=1e3 * jnp.ones((num_segments,)),
    material_damping_coefficient=362.0,
    gravity=jnp.array([0.0, 0.0, -9.81]),
    reference_strain=jnp.tile(
        jnp.array([0.0, 0.0, 0.0, 1.0, 0.0, 0.0]), num_segments
    ),
)
robot = PCS(params=params)

D_full = robot.D_full
D_active = robot.damping_matrix(jnp.zeros(robot.num_dofs))
```

Users who need exact damping ratios can still provide the full matrix directly:

```python
params = params.replace(
    material_damping_coefficient=None,
    damping_matrix=jnp.eye(6 * num_segments),
)
robot = robot.with_params(params)
```

## Internal Conversion

For spatial PCS systems, a segment coefficient `c_i` is converted into a block of the full damping matrix as:

```text
D_i = L_i * c_i * diag([Ix, 3Iy, 3Iz, 3A, A, A])
```

where:

- `L_i` is the segment length,
- `A` is the segment cross-sectional area,
- `Ix`, `Iy`, `Iz` are the local second moments of area.

The full and active-coordinate matrices are assembled as:

```text
D_full = block_diag(D_0, D_1, ..., D_{n-1})
D_active = B_xi.T @ D_full @ B_xi
```

For planar PCS systems, the corresponding block is:

```text
D_i = L_i * c_i * diag([3I, 3A, A])
```

This mirrors the stiffness cache naming:

```text
K_full = block_diag(K_0, K_1, ..., K_{n-1})
K_active = B_xi.T @ K_full @ B_xi
```

The implementation deliberately calls the model geometry helpers while assembling damping. This means inherited systems use their actual geometry: for example, I-SUPPORT computes material damping with its actuator cross-section geometry rather than the default circular backbone geometry.

For migrated examples, the coefficient values are rounded least-squares fits from the previous diagonal damping matrices:

- Standard spatial PCS with `r = 2e-2`: `362.0`
- Standard planar PCS with `r = 2e-2`: `318.0`
- I-SUPPORT: `1.96e3`

The least-squares projection used for I-SUPPORT is documented in the example:

```text
c = <D_old, D_material(1)>_F / <D_material(1), D_material(1)>_F
```

## Example Migration

Migrated selected tutorial-style examples in:

- `examples/control`
- `examples/simulation/pcs`
- `docs`

Intentionally not migrated:

- `examples/control/operational_space/control_pcs_with_cf_cbf_clf.py`
- `examples/control/rl/parallel_soromox_env.py`
- `examples/simulation/pcs/simulate_pressure_actuated_planar_pcs.py`
- `examples/optimization/**`
- `paper_results/benchmarking_sequential_cpu/**`

## Geometry and Coordinate Documentation

- PCS and PlanarPCS now explicitly define the material-frame local x-axis as the undeformed longitudinal backbone tangent, independent of the base pose in the inertial frame.
- PCS-family solid circular cross-sections are documented where radius-derived area and second moments feed mass, stiffness, and material damping.
- Tendon-actuated PCS and GVS models now document tendon offsets in the local cross-sectional plane normal to local x.
- GVS retains configurable circular, rectangular, and elliptical cross-sections; tendon actuation does not impose a circular-section assumption.
- Pendulum documents that zero-angle links point along local +x and that radius is visualization-only.
- ArticulatedSoftRobot remains general and unchanged because its link direction is defined by arbitrary `tip_position` vectors.
- Existing GVS damping-coefficient documentation was corrected from `N·s/m` to `Pa·s` (`N·s/m²`) to match the constitutive formula.

## Implementation Notes

- Removed the duplicate PCS `D` cache; `D_full` is now the single full resolved damping cache, and `D_active` remains the active-coordinate cache.
- Renamed the active-coordinate stiffness cache from `K` to `K_active` and updated all usages; no compatibility alias is kept.
- Renamed the explicit matrix helper to `_explicit_damping_full_matrix()` to avoid implying that a zero placeholder is part of the runtime damping path.
- Validation rejects ambiguous or incomplete damping input: both provided, neither provided, bad coefficient shape, or bad matrix shape.

## Validation

Ran:

```text
uv run ruff check ...
uv run python -m py_compile ...
uv run python -c 'import the updated PCS, GVS, tendon-actuated, and Pendulum modules'
Documentation-only AST comparison against HEAD
uv run pytest tests/systems/test_typed_params_api.py
uv run pytest tests/systems/test_pcs.py
uv run pytest tests/systems/test_planar_pcs.py
uv run pytest tests/systems/test_tendon_actuated_pcs.py tests/systems/test_tendon_actuated_planar_pcs.py tests/systems/test_pressure_actuated_pcs_models.py
```

All checks passed.


examples/simulation/pcs/simulate_isupport.py
```diff
@@ -39,27 +39,13 @@
...
-            * pneumatic_segment_lengths[:, None]
-        ).flatten()
-    )
+    # Previous explicit damping used gamma_t = 806e-3 and gamma_r = 1.0e-3:
+    # D_i = L_i * diag([gamma_r, gamma_r, gamma_r, gamma_t, gamma_t, gamma_t]).
+    # The material damping coefficient below is the least-squares scalar c for
+    # D_i(c) = L_i * c * diag([Ix, 3Iy, 3Iz, 3A, A, A]), evaluated with
+    # I-SUPPORT's actuator cross-section geometry. Equivalently,
+    # c = <D_old, D_material(1)>_F / <D_material(1), D_material(1)>_F.
+    material_damping_coefficient = 1.96e3
```
Elektron97, line 48: Why scale on L_i?

To the best of my knowledge, 

Psi = [Ix, 3Iy, 3Iz, 3A, A, A]

and after integrate as usual. Maybe I miss something with the computation of Psi?

Elektron97: Minor: the "3" scale in reality is valid only for circular cross-section

docs/api/systems/pcs/tendon-actuated-pcs.md
```diff
@@ -30,7 +30,7 @@ body_params = PCSParams(
     density=jnp.array([1000.0, 1000.0]),
     young_modulus=jnp.array([1e6, 1e6]),
     shear_modulus=jnp.array([1e5, 1e5]),
-    damping_matrix=jnp.eye(12),
+    material_damping_coefficient=jnp.array([362.0, 362.0]),
```
Elektron97, line 33: btw, why just call $\eta$ as in the Cosserat/GVS papers?

Elektron97 (commented):
The implementation works but I find confusing why the damping tensor depends on the length of the robot and the value of c. Is just for migration from previous implementation?

docs/api/systems/pcs/tendon-actuated-pcs.md
mstoelzle: You mean calling the parameter \eta? The reason is that we are spelling out all other parameters and not using greek/mathematical symbols (e.g., `young_modulus` instead of `E`).

examples/simulation/pcs/simulate_isupport.py
mstoelzle: This is just a placeholder anyway. We will remove it as soon as we can put there the `material_damping_coefficient` stemming from your system identification.

mstoelzle: > Minor: the "3" scale in reality is valid only for circular cross-section

What do you think we should use for the ISupport to go from the material damping coefficient to the full damping matrix? Can you send me what model I should implement here?

I agree that the ISuport is not a _truly_ circular cross section so the assumption might not be correct.t

Elektron97 (approved):
Perfect, everything clear now. Thank you.

PR #109, opened 2026-07-11
Adopt explicit I-SUPPORT chamber azimuths

mstoelzle:
## Summary

- replace the implicit chamber-angle offset with explicit per-channel chamber azimuth arrays
- adopt the right-handed local +X convention from local +Y toward local +Z
- share chamber geometry between the PCS actuation model and specialized Viser renderer
- validate rotational symmetry while preserving pressure-channel order
- provide a canonical default layout of 0°, 120°, and 240°
- update examples, API documentation, and regression tests

## Why

The previous geometry used a non-right-handed azimuth convention and duplicated its trigonometry across the model and renderer. It also inferred chamber positions from one offset, obscuring the mapping between physical chambers and pressure inputs. This change makes the model the single geometry source of truth and gives each pressure channel an explicit azimuth.

## Impact

`chamber_angle_offset` is replaced by `chamber_azimuth_angles`. Existing configurations can omit the new field to receive the canonical three-chamber layout, or pass an explicit `(num_pneumatic_segments, num_chambers_per_segment)` array when channel order differs.

## Validation

- `uv run ruff check` on affected Python files
- focused I-SUPPORT system, renderer, and length tests: 46 passed
- `git diff --check`

### DIFF
--- a/docs/api/systems/pcs/isupport.md
+++ b/docs/api/systems/pcs/isupport.md
@@ -51,7 +51,7 @@ params = ISupportParams(
     chamber_inner_radius=jnp.array([0.00639]),
     chamber_outer_radius=jnp.array([0.00779]),
     chamber_distance=jnp.array([0.020]),
-    chamber_angle_offset=jnp.array([0.0]),
+    chamber_azimuth_angles=(2.0 * jnp.pi * jnp.arange(3) / 3)[None, :],
 )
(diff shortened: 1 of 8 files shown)


mstoelzle:
@Elektron97 I am going to merge now as it is just an API change, but might be relevant for you if you start implementing code based on soromox yourself.

PR #111, opened 2026-07-11
AM-I Support Experimental Adaption

### LINKED ISSUE #98: Use outer chamber radius for i-support actuation area
## Context

In `ISupportPcsSystem.actuation_matrix`, the pressure area of one pneumatic chamber is currently computed from the inner chamber radius:

```python
A_one_chamber = jnp.pi * self.r_chamber_in[i] ** 2
```

Reference: https://github.com/tud-phi/soromox/blob/am_isupport/src/soromox/systems/pcs/isupport.py#L366

## Requested change

Switch this calculation to use the outside chamber radius instead of the inside chamber radius, i.e. use `self.r_chamber_out[i]` for the chamber pressure area.

## Acceptance criteria

- `actuation_matrix` uses the outside chamber radius for `A_one_chamber`.
- Any affected tests or fixtures are updated to reflect the corrected actuation area.
- Existing geometry calculations that intentionally depend on the annular cross-section remain unchanged unless explicitly related to this pressure-area correction.

Elektron97:
# AM-I Support Experimental Adaption

## Summary

This PR reworks the I-SUPPORT simulation example so that it is driven by **recorded chamber pressures** exported from MATLAB (`6 × M` matrix on `merged_time`) instead of a constant open-loop actuation vector. The recorded input is fed through the closed-loop rollout hook as a time-varying open-loop signal, and the simulated end-effector trajectory is now validated against recorded VICONtip positions.

Alongside the new input pipeline, the system parameters have been updated to a **identified parameter set** and the geometry/frame conventions have been adapted to match the real experimental setup.

### 1. Identified system parameters

The material and geometric parameters were replaced with the values identified for the physical AM I-SUPPORT arm:

| Parameter | Before | After |
|---|---|---|
| Young's modulus `E` | `1.6464e6` Pa | `1.955596117275246e6` Pa |
| Material damping `η` | `1.96e3` Pa·s | `1.356440907178834e5` Pa·s |
| Density `ρ` | `1104` kg/m³ | `1.015625845884526e4` kg/m³ |
| Segment radius | `35.6e-3` m | `28.6e-3` m |
| Poisson ratio | `0.5` | `0.5` (unchanged) |

Shear modulus is still derived as `G = E / (2(1 + ν))`. For more details, please check this [repo](https://github.com/Elektron97/am_isupportGVS).

### 2. Experimental adaptation for matching the data

Several changes align the simulation with the real experimental frame, geometry, and actuation hardware:

- **Base pose / frame convention.** The base orientation now encodes a `-30°` experimental offset about the first quaternion axis (`theta = -pi/6`) and a base position of `21 mm` along `x`, replacing the previous fixed quaternion `[0.5, 0.5, 0.5, -0.5]`.
- **Gravity direction.** Gravity now points along `+x` (`[9.81, 0, 0]`) instead of `-z`, matching the mounting orientation of the physical rig.
- **Segment geometry.** Both pneumatic segments are now `190 mm` (`2 × 190e-3`) with a single-PCS discretization per segment (`pcs_segment_counts = (1, 1)`, `pcs_segment_lengths = None`), replacing the previous mixed `190/180 mm` segmentation.
- **Rigid connector lengths.** Base connector shortened to match the real assembly: `[6e-3, 27e-3, 6e-3]` instead of `[41e-3, 27e-3, 6e-3]`.
- **Chamber azimuth phases.** Azimuth angles are shifted by `+90°` to `[90°, 210°, 330°]` (per the Alessi paper) instead of `[0°, 120°, 240°]`.
- **Chamber permutation.** A permutation matrix (`perm_idx = [2, 1, 0, 5, 4, 3]`) reorders the pressure channels to match the physical chamber wiring of the ROS node.
- **Air-leak compensation.** A diagonal gain matrix built from measured pressures at 3 bar (`p_real`) scales the commanded pressures to account for per-chamber air leaks (`gain = diag(p_real) / 3`).
- **Units.** Recorded pressures are converted from bar to Pa (`× 1e5`).

### 3. Recorded input via the closed-loop rollout hook

The core mechanism change: instead of `rollout_to(u=...)` with a constant `u`, the simulation now uses `rollout_closed_loop_to(...)` with a state-ignoring "controller" that interpolates the recorded pressures at the live solver time.

- New `RecordedInput(eqx.Module)` implements `__call__(self, state)`:
  - reads `state.t`,
  - interpolates each of the 6 recorded channels via `jnp.interp` (vectorized with `jax.vmap`),
  - applies `air_leaks @ perm @ u_t`,
  - returns `(u_t, None)` — a flat `(6,)` input and no control-state derivative.
- This effectively implements an **open-loop, time-varying** control law expressed through the closed-loop API (`base_u = 0`, controller ignores `q`/`qd`).
- The applied actuation is captured per save step (`trajectory.u`) and plotted as a sanity check against the recorded data.

### 4. Adaptive-step integration

The solver setup was updated for the longer, data-driven rollout:

- `Tsit5()` with an adaptive `PIDController(rtol=1e-5, atol=1e-7)` replacing fixed-step integration.
- Solver step relaxed to `solver_dt = 1e-3` (previously `2e-5`), with `save_dt = 1e-3`.
- Simulation window set to the experiment segment `t0 = 15.22 s`, `t1 = 60.22 s`, with `save_ts` guarded against float-accumulation overshoot past `t1`.
- Initial configuration reset to the **undeformed reference strain** (zeros) rather than the previous non-trivial `q0`.

### 5. Vicon overlay and CSV export

- Loads recorded data (`poses_data`, `markers`, `merged_time`, `N`, `M`, `actuation_input`) from the `.mat` file, with shape assertions on the actuation array.
- Extracts the Vicon tip position (last cross-section) and interpolates each axis onto the simulation save times, skipping NaNs and refusing to extrapolate outside the recording window.
- Adds **per-axis** and **3D overlay** plots of simulated EE vs. Vicon tip.
- Exports `tip_positions.csv` (sim + Vicon `x/y/z`) and saves all figures to `amisupport_dataset/results/`.


Elektron97:
# AM-I Support Experimental Adaptation - 2

## 1) New Results in SoRoMox (AM-I Support v1)

System identification was repeated using a chamber area of $\pi r_{in}^2$, as decided [here](https://github.com/tud-phi/soromox/issues/98).

| Parameter | Before | After |
| :--- | :--- | :--- |
| **Young's modulus `E`** | 1.6464e6 Pa | 1.076632416072824e+6 Pa |
| **Material damping `η`** | 1.96e3 Pa·s | 8.3795344450652e+4 Pa·s |
| **Density `ρ`** | 1104 kg/m³ | 6.471046068502e+3 kg/m³ |
| **Poisson ratio** | 0.5 | 0.5 (unchanged) |

### 1.1) Methodology
For detailed information on the methodology, please refer to the [original repository's README](https://github.com/Elektron97/am_isupportGVS).

### 1.2) Predicted vs. Experimental Tip Position

tip_position

3dtip_position

To match the realistic control box behavior, a first-order low-pass filter was also implemented in the code:

```python
# First-order low-pass dynamics: tau * du/dt = u_raw - u_filt.
du_dt = (u_raw - u_filt) / self.tau

```
u_filtered

---

## 2) Experimental Measurements on the AM-I Support Segment

### 2.1) Mass & Density

#### 2.1.1) System Geometry and Measurements

* **Total Segment Mass:** 183 g (0.183 kg)
* **Total Segment Length:** 178 mm (0.178 m)
* **Chamber Inner Radius:** 6.39 mm
* **Chamber Outer Radius:** 7.79 mm

Based on these radii, the cross-sectional area of a single chamber is calculated, which is then used to determine the volume of a single chamber over the length of the segment.

#### 2.1.2) Rigid Interface Properties

The segment is capped by two rigid PLA interfaces (one short, one long).

* **Material Density (PLA):** 1210 kg/m³
* **Interface Radius:** 30 mm
* **Short Interface Length:** 6 mm
* **Long Interface Length:** 21 mm

The calculated mass of the interfaces is **0.0924 kg**. By subtracting this from the total mass, we obtain the soft material mass: **0.0906 kg**.

#### 2.1.3) Density of the Soft Link

The soft segment consists of three identical chambers ($3 \cdot V_c$), where the volume of a single chamber $V_c$ is:

$$V_c = \pi (r_{out}^2 - r_{in}^2) \cdot L$$

The final density of the soft material is calculated by dividing the isolated soft mass by the combined volume of these three chambers:

```matlab
rho_soft = m_soft / (3 * V_c) = 2721.2335 % [kg/m^3]

```

### 2.2) Pressure-Dependent Young's Modulus

As analyzed [here](https://github.com/Elektron97/am_isupportGVS#31-managing-the-folding-and-unfolding-of-the-actuators), the unfolding behavior of the actuator makes the Young's modulus pressure-dependent. Specifically, five distinct phases were identified:

1. **Dead Zone (Blue):** The control box does not apply pressure to the robot at these low values.
2. **Pre-Unfolding:** The actuators begin to unfold but do not produce significant elongation as pressure increases.
3. **Unfolding:** The actuators rapidly unfold due to their geometry with only a small increase in pressure, resulting in a high equivalent Young's modulus.
4. **Linear Fully Unfolded:** The actuators elongate due to material elasticity, exhibiting a classical linear pressure-elongation relationship.
5. **Non-Linear Fully Unfolded:** The elongation of the actuators no longer follows a strict linear path.

AM-I Support Parameter Identification

By applying linear regression for each region (excluding the dead zone), we can compute the corresponding equivalent Young's Modulus:

* **Pre-Unfolding:** 13.334 MPa
* **Unfolding:** 0.042 MPa
* **Linear Fully Unfolded:** 1.657 MPa
* **Non-Linear Fully Unfolded:** 2.605 MPa

While the pre-unfolding phase can practically be considered a continuation of the dead zone, the unfolding phase demonstrates elongation due to geometry rather than material elasticity. Interestingly, the "Linear Fully Unfolded" modulus closely matches the findings in the [Alessi et al. paper](https://ieeexplore.ieee.org/abstract/document/10098800).

Finally, this piecewise constant Young's modulus can be fitted with several sigmoids to create a smooth, continuous function with respect to pressure:

Young&#39;s Modulus Transition (E2 to E4)

---

## Next Steps

* Test AM-I Support v2 using these updated parameter values.
* Incorporate this pressure-dependent Young's modulus into the SoRoMox model.

Elektron97:
**Pull Request will be uploaded soon.**

square_tip_3d_equal

Elektron97:
## Update: re-identified configuration, and parity re-verified at **0.019 mm**

Two new commits on `am_isupport`. The port itself needed **no changes** to accept the new configuration — every parameter already travels in the exported `.mat`, which is the point of the parameter-free design in this PR. The only fix on the MATLAB side was that `python_exporter.m` still hard-coded `perm_idx`; it now reads `cfg.perm3`.

### What changed on the MATLAB side

The previous reference (`T1_fc28`) put the geometry fix in a **base roll of −30° plus a +15° chamber-triad clock**. That pair was re-derived rather than inherited, at the constant-strain basis (`Phi_odr = 0`) and at the soromox chain length (35 mm base offset + 178 mm modules = 430 mm), with `E`, `Rho` and `Poi` frozen at their characterised values and only `Eta` + three tied gains free.

Three results worth flagging:

**1. `g0_roll` and `clock_deg` are the same angle for tip *position*.** Roll −15 with clock 0 reproduces roll −30 with clock +15 to every digit of tip RMSE. They differ only in the rest-frame *orientations*, which a position-only objective cannot see. Collapsing to roll-only is therefore free on the scored metric — and simpler to port, since the clock is what forced this PR to snap a non-uniform `dc` triad. The honest cost is a ~17° rest-orientation bias, so the identified model should not be used for orientation control without revisiting it.

**2. The chamber permutation was re-answered, not inherited.** A cyclic permutation is a 120° triad rotation and is reachable by roll; the reflection `[3 2 1]` is not, so the question is genuinely open once the clock is pinned at 0. The reflection still wins — 39.26 mm vs 40.24 mm fitted-vs-fitted on the square — and the chirp rejects the cyclic wiring outright (463 mm, `y` inverted).

roll-only sweep

**3. The two bags are different experiments, so their measured bending planes cannot be compared as a geometry check.** The square drives two chambers against one and the tip sweeps a line (planarity 0.07); the chirp drives three sinusoids 120° apart, swept 0.221 → 1.786 Hz, so the tip *whirls* (planarity 0.32) with a ~1.9 bar common mode the square never exercises. The 81° between the two measured plane normals is the **input**, not the robot.

excitation

### New reference configuration

```
run            U1_rollonly_g35178   (config hash c2ac3587)
geometry       g_ini roll -13 deg, chamber clock 0, perm_idx [3 2 1 6 5 4]
               35 mm base offset + 2 x 178 mm soft modules = 430 mm chain
strain basis   Phi_odr = 0 (constant), ndof 12
FROZEN         E = 2.205786e6 Pa | Rho = 2721.2 kg/m3 | Poi = 0.50
IDENTIFIED     Eta = 3.8120e5 Pa*s | g = [2.8067, 2.2666, 2.4509] tied across modules
               fc = 2.8 Hz on pressure, upstream of the force map
```

### Parity: soromox vs MATLAB, square bag

```
LADDER RUNG 0 (rest pose, q = 0)
  tip soromox : [430.0000, 0.0000, 0.0000] mm
  tip MATLAB  : [430.0000, 0.0000, 0.0000] mm
  |diff|      : 5.55e-14 mm                                   PASS

PARITY (TIP)
  RMS difference        :   0.0190 mm   (x 0.0077 | y 0.0042 | z 0.0169)
  max difference        :   0.0686 mm   (sample 611)
  mean signed offset    :  +0.0003  +0.0014  +0.0016 mm
  reference peak-to-peak:    633.4 mm
  RMS / peak-to-peak    :   0.0030 %
  VERDICT               : PASS (threshold 1.0 mm)

PARITY (MID)
  RMS 0.0059 mm | max 0.0225 mm | 0.0029 % of 200.1 mm        PASS
```

parity residual

Against VICON the two engines are indistinguishable:

| metric | MATLAB `ode15s` | soromox `Tsit5` |
|---|---|---|
| tip RMSE (3D) | 39.324 mm | 39.320 mm |
| tip RMSE per axis (x, y, z) | 20.514 / 12.200 / 31.252 mm | 20.514 / 12.200 / 31.248 mm |
| mid RMSE | 24.35 mm | 24.35 mm |
| static offset at t0 | 7.083 mm | 7.083 mm |
| mean amplitude error | 18.244 mm (6.9 %) | 18.244 mm (6.9 %) |
| mean phase lag | −3.60° | −3.60° |
| spectrum 1 / 3 / 5 Hz | −1.0 % / +11.9 % / −82.0 % | −1.0 % / +11.9 % / −82.0 % |

tip traces

Both configurations now pass:

| configuration | chain | rung 0 | tip parity RMS | vs VICON (MATLAB / soromox) |
|---|---|---|---|---|
| `T1_fc28` (roll −30 + clock +15) | 435 mm | 0.00e+00 mm | 0.0276 mm | 40.128 / 40.128 mm |
| `U1_rollonly_g35178` (roll −13 + clock 0) | 430 mm | 5.55e-14 mm | **0.0190 mm** | 39.324 / 39.320 mm |

Nothing was tuned to close the gap — parity is slightly better simply because the 430 mm trajectory is a little smaller.

### MATLAB-side validation

Independently of parity, the same configuration was validated in MATLAB. `identification.m` now asserts `E`/`Rho`/`Poi` frozen for these runs, and every result file is config-hashed.

| run | roll | clock | perm | chain | tip RMSE | x / y / z |
|---|---|---|---|---|---|---|
| `U0_ctrl` (control = `T1_fc28`) | −30 | +15 | [3 2 1] | 435 | 40.127 mm | 20.09 / 14.49 / 31.57 |
| **`U1_rollonly`** | **−13** | **0** | [3 2 1] | 435 | **39.262 mm** | 20.24 / **12.21** / 31.35 |
| `U2_perm312` | +17 | 0 | [3 1 2] | 435 | 40.238 mm | 20.43 / 14.92 / 31.29 |
| `U0_ctrl_g35178` | −30 | +15 | [3 2 1] | 430 | 40.134 mm | 20.53 / 14.34 / 31.36 |
| **`U1_rollonly_g35178`** | **−13** | **0** | [3 2 1] | 430 | **39.323 mm** | 20.52 / 12.20 / 31.25 |

The whole improvement is in the out-of-plane axis `y`: 14.49 → 12.21 mm. That is the first configuration to beat the 13.64 mm a model outputting *constant* `y` would score, i.e. `y` is no longer the dominant defect. The 5 mm chain-length difference costs 0.007 mm (control) / 0.061 mm (roll-only) and leaves the ordering intact.

square fits

### Cost, for anyone re-running

The square rollout took **13 608.9 s (3 h 47 m)** with `Tsit5` against MATLAB `ode15s`'s **27.8 s**. Loosening tolerances changes parity by nothing (the model is stiff, so an explicit step is capped by stability, not accuracy) and `Kvaerno5` is worse. The chirp is 12–17 h and was deliberately not run.

Reproduce:

```bash
python simulate_isupport_exp.py --bag square --tag _U1roll430 --ladder   # rung 0, seconds
python simulate_isupport_exp.py --bag square --tag _U1roll430            # full rollout
```

Elektron97:
## New parity run: exp_veronica pre-impact tail (W7, 430mm geometry)

Extended the parity check to a run outside the square/chirp pair: the
**pre-impact tail** (t = 11.00–13.06 s) of the longest `exp_veronica` bag
(`Eco50_3bar_default_intermadiate_170006`), fit as `W7_veronica_tail_t11_g35178`
on the MATLAB side. This is a different physical experiment from square/chirp
— single-channel step actuation (one chamber per module stepped, not a cyclic
drive) — using an independently-swept offset config (permutation `[2 1 3]`,
base roll `-7°`, chamber clock `0`; confirmed optimal for this bag by a
dedicated perm×roll sweep, not inherited from the square fit) and this repo's
430mm chain geometry (35mm base offset / 178mm soft module) rather than the
435mm one the square/chirp parity uses.

*(Two earlier iterations superseded here: W5 warm-started the initial state
from the model's own simulated state mid-trajectory — flagged on review and
dropped in favor of a verified rest pose (W6); W7 then repeats W6 at this
repo's 430mm geometry instead of 435mm. This IS the final setup, rolled out
end-to-end on both sides.)*

### Parameters

| | value |
|---|---|
| E (frozen) | 2.2058e6 Pa |
| Rho (frozen) | 2721.2335 kg/m³ |
| Poi (frozen) | 0.5 |
| Eta (fit) | 6.1917e5 Pa·s |
| gains g1,g2,g3 (fit, tied) | 1.7807, 0.8961, 1.0049 |
| perm / roll / clock | [2 1 3] / -7° / 0° |
| geometry | 35mm base offset + 178mm soft module (430mm total) |
| fc | 2.8 Hz |
| window | t = [11.00, 13.06] s, 207 samples |
| x0 | **0** (true rest — tip position verified static to <0.6mm before the ~11.66s actuation step) |

### Result: parity PASS

```
tip RMS difference : 0.0118 mm (max 0.0315 mm) on 383.9 mm peak-to-peak
mid RMS difference : 0.0041 mm
rung 0 (rest-pose FK) : exact to 5.6e-14 mm
vs VICON : MATLAB 35.758 mm | soromox 35.760 mm
rollout cost : 1308.5 s for 2.06 s simulated (~635 s/simulated-s)
```

Both engines track VICON almost identically and agree with each other to
~0.01mm, starting cold from a genuine rest pose (not warm-started), on the
geometry this repo actually uses.

(For reference, the same fit at 435mm geometry, W6: TIP RMSE 36.366mm MATLAB /
36.369mm soromox, parity 0.0085mm. The 5mm chain-length difference moves the
score by ~0.6mm either way — consistent with every other geometry sweep on
this model, i.e. noise-level, not a fit issue.)

### Plots

tip traces

parity residual

tip 3D

axis error

### Files
- MATLAB: `python_exporter_veronica_last.m`, `stage_veronica_refit_t11_g35178.m`
  (`stage_veronica_refit_t11.m` for the 435mm sibling, `stage_veronica_perm_roll_sweep_t12.m`
  for the offset sweep)
- soromox: `examples/simulation/pcs/simulate_isupport_exp_last.py` (unchanged
  for this geometry variant — everything is read from the export)

cc reviewers — flagging for visibility before merge, not asking to block on it.

PR #116, opened 2026-07-13
Migrate I-SUPPORT to tendon-based actuation

mstoelzle:
## Summary

- make `ISupport` inherit from `TendonActuatedPCS`
- represent pneumatic chambers as segment-local straight tendon routings
- convert chamber pressure to virtual tendon force with an optional effective pressure area
- expose pressure-conjugate chamber-volume coordinates while preserving pressure inputs and channel ordering
- update parameter validation, runtime updates, documentation, examples, and analytical regression coverage

## Why

The previous I-SUPPORT actuation matrix applied a configuration-independent wrench at each chamber cross-section. That captures the axial force and its nominal moment arm in the straight configuration, but it does not account for how the chamber's direction and leverage change as the bellows bends, shears, twists, or extends.

A threadlike transmission is a better reduced-order representation for an I-SUPPORT bellows chamber because the chamber runs longitudinally and off-axis along the deformable segment. Pressure acting on an effective area produces an axial resultant,

`f = p A_eff`,

which follows the chamber path in the same way that a routed tensile element follows its routing. The tendon formulation then integrates the local force direction and moment arm along the complete pneumatic section. This provides several useful properties:

- **Configuration-dependent loading:** the generalized pressure force changes with the current strain state instead of remaining fixed at its straight-configuration value.
- **Distributed actuation:** pressure acts along the bellows length rather than as a lumped cross-section wrench.
- **Consistent PCS subdivision:** when one physical pneumatic section is represented by several PCS children, the same pressure channel contributes through the length-integrated routing of every child.
- **Local pneumatic sections:** generated chamber routings are masked to their owning pneumatic section, so they do not transmit actuation through the robot base, rigid connectors, or preceding pneumatic sections.
- **Virtual-work consistency:** with the reduced-order approximation `V ≈ A_eff l`, pressure work satisfies `p dV = (p A_eff) dl`. Consequently, the pressure actuation matrix is the transpose of the Jacobian of the modeled chamber-volume coordinates, which also makes it compatible with the existing actuation-space dynamics machinery.

This remains a reduced-order model: it assumes a constant effective pressure area and a material-frame chamber-center routing. It does not attempt to model detailed bellows-wall contact, radial expansion, hysteresis, or a pressure-dependent effective area. Those effects can be introduced later through the explicit `chamber_effective_pressure_area` parameter or a richer constitutive transmission model.

## User impact

`ISupportParams` gains an optional `chamber_effective_pressure_area` with one value per pneumatic segment. When omitted, it defaults to `pi * (r_out**2 - r_in**2)`. Existing pressure inputs remain in pascals.

## Validation

- `202 passed` across I-SUPPORT, typed-parameter, and actuation-space focused suites on the final commit
- changed-file Ruff formatting and lint checks passed
- `git diff --check` passed
- full repository suite passed earlier in the implementation cycle (`1107 passed`), before the final cache-removal and helper-renaming cleanup

### DIFF
--- a/docs/api/systems/pcs/isupport.md
+++ b/docs/api/systems/pcs/isupport.md
@@ -4,7 +4,9 @@
 
 ## Overview
 
-`ISupport` is a specialized PCS implementation for 3D pneumatic soft robots with chamber-based actuation. It extends the base `PCS` class with:
+`ISupport` is a specialized tendon-actuated PCS implementation for 3D pneumatic
+soft robots. Each pressure chamber is represented internally by a straight,
+segment-local virtual tendon:
 
 - **3 chambers per segment**: Radially arranged pneumatic chambers
 - **Pressure-based control**: Direct pressure input for actuation
@@ -91,6 +93,8 @@ params = ISupportParams(
     chamber_outer_radius=jnp.array([7.79e-3]),
     chamber_distance=jnp.array([20e-3]),
     chamber_azimuth_angles=(2.0 * jnp.pi * jnp.arange(3) / 3)[None, :],
+    # Optional override. The default is pi * (r_outer**2 - r_inner**2).
+    chamber_effective_pressure_area=None,
 )
 robot = ISupport(
     params,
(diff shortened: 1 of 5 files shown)


Elektron97 (approved):
Everything good!

Issue #129, opened 2026-08-02
Fix gain-optimization history pairing and reject non-finite candidates


mstoelzle:
Found while reviewing #95.

## Problem

Both control-gain optimization scripts evaluate the loss and auxiliary trajectory at the current parameter state, then apply an optimizer update, and finally append the updated parameters together with the pre-update loss/trajectory:

- `paper_results/secVd_control_gain_optimization/code/control_gain_optimization_with_collocated.py`
- `paper_results/secVd_control_gain_optimization/code/control_gain_optimization_with_synergistic.py`

This makes each saved history entry internally inconsistent. In the collocated run, the first update can produce non-finite gains while retaining the finite loss and trajectory evaluated before that update, so the later best-batch selection may save/print NaN gains as if they produced that finite rollout.

## Expected fix

- Pair every loss/trajectory with the exact gain state at which it was evaluated (either store the pre-update parameters or re-evaluate after the update).
- Reject or clearly mark non-finite losses, gains, and trajectories before selecting or saving a best candidate.
- Add regression coverage that detects a one-step offset between parameter and loss history.

Issue #128, opened 2026-08-02
Select best optimization batches independently in collocated and synergistic plots


mstoelzle:
Found while reviewing #95.

## Problem

`paper_results/secVd_control_gain_optimization/code/plot_control_gain_optimization.py` computes `best_batch` only from `data_c["history_loss"]` and then reuses that collocated-run index when plotting the synergistic end-effector trajectory.

The committed datasets currently happen to share the same best-batch index, which hides the coupling. With independently generated results, the synergistic panel can therefore plot a non-optimal batch.

## Expected fix

- Compute separate `best_batch_c` and `best_batch_s` values from their respective loss histories.
- Pass the matching index to each panel.
- Add a regression test with collocated and synergistic loss minima at different batch indices.

PR #135, opened 2026-08-03
Tune collocated integral-error saturation to tendon scale

mstoelzle:
## Summary

- replace the collocated optimizer's fixed `gamma=10` with a physically interpretable default tendon-error scale of `e_sat = 10 mm`, i.e. `gamma = 100 1/m`
- expose the scale as `--integral-error-saturation-scale` (meters) and reject non-finite or non-positive values
- reject non-finite `gamma` values universally in `PIDControl` before positivity/SPD validation
- document the trajectory-based derivation and prominently mark the committed Section V.d data and plots as stale

## How we arrived here

PR #133 corrected the built-in saturation from the former `tanh(gamma * e)` expression to the unit-preserving

```text
sat(e) = tanh(gamma * e) / gamma,
gamma = 1 / e_sat.
```

The collocated control-gain optimizer integrates only tendon-length errors, so every controlled coordinate has units of meters and a scalar `gamma` has units `1/m`. The optimizer had retained `gamma=10`, which now means `e_sat=100 mm`: the full segment length and far outside the observed error regime. It retained 99.3% of the largest initial error, making saturation effectively inactive.

I transformed every saved configuration in the six committed initial and optimized collocated trajectories into tendon coordinates and compared them with the tendon-length setpoints:

| Quantity | Tendon 1 | Tendon 2 | Tendon 3 |
| --- | ---: | ---: | ---: |
| undeformed length | -100.00 mm | -100.00 mm | -100.00 mm |
| setpoint | -85.49 mm | -96.41 mm | -100.05 mm |
| initial error | 14.51 mm | 3.59 mm | -0.05 mm |

Apart from the exceptional 14.51 mm reference step, all observed initial/transient absolute errors across the saved trajectories remain below 3.85 mm.

A 10 mm saturation scale is therefore a deliberate boundary between the large reference step and the ordinary closed-loop error regime:

- at 14.51 mm, `sat(e)=8.96 mm`, reducing integral accumulation by 38.25%
- at 3.85 mm, 95.35% of the error is retained
- at 1 mm, 99.67% is retained

This limits windup during the large step without materially reshaping the ordinary transient and regulation errors.

## Result status and regeneration order

> [!IMPORTANT]
> **All currently committed Section V.d MAT files and derived plots must now be considered stale.** The collocated results were generated with the former non-unit-preserving saturation and the former gamma setting. This PR intentionally does not replace any result artifacts.

The complete Section V.d results should be regenerated only after:

1. this PR has been merged; and
2. the other control-gain-optimization defects in #128 and #129 have been fixed.

A one-iteration temporary smoke run completed the simulation/save path, but it reproduced the non-finite optimized gains tracked in #129. Its setpoint rollout also warned that it had not reached steady state. Those outputs were written only under `/tmp` and are not included here.

## Validation

- `ruff check` and `ruff format --check` on the collocated optimizer
- `python -m py_compile` on the collocated optimizer
- `pytest -q tests/control/test_pid_control.py`: 41 passed
- numerical saturation check at 14.51 mm, 3.85 mm, and 1 mm
- one-iteration collocated end-to-end smoke run to temporary output, confirming the new code path and reproducing the separate #129 blocker

@Michele-Martini, could you please review the physical scale selection and the intended regeneration sequencing?

### DIFF
--- a/docs/development/changelog.md
+++ b/docs/development/changelog.md
@@ -55,6 +55,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Updated the dependency metadata and lockfile for a CUDA 13-enabled JAX and
   PyTorch stack, and declared `ipywidgets` with the optional Open3D extras.
+- Made the Section V.d collocated optimizer derive ``gamma`` from a configurable
+  tendon-length error scale (default 10 mm), validate finite saturation
+  parameters, and document stale committed artifacts.
 
 ### Fixed
 
--- a/paper_results/secVd_control_gain_optimization/README.md
+++ b/paper_results/secVd_control_gain_optimization/README.md
@@ -4,6 +4,37 @@ This case compares gain optimization for collocated actuation-space control and
 synergistic operational-space control. Each generator writes a MAT result under
 its controller-specific directory in `data/`; the plotter combines both files.
 
+> [!WARNING]
+> The committed Section Vd MAT files and derived plots are stale. In particular,
+> the collocated results were generated with the former, non-unit-preserving
+> integral-error saturation and the former `gamma=10` setting. Do not treat the
+> current files as canonical results. Regenerate the complete Section Vd results
+> only after the saturation-scale change has been merged and the other open
+> control-gain-optimization issues (#128 and #129) have been fixed.
+
+## Collocated integral-error saturation
+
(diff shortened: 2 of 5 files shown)


Michele-Martini (approved):
Makes sense, thanks for letting me know.

PR #139, opened 2026-08-04
Harmonize PCS and GVS parameter models

mstoelzle:
## Summary

This PR harmonizes PCS, PlanarPCS, and GVS around a shared public parameter vocabulary and a common link-oriented construction/update workflow.

The central model is now:

```text
PCSParams / PlanarPCSParams
├── link: ContinuumLinkParams
├── base_pose
└── gravity

GVSParams
├── link: ContinuumLinkParams
├── joint: JointParams
├── base_pose
└── gravity
```

Generalized link stiffness and damping matrices are the canonical runtime parameters. Isotropic Young's modulus, shear modulus, and material damping remain ergonomic construction and differentiable identification variables through a separate `IsotropicMaterialParams` PyTree.

## Motivation

PCS and GVS previously exposed substantially different names and ownership rules for the same physical concepts. GVS construction used Greek/abbreviated fields and GVS-local link/cross-section structures, while PCS stored mostly flat system parameters and a global damping matrix. This made it difficult to:

- transfer code and intuition between PCS and GVS;
- reuse typed parameter definitions;
- update one link without rebuilding unrelated system state;
- distinguish construction inputs from canonical runtime mechanics;
- optimize interpretable material parameters without duplicating them in runtime params;
- support joint stiffness and damping consistently in GVS; and
- extend shared link and joint concepts to future system families.

This is intentionally a pre-1.0 clean break. It removes the divergent APIs instead of preserving aliases that would keep both vocabularies alive.

### Shared continuum components

New public components live in `soromox.systems.components` and are re-exported from `soromox.systems`:

- `ContinuumLinkParams`
- `JointParams`
- `CrossSectionParams`
- `CrossSectionGeometry`
- `IsotropicMaterialParams`
- `LinearProfile`
- `LinkSpec`
- `JointSpec`
- `shear_modulus_from_poisson_ratio`

The package is split by responsibility:

- `cross_sections.py`: coefficient packing, constant/linear profiles, and area/inertia geometry;
- `links.py`: canonical per-link runtime params, link specifications, validation, and matrix construction inputs;
- `joints.py`: joint definitions, DOF metadata, factories, validation, stiffness, and damping;
- `materials.py`: isotropic material PyTree and Poisson-ratio conversion.

GVS keeps only genuinely GVS-specific concepts: `GVSSegment`, `StrainBasisSpec`, basis evaluation, quadrature, padded operands, and GVS structure/runtime machinery.

### Canonical link and joint mechanics

`ContinuumLinkParams` owns length, density, reference strain, cross section, generalized stiffness, and generalized damping. Both matrices are finite, symmetric, per-link arrays with matching shapes.

PCS assembles its global matrices from link-local blocks. GVS interleaves joint and link blocks so both joint stiffness and joint damping contribute to the global dynamics. Link blocks in GVS are zero-padded beyond each link's active basis coordinates.

The old PCS cross-link damping representation is deliberately removed: damping now has the same ownership and update semantics as stiffness.

### Construction specifications

`LinkSpec` accepts exactly one stiffness source:

- `young_modulus` plus `shear_modulus`; or
- an explicit generalized `stiffness` matrix.

It similarly accepts exactly one damping source:

- `material_damping_coefficient`; or
- an explicit generalized `damping` matrix.

Circular, rectangular, and other shared cross-section factories populate a single coefficient-based `CrossSectionParams`; there is no profile-parameter class hierarchy. `LinearProfile(base, tip)` represents linear geometry changes.

`reference_strain` now belongs to the link. `StrainBasisSpec` describes only GVS basis selection and order.

### Differentiable isotropic material mapping

PCS, PlanarPCS, and GVS expose the same APIs:

```python
stiffness, damping = robot.link_matrices_from_material(material)
updated = robot.with_isotropic_material(material)
```

Unit-response operators implement:

```text
K_i = E_i K_i,E + G_i K_i,G
D_i = eta_i D_i,eta
```

This keeps material identification cheap, batched, JIT-compatible, and differentiable without storing a second material representation inside system params. `with_isotropic_material` returns a new robot and does not mutate the original or retain the caller-owned material PyTree.

Geometry changes refresh unit-response operators but leave explicitly supplied canonical matrices unchanged. Reapplying `with_isotropic_material` is the explicit step that rebuilds them.

### Updates and immutable replacement

All three continuum systems now support link-local updates:

```python
robot = robot.update_link_params(
    stiffness=1.1 * robot.params.link.stiffness,
    damping=0.9 * robot.params.link.damping,
)
```

GVS additionally supports:

```python
gvs = gvs.update_joint_params(
    damping=1.2 * gvs.params.joint.damping,
)
```

Nested immutable replacement remains available:

```python
params = robot.params.replace(
    link=robot.params.link.replace(density=new_density)
)
robot = robot.with_params(params)
```

### PCS from isotropic material properties

```python
from soromox.systems import PCS, LinkSpec, shear_modulus_from_poisson_ratio

young = 1.0e6
shear = shear_modulus_from_poisson_ratio(young, poisson_ratio=0.45)

pcs = PCS.from_links([
    LinkSpec.circular(
        length=0.20,
        radius=0.012,
        density=1000.0,
        young_modulus=young,
        shear_modulus=shear,
        material_damping_coefficient=1.0e4,
        reference_strain=[0.0, 0.0, 0.0, 1.0, 0.0, 0.0],
    )
])
```

### GVS with tapered geometry and joint mechanics

```python
from soromox.systems import (
    GVS,
    GVSSegment,
    JointSpec,
    LinearProfile,
    LinkSpec,
    StrainBasisSpec,
)

gvs = GVS.from_segments([
    GVSSegment(
        link=LinkSpec.rectangular(
            length=0.20,
            height=LinearProfile(base=0.030, tip=0.020),
            width=0.025,
            density=1000.0,
            young_modulus=1.0e6,
            shear_modulus=3.4e5,
            material_damping_coefficient=1.0e4,
            reference_strain=[0.0, 0.0, 0.0, 1.0, 0.0, 0.0],
        ),
        joint=JointSpec.revolute(
            axis="z",
            stiffness=[[0.30]],
            damping=[[0.02]],
        ),
        basis=StrainBasisSpec(
            type="legendre",
            strain_selector=("kappa_y", "sigma_x"),
            basis_order=1,
        ),
        num_gauss_points=7,
    )
])
```

### Optimizing interpretable material values

```python
import jax
import jax.numpy as jnp
from soromox.systems import IsotropicMaterialParams

material = IsotropicMaterialParams(
    young_modulus=jnp.array([1.0e6]),
    shear_modulus=jnp.array([3.4e5]),
    material_damping_coefficient=jnp.array([1.0e4]),
)

def loss(candidate_material):
    candidate = robot.with_isotropic_material(candidate_material)
    return jnp.mean((candidate.stiffness_matrix() - target_stiffness) ** 2)

value, gradient = jax.value_and_grad(loss)(material)
```

The user guide also includes positive log-space parameterization, a complete Optax loop, geometry/material co-optimization, and direct generalized-matrix optimization.

## Documentation

Rather than a narrowly scoped material-only tutorial, this PR adds a balanced
component/parameter documentation set:

- **Continuum components** documents shared cross sections, links, joints, material models, ownership, dimensions, and construction alternatives.
- **Parameters and optimization** documents construction, immutable updates, material identification, log-space/Optax optimization, geometry co-optimization, and PCS/GVS-specific notes.
- **PCS/GVS parameter migration** provides a concise conversion checklist for imports, constructors, nested field paths, GVS argument names, link-local damping, and material updates.

The enduring component and parameter pages are integrated into the existing API and user-guide navigation and cross-linked from the quick start, examples, parameter utilities, PCS pages, and GVS pages. The release-specific migration guide is intentionally linked only from the Unreleased changelog. Public APIs introduced or changed here include package-style docstrings with arguments, return values, validation behavior, and examples.

The Unreleased changelog entry separately highlights the public additions,
behavior changes, fixes, and breaking changes, with a direct link to the compact
migration guide.

### Imports

Shared specifications are no longer GVS-owned.

```python
# Before
from soromox.systems.gvs import LinkSpec, JointSpec

# After
from soromox.systems import LinkSpec, JointSpec
# or: from soromox.systems.components import LinkSpec, JointSpec
```

Keep importing `GVSSegment`, `StrainBasisSpec`, and GVS structures from the GVS namespace or the top-level systems exports.

### PCS and PlanarPCS params

Replace flat constructors with `from_links` or `params_from_links`:

```python
# Before
params = PCSParams(
    length=lengths,
    radius=radii,
    density=densities,
    young_modulus=young,
    shear_modulus=shear,
    damping=global_damping,
    reference_strain=reference,
)

# After
params = PCS.params_from_links([
    LinkSpec.circular(
        length=L,
        radius=r,
        density=rho,
        young_modulus=E,
        shear_modulus=G,
        material_damping_coefficient=eta,
        reference_strain=xi_ref,
    )
    for L, r, rho, E, G, eta, xi_ref in link_rows
])
```

Split an old block-diagonal PCS damping matrix into one generalized damping block per link. Cross-link damping terms are no longer supported.

Field access changes from flat fields to link-owned fields:

```python
robot.params.length             -> robot.params.link.length
robot.params.density            -> robot.params.link.density
robot.params.reference_strain   -> robot.params.link.reference_strain
robot.params.stiffness          -> robot.params.link.stiffness
robot.params.damping            -> robot.params.link.damping
```

### GVS link specifications

Replace abbreviated/Greek construction names with public descriptive names:

```text
E       -> young_modulus
nu      -> derive shear_modulus with shear_modulus_from_poisson_ratio
rho     -> density
eta     -> material_damping_coefficient
L       -> length
r_i/r_f -> radius=LinearProfile(base=..., tip=...)
```

Move `reference_strain` from `StrainBasisSpec` to `LinkSpec`. Replace legacy basis `active`/`orders` arguments with `strain_selector`/`basis_order`.

### Cross sections

Replace endpoint-specific geometry fields and GVS-local cross-section params with shared factories. Constant scalars and `LinearProfile` values are packed into `CrossSectionParams.coefficients` by `LinkSpec`; callers normally do not construct coefficient arrays by hand.

### Runtime material updates

Do not expect Young's modulus, shear modulus, or damping coefficient inside `robot.params`. Keep them in a caller-owned `IsotropicMaterialParams` and explicitly apply them:

```python
material = material.replace(young_modulus=1.1 * material.young_modulus)
robot = robot.with_isotropic_material(material)
```

For anisotropic or coupled identification, optimize and replace `params.link.stiffness` and `params.link.damping` directly.

### Joint stiffness and damping

GVS joint values are now stored in `params.joint.stiffness` and `params.joint.damping` and contribute to global assembly. Existing joint specs that omit them continue to construct zero joint matrices.

### Compatibility policy

No aliases for `GVSLinkParams`, old flat PCS fields, GVS-local shared specs, Greek construction fields, or the global PCS damping matrix are added. Downstream code should migrate in one step using the mappings above.

## Validation

Completed locally on the feature branch:

- focused shared-component, PCS, PlanarPCS, and GVS tests;
- autodiff, scalar/per-link material, matrix-bypass, joint assembly, geometry-refresh, and immutable replacement coverage;
- Ruff checks, Python compilation checks, and staged whitespace validation;
- strict MkDocs and Zensical documentation builds;
- all project example model workloads at their complete simulation horizons, including full encoded videos where headless renderers support them;
- all four Section IV-A SoRoMoX simulations, the available planar/spatial
  PyElastica simulations, a complete 3,001-frame planar video, and all four
  comparison plots;
- the complete 100-step Section V-A parameter identification, all 22 residual evaluations, and all four residual plots;
- all Section V-C configuration-space generators (setpoint, slow/fast
  trajectory, and regulation-to-tracking), all controller rollouts, all derived
  plots, the complete 12-DOF operational-space trajectory, the matched
  full/partial feedback-linearization comparison, finite-array audits, metrics,
  canonical paper figure, and composite PDF/SVG;
- both Section V-D gain-optimization programs at their configured iteration
  counts: synergistic completed two finite iterations; collocated reproduced
  the known iteration-1 NaN guard and saved its partial result;
- both Section V-E CBF/no-CBF simulations, finite-array audits, and publication
  plots;
- Section V-F released-checkpoint evaluation (95.3% success), a fresh complete
  105-step policy rollout with finite arrays, and reward plotting from every
  committed training log;
- three complete PPO rollout/update cycles through 38,400 timesteps, plus a
  finite reward record at 40,192 timesteps; the user then requested that the
  multi-hour million-step CPU training be stopped;
- 868 repository tests passed without failure (68% of 1,255 collected) before
  the user requested that the remaining unusually long numerical tests be
  skipped.

Known baseline and environment-specific limitations:

- Section IV-B targets a GPU, but this host exposes only a CPU. An exact paper-grid
  CPU attempt completed every batch size for one- and two-link systems and part
  of the four-link sweep before being stopped as an impractical and invalid
  substitute for the publication GPU benchmark.
- SoRoSim reproduction requires MATLAB plus SoRoSim, neither of which is installed here.
- the PyElastica tendon case imports an external `muscles` module that is not present in this repository/environment.
- the Section V-D fresh-result plotter cannot consume the intentionally partial
  collocated MAT output after the known NaN guard because `q_des_ts` is absent;
  canonical complete result plotting remains separate from this known optimizer
  behavior.
- browser/display-backed Viser and Open3D recording paths require an interactive
  or headless graphics service. The Section V-E renderer loaded the fresh
  trajectory and built the migrated robot geometry, then blocked in the
  unavailable macOS graphics service without producing an MP4; the corresponding
  simulations and noninteractive plots were validated separately.
- repository-wide Ruff lint passes. The full-tree format audit identifies nine
  pre-existing out-of-scope files that would be reformatted; changed files in
  this PR are formatted.

## Breaking changes

This PR changes public construction signatures and parameter field paths for PCS, PlanarPCS, and GVS. It also removes cross-link PCS damping and GVS-local ownership of shared link/joint specifications. These changes are deliberate and covered by the migration guide above.


tools/benchmarks/_benchmark_common.py
```diff
@@ -149,23 +149,21 @@ def _planar_pcs_factory(num_segments: int, gauss_points: int = 5) -> PlanarPCS:
     lengths = jnp.full((num_segments,), 0.12)
     radii = jnp.full((num_segments,), 0.015)
     rho = 1070.0 * jnp.ones((num_segments,))
```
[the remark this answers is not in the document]
mstoelzle: Thanks. I verified that these factories currently run during benchmark setup, outside both JIT tracing and the timed benchmark functions, so the reported ConcretizationTypeError was not an existing runtime failure. The JAX allocations and scalar device-to-host conversions were nevertheless unnecessary overhead and made the factories less reusable. Commit 697e329 replaces them with Python scalars in both the PlanarPCS and PCS factories; a focused construction check for both factories passes.

src/soromox/systems/components/cross_sections.py
```diff
@@ -0,0 +1,159 @@
...
+def section_properties(
+    geometry: CrossSectionGeometry | int,
+    dimensions: Array,
+) -> tuple[Array, Array, Array, Array]:
+    """Return ``(I_x, I_y, I_z, area)`` for a solid section.
+
+    The material-frame x-axis is longitudinal. Rectangular dimensions are
+    ordered ``[height, width]`` and elliptical dimensions are ordered
+    ``[semi_major, semi_minor]``.
+
```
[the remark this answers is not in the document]
mstoelzle: Agreed—this was a real public composition bug. Commit 697e329 makes GVS.cross_section_geometry return rectangular dimensions in the shared [height, width] order, documents that contract on the common robot/component APIs, and updates every Open3D rectangular consumer so rendering keeps the intended width/height orientation. A non-square rectangle regression test now checks the returned order and the resulting I_y/I_z formulas through section_properties.

vdperfetta01 (approved):
The merge commit looks well organized, with the new shared components and clearer separation between construction inputs and runtime parameters make the codebase easier to understand and more flexible for future expansion.

After a careful read of the changes I was not able to find any typos or other mistakes. Instead, the code and docs appear consistent and well-written. I would probably just double-check the tests (especially the comparison between gvs and pcs) to certify the consistency with the new harmonization.

docs/api/systems/continuum-components.md
```diff
@@ -0,0 +1,204 @@
...
+
+joint = JointSpec.revolute(
+    axis="z",
+    stiffness=jnp.array([[0.3]]),
+    damping=jnp.array([[0.02]]),
+)
+```
+
+The GVS global matrices interleave joint and link contributions before
+projection into active coordinates. PCS has no separate joint component.
```
Michele-Martini, line 186: Forgive me if I am saying nonsense, but why is the PCS not using links the same way of GVS? If I am understanding correctly, the joints in the GVS acts between two links, like a rigid robot, acting on the dofs. The same logic can be applied to the PCS as well, I guess. In this way aren't we implying a fixed joint among the links? Just for the sake of sharing the same conventions.

mstoelzle: Yeah, this is exactly the goal of this PR: harmonize more the interface between PCS and GVS:

>  If I am understanding correctly, the joints in the GVS acts between two links, like a rigid robot, acting on the dofs.
Indeed, this is also my understanding of GVS: you have soft links that are connected by joints similar to rigid arms.

For PCS, we are at this time basically _just_ adding the soft links that are connected by assumed rigid joints. However, making the interface more coherent is a first step. It leaves the door open for additional changes down the line, such as adding joints to connect the soft links in PCS, or adding support for non-cylindrical cross sections to PCS.

mstoelzle: @Michele-Martini Does this answer your comment? Or do you see any issue with this revised interface?

Michele-Martini: It does, thanks! My only comment is that the parametrization API has been changing substantially multiple times in a short period, so maybe if the plan is to extend the joints also to PCS, it could be worth doing in this PR, if it not too much work (the PR already has big changes). Just a thought

mstoelzle: Thanks for sharing your thoughts, @Michele-Martini! 

- I agree that the parametrization API has been changing quite a bit. The idea was to not care about backwards compatibility during development and converge to an API at release time that we can keep more stable afterwards (which we are hopefully at now). 
- Adding the joints to PCS shouldn't really change the API. It should "only" add the joints field/attribute to the PCS parameters which is None/deactivated by default. Therefore, it should be fully backwards compatibility if/when we add it.

src/soromox/systems/components/joints.py
```diff
@@ -0,0 +1,343 @@
...
+    Raises:
+        ValueError: If ``joint_type`` is unknown.
+    """
+    try:
+        return JOINT_DOF[joint_type]
+    except KeyError as exc:
+        raise ValueError(f"Unknown joint type {joint_type!r}.") from exc
+
+
+class JointParams(BaseSystemParams):
```
Michele-Martini, line 58: Is there a reason why JointParams and JointSpec are different classes? I see that params inherit from BaseSystemParams, however it may be compact to have one class like "Joint" with everything in it

src/soromox/systems/pcs/params.py
Michele-Martini, line 1: Forgive me again in case of a misunderstanding on my part, but this file shows that all the *Params API that in the previous version contained the various parameters, now just mostly contains ContinuumLinkParams. Could this *Params container be deprecated and no longer needed?

mstoelzle: The reason is twofold:
- To give `PCSParams` a structure closely aligned with `GVSParams`, namely a typed dataclass containing a link field.
- To preserve the option of adding a joint field of type `JointParams` in the future, as discussed.

src/soromox/systems/components/joints.py
mstoelzle: - `JointParams` only contains the user-configurable paramaters (i.e., the ones that can be updated also without requiring a new construction of the system class. For example, via optimization). This mirrors the `PCSParams` or `GVSParams` dataclasses.
- `JointSpec` is the actual component class. This mirrors the system classes such as `PCS` or `GVS`. 

One difference is that we currently don't put the `JointParams` in a dedicated `components/params.py` file. This indeed could be something to consider.

Michele-Martini:
This PR introduces some very nice changes in the PCS and GVS classes, by making their definitions and parametrizations uniform, coherent, and accessible. I like the direction, where the GVS stands as the most complent and expressive model, while PCS and planar PCS represents lower-dimensional simplifcations that should benefit from higher computational speed, but sharing the same construction of links + (eventually) joints.

My few comments are general and moslty address clarity for users. I have a simple proposal for the material, just for discussion:
material.py -> alternatively, the class accepts E and nu and internally computes G. *why*: more compact. It skips the public function and make the material definition a one-liner if someone puts only E and nu. Likewise for LinkSpec.circular().

What do you think?

mstoelzle:
Thanks, Michele — good suggestion. I added this in [3e02ac5](https://github.com/tud-phi/soromox/commit/3e02ac5).

`IsotropicMaterialParams` and the `LinkSpec` factories now accept Young’s modulus together with exactly one of shear modulus or Poisson’s ratio. When Poisson’s ratio is supplied, `G = E / (2 (1 + nu))` is evaluated inside the differentiable material mapping, so gradients can still be traced directly with respect to `E` and `nu`. The selected parameterization is encoded in the PyTree structure, and the existing `E`/`G` form remains supported.

Material damping is now optional as well; omitting it produces zero link damping.

PR #146, opened 2026-08-14
Stabilize differentiable numerics and Lie operators; harden Section Vd workflows

### LINKED ISSUE #128: Select best optimization batches independently in collocated and synergistic plots

### LINKED ISSUE #129: Fix gain-optimization history pairing and reject non-finite candidates

Michele-Martini:
## Summary

This PR stabilizes differentiable numerical and Lie-group operations across SoRoMoX and completes the Section Vd control-gain optimization workflow relative to `main`.

- Add reverse-mode-safe numerical primitives, strict singularity diagnostics, and pytree finiteness reporting for degenerate model configurations.
- Stabilize `SO(3)`, `SE(2)`, and `SE(3)` exponential/logarithmic maps, Jacobian coefficients, constant-strain adjoints/tangents, and PlanarPCS pose integration with dtype-aware near-zero series and explicit analytic derivatives.
- Fuse repeated constant-strain work across PlanarPCS, PCS, and GVS, including exact contracted `Jd @ qd` paths for dynamics-only recurrences and arclength-independent PCS adjoint-power preparation.
- Refactor the Section Vd optimizers around a shared, testable optimization loop and shared result/output helpers.
- Improve Section Vd setpoint generation, history pairing, candidate validation, plotting selection, serialization, and offline trajectory post-processing.
- Add explicit Open3D visualization failure reporting and update the CUDA 13 dependency metadata.

## Section Vd control-gain optimization

The collocated and synergistic examples now form a coherent, reproducible workflow:

- The synergistic objective tracks the operational-space pose regulated by its controller, while optimization history preserves the parameters associated with each evaluated loss and trajectory.
- The collocated example uses a stable asymmetric tendon reference, verifies that the generated state has reached steady state, and rejects tendon paths too close to collapse before reverse-mode optimization can diverge.
- Both cases share CLI parsing, path resolution, overwrite protection, diagnostic-figure finalization, MAT summary serialization, and animation pickle serialization.
- Diagnostic figures are always written to the configured output directory; the obsolete `--save-figures` switch is no longer needed.
- Core optimization data is saved before metrics, plotting, or interactive rendering, so result files remain available even if visualization is interrupted.
- The common output schema keeps optimization histories, initial/best trajectories, targets, controller parameters, and animation metadata aligned between cases.
- An offline post-processing renderer produces tracking PDFs/PNGs, MP4 animations, optional GIFs, and intermediate metrics from saved MAT trajectories without rerunning optimization. It is a separate rendering step, not a self-contained data generator.
- The reproduction commands and defaults match the 100-iteration paper-result histories.

The workflow also handles interrupted optimization safely by retaining the finite completed prefix and reporting an explicit stop reason. Non-finite candidates and empty histories receive clear diagnostics, and collocated and synergistic plots select their best batches independently.

## Numerical and Lie-operator changes

Several expressions had finite forward values but inaccurate or unsafe reverse-mode derivatives at zero and small rotations. Safe primitives now sanitize singular inputs before evaluation, while strict-singularity mode exposes raw quotients for diagnosis.

Lie-group and constant-strain paths use shared scalar coefficients, dtype-aware Taylor series, direct rotation/Jacobian blocks, and explicit analytic tangent derivatives. No `jvp`, `grad`, `jacrev`, or other autodiff transform is used in these production paths; autodiff remains a test oracle.

The API is organized as follows:

- `lie_algebra.jacobian_coefficients` is a public module for reusable stable forward and inverse left-Jacobian coefficients.
- `lie_algebra.constant_strain.se2` and `.se3` are public algebra-specific modules with unsuffixed `adjoint`, `adjoint_inverse`, `tangent`, `tangent_derivative`, and `operators` functions.
- `operators` returns a fixed `ConstantStrainOperators` bundle for callers that need multiple related expressions at one strain and arclength.
- The `constant_strain` package root exports only `se2`, `se3`, and the shared `ConstantStrainOperators` result.
- Reduced-polynomial coefficient selection and matrix-power recurrences remain private in `constant_strain._shared` because they are implementation details rather than a stable user abstraction.

PlanarPCS, PCS, and GVS share relevant intermediates without materializing unused dynamics quantities, while public kinematics continues to return complete Jacobian derivatives. Internal state is named explicitly according to whether it contains `Jd` or the contracted `Jd @ qd`.

## New regressions that fail on `main`

The complete PR test suite passes on this branch. A detached `main` worktree was also evaluated with adapted copies of the newly added oracle tests so the comparison exercises `main` behavior rather than failing only because new modules are absent.

The following newly added tests pass on this PR and expose numerical errors on `main`:

- `test_log_se2_matches_analytic_reverse_mode_jacobian_near_identity`: fails for float32 and float64 because the translation derivative is inaccurate near identity.
- `test_log_se3_arbitrary_axis_hessian_at_series_boundary`: fails at and just above the cutoff for float32 and float64.
- `test_log_se3_arbitrary_axis_round_trip_jacobian`: fails for float32 because the translation rows deviate from the identity Jacobian.
- `test_constant_strain_operators_match_independent_series`: fails at zero rotation for both SE(2) and SE(3), in float32 and float64, because `main` omits higher-order rotation/translation contributions from the tangent Jacobian.

Additional tests for the public coefficient module, bundled operators, and prepared SE(3) adjoint-power path cannot be collected unchanged on `main` because those APIs do not exist there; they pass on this branch. The adapted main comparison deliberately separates those API-addition checks from the functional failures listed above.

## Performance against `main`

On the CPU/JAX 0.11.0 comparison used for this PR:

- Across five fresh processes, the 40-case Lie benchmark reduces geometric-mean steady-state execution time by 24.3% and compilation time by 5.3%.
- In a final interleaved five-pair comparison, complete non-rollout PlanarPCS/PCS/GVS method suites change execution by -10.4%/-12.2%/+0.1% and compilation by -4.3%/+6.5%/+6.4% (negative execution values are faster).
- One-second PlanarPCS rollouts improve by 24.7%, 27.8%, and 36.0% for one, two, and four segments, respectively (29.3% aggregate).
- One-second PCS rollouts improve by 16.6%, 27.3%, and 33.3% for one, two, and four segments, respectively (26.3% aggregate).
- One-second GVS rollouts change by 0.0%, -21.0%, and -7.6% for one, two, and four segments, respectively (10.1% faster in aggregate).

The algebra-specific package split, documentation changes, and variable renames do not alter the compiled numerical expressions underlying those measurements.

## Validation

- 162 Lie-algebra tests pass, including value, reverse-mode Jacobian, reverse-over-reverse Hessian, cutoff-boundary, arbitrary-axis, batching, strict-mode, and analytic directional-derivative coverage.
- 328 PCS, PlanarPCS, GVS, and 2D/3D coherence tests pass, with one pre-existing expected xfail.
- 12 focused PCS integration/dynamics tests pass after the final prepared-adjoint-power naming cleanup.
- Ruff formatting/lint, strict documentation build, Python compilation, changelog extraction, and whitespace checks pass.
- Focused Section Vd, plotting, and Open3D regression tests pass.
- Production Lie, PlanarPCS, PCS, and GVS paths contain no runtime autodiff calls.
- Benchmark and paper-result artifacts are reproducible from the documented workflows.

Closes #128  
Closes #129


tests/lie_algebra/test_lie_algebra_branch_safety.py
mstoelzle, line 1: The tests contained in this test files be merged into the existing test files in `tests/lie_algebra`

src/soromox/utils/numerics.py
mstoelzle, line 1: Rename from `_numerics.py` to `numerics.py`

src/soromox/utils/numerics.py
```diff
@@ -20,3 +63,195 @@ def eps_for_dtype(eps: float | Array | None, dtype: jnp.dtype) -> Array:
...
+    if strict_singularities_enabled():
+        return numerator / denominator
+
+    eps_arr = eps_for_dtype(eps, denominator.dtype)
+
+    regular = jnp.abs(denominator) > eps_arr
+    denominator_safe = jnp.where(regular, denominator, jnp.ones_like(denominator))
+    quotient = numerator / denominator_safe
+
+    return jnp.where(regular, quotient, jnp.broadcast_to(fallback, quotient.shape))
```
mstoelzle, line 228: Does't give this also the usual `jnp.where` issues that NaN/singularities still appear in the gradients? Would it be better to use `jnp.cond` here?

tests/actuation/test_mckibben.py
mstoelzle, line 1: I would say let's call this test file just `test_mckibben.py`. Also, reflect on if more tests for the McKibben actuation are necessary.

tests/utils/test_numerics_safe_ops.py
mstoelzle, line 1: This file should be just called `test_numerics.py`

mstoelzle (changes requested):
General comments:
- Make sure that all singularity issues that we are fixing in this PR are covered by added tests to prevent future regressions
- Should `tolerance.py` make more use of `utils.py` for setting the tolerances?

tests/actuation/test_mckibben.py
mstoelzle: Implemented in 8b40b38. The file is now tests/actuation/test_mckibben.py. I also expanded coverage beyond the private geometry hooks: the public coordinate and moment-matrix Jacobians are tested for both regular and coincident attachment geometry.

src/soromox/utils/numerics.py
mstoelzle: Implemented in 8b40b38. The module is now public as src/soromox/utils/numerics.py, and all internal imports plus the utils package export were updated.

src/soromox/utils/numerics.py
mstoelzle: The elementwise jnp.where calls are safe here because denominator_safe is formed before division, so neither evaluated arm contains a singular quotient. lax.cond requires a scalar predicate and would not preserve array-wise fallback semantics. I documented this in numerics.py and added a mixed-array reverse-mode Jacobian regression in tests/utils/test_numerics.py; it passes for both numerator and denominator derivatives.

mstoelzle:
Addressed the general review in commit 8b40b38.

- I audited the singularity fixes and added direct site-level regressions where coverage had previously been only indirect: duplicate reference-trajectory timestamps; straight PlanarHSA and PlanarPCS inverse kinematics; zero-length articulated tips; collapsed threadlike tangents for PCS, PlanarPCS, and GVS; centered i-Support chamber inertia; and the public McKibben coordinate and moment-matrix paths. Existing rotation/Lie-algebra and optimization-loss tests cover the remaining changed sites. The focused result is 183 passed.
- I considered coupling tolerance.py to numerics.py and kept them separate deliberately. Tolerance.atol/rtol defines host-side comparison policy for tests, whereas eps_for_dtype and the safe helpers define dtype-aware runtime singularity behavior inside JAX traces. Reusing comparison tolerances as branch thresholds would conflate two different contracts and could change model behavior when test tolerances are tuned.

Formatting, lint, and git diff checks are clean.

mstoelzle (approved):

paper_results/secVd_control_gain_optimization/code/gain_optimization_loop.py
```diff
@@ -0,0 +1,185 @@
...
+    def stacked_aux(self, key: str) -> Array:
+        """Stack one auxiliary field across iterations.
+
+        Args:
+            key: Auxiliary dictionary key, e.g. ``"q_ts"``.
+
+        Returns:
+            Array with a leading iteration axis.
+        """
+        return jnp.stack([entry[key] for entry in self.aux], axis=0)
```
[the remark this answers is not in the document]
Michele-Martini: I have currently fixed this. Gonna push shortly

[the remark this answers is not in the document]
mstoelzle: Fixed in bba2231 and 3cd47d0. Both history accessors now share the documented empty-history guard, and both generators print the stop reason and exit before saving or plotting. I also addressed the suppressed update-overflow finding: a finite evaluated iterate is recorded before validating the next optimizer state, invalid updates are never adopted, and no unused final update is computed. Regression coverage exercises both empty accessors, a first-update overflow, and the final-evaluation case.

src/soromox/systems/hsa/planar_hsa.py
```diff
@@ -1393,25 +1394,32 @@ def inverse_kinematics_end_effector(self, chiee: Array) -> Array:
...
+                    safe_divide(px, self.Lmax, self.global_eps),
+                    safe_divide(py, self.Lmax, self.global_eps),
                 ]
             )
-        )
+
+        def _curved_branch(_: None) -> Array:
+            divisor = jnp.cos(th) - 1
+            sin_ratio = safe_divide(jnp.sin(th), divisor, self.global_eps)
+            return safe_divide(th, 2 * self.Lmax, self.global_eps) * jnp.stack(
```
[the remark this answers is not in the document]
mstoelzle: Fixed in 3cd47d0. PlanarHSA now recovers the virtual-backbone strain through the shared SE(2) logarithm, whose inverse left Jacobian uses the stable half-angle form (theta/2) cot(theta/2) I - (theta/2) J instead of 1 - cos(theta). The new public inverse-kinematics regression uses an angle at 2 * global_eps and verifies the complete recovered strain.

src/soromox/utils/numerics.py
```diff
@@ -0,0 +1,261 @@
...
+``jnp.where`` and ``lax.select`` evaluate both sides and their transpose
+multiplies the unused side by a materialized zero, so a ``NaN`` produced there
+survives as ``0 * NaN``::
+
+    jnp.where(x > 0, jnp.sqrt(x), 0.0)       # broken at x == 0
+    x_safe = jnp.where(x > 0, x, 1.0)        # safe: no branch is ever singular
+    jnp.where(x > 0, jnp.sqrt(x_safe), 0.0)
+
+``lax.cond`` does not have this problem: it differentiates only the taken
+branch, including under ``vmap``.
```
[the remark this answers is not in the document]
mstoelzle: Fixed in 3cd47d0. The documentation now distinguishes scalar lax.cond from vmap batching: batched predicates may lower to a select and execute both branch computations, so vectorized branch inputs must remain forward-safe.

src/soromox/systems/pcs/planar_pcs.py
```diff
@@ -846,24 +847,31 @@ def _inverse_kinematics_single_segment(chi_rel_i: Array, s_i: Array) -> Array:
...
+                        jnp.zeros((), dtype=th.dtype),
+                        safe_divide(px, s_i, self.global_eps),
+                        safe_divide(py, s_i, self.global_eps),
+                    ]
+                )
+
+            def _curved_branch(_: None) -> Array:
+                divisor = jnp.cos(th) - 1
+                sin_ratio = safe_divide(jnp.sin(th), divisor, self.global_eps)
+                return safe_divide(th, 2 * s_i, self.global_eps) * jnp.stack(
```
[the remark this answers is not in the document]
mstoelzle: Fixed in 3cd47d0. PlanarPCS now maps each relative tip transform through the shared stable SE(2) logarithm and divides the integrated twist by segment length, removing the cancellation-prone sin(theta)/(cos(theta)-1) formula. Regressions at 1e-10 and 1e-8 radians verify all recovered strain components and the shared logarithm reverse-mode Jacobian.

mstoelzle:
PR 146 has been split into two focused draft PRs:

- #148 — numerical, Lie-algebra, gradient-safety, and PlanarPCS/PCS/GVS performance changes: https://github.com/tud-phi/soromox/pull/148
- #149 — Section Vd case-study and Open3D renderer changes, stacked on #148: https://github.com/tud-phi/soromox/pull/149

PR #149 includes `Closes #128` and `Closes #129`. Closing this PR as superseded; its source branch is retained for reference.

PR #148, opened 2026-08-17
Stabilize numerics, Lie operators, and system performance

mstoelzle:
## Summary

Compared with `main`, this PR contains the numerical/Lie-algebra safety and system-kernel work split out of PR #146. It makes degenerate configurations and small rotations safer for forward and reverse-mode computation, reorganizes the public Lie/constant-strain API, and reduces repeated work in PlanarPCS, PCS, and GVS. The Section Vd control-gain optimization workflow from PR #146 is not part of this PR.

- Adds reverse-mode-safe numerical primitives, strict singularity diagnostics, and pytree finiteness reporting.
- Hardens inactive square-root and quotient inputs in SO(3)/SE(3) magnitude and quaternion-conversion branches, covering nested `grad(vmap(jvp(...)))` at straight and identity configurations.
- Stabilizes `SO(3)`, `SE(2)`, and `SE(3)` exponential/logarithmic maps, left-Jacobian coefficients, constant-strain operators, and PlanarPCS pose integration.
- Uses explicit analytic directional derivatives for Lie Jacobians and constant-strain tangent derivatives; production paths do not invoke runtime autodiff for these expressions.
- Organizes public constant-strain APIs under `constant_strain.se2` and `constant_strain.se3`, with shared operator bundles, while keeping generic matrix-polynomial mechanics private.
- Fuses repeated constant-strain work and contracted `Jd @ qd` paths across PlanarPCS, PCS, and GVS, including the GVS blockwise exponential/Jacobian/derivative kernel.
- Updates the affected system call sites, documentation, benchmark harness, dependency metadata, CI scheduling, and regression coverage.

## Numerical stability improvements relative to `main`

Several expressions on `main` had finite primal values but undefined, non-finite, or inaccurate derivatives at removable singularities and small rotations. This PR addresses those cases as follows:

- **Singular scalar/vector operations:** `safe_sqrt`, `safe_norm`, `safe_divide`, and `safe_normalize` sanitize singular inputs before evaluation and provide finite custom-JVP behavior. Analytic limiting values are used where available, with zero derivatives at the singular point and configurable fallbacks for division/normalization. `strict_singularities_mode()` restores the raw operations so problematic configurations remain diagnosable.
- **Small-angle Lie coefficients:** forward and inverse left-Jacobian coefficients use dtype-aware high-order Taylor series near zero instead of cancellation-prone quotients. For the reusable scalar coefficients, the forward and inverse cutoffs are derived from machine precision (scaling as `u^(1/12)` and `u^(1/8)`, respectively); the full SE(2)/SE(3) assemblies apply their own dtype-aware derivative thresholds. Closed-form inputs are sanitized before branch selection so vectorized `where` evaluation cannot leak `NaN` values from an unused singular branch.
- **Stable `SO(3)`/`SE(2)`/`SE(3)` derivatives:** exponential/logarithmic maps, arbitrary-axis branches, left Jacobians, and constant-strain tangents now share stable coefficients and direct analytic derivatives. The SE(2) path uses direct rotation/translation blocks; the SE(3) path uses exact reduced matrix polynomials and shared powers. This keeps values, Jacobians, and Hessians finite and accurate at zero, near-zero, and cutoff-boundary rotations without adding autodiff to production code.
- **System-level propagation:** PlanarPCS, PCS, and GVS reuse the same stable operators and intermediates. Dynamics-only recurrences propagate the contracted `Jd @ qd` quantity instead of materializing unused full derivative matrices, while public kinematics still returns complete Jacobian derivatives. Degenerate normalization, division, inverse-kinematics, and routing paths now use the same finite conventions.

The new regressions cover near-identity SE(2) reverse-mode Jacobians, arbitrary-axis SE(3) round trips and Hessians at the series boundary, constant-strain operators at zero rotation, strict-mode behavior, batching, and the fused system paths.

- **Nested batched autodiff at branch points:** `lax.cond` may lower to selection under `vmap`, so inactive square-root and quotient inputs are sanitized before evaluation. This keeps nested JVP/reverse-mode transforms finite and restores the SoftRobot forward-kinematics and spatial-Jacobian regressions at the straight configuration.

## Public API and implementation organization

- `soromox.utils.numerics` provides the safe numerical primitives.
- `soromox.utils.diagnostics` reports non-finite leaves and their pytree paths.
- `lie_algebra.jacobian_coefficients` exposes reusable stable scalar coefficients for forward and inverse Jacobians.
- `lie_algebra.se2` and `lie_algebra.se3` expose general left Jacobians and analytic directional derivatives for accumulated Lie-algebra coordinates.
- `constant_strain.se2` and `constant_strain.se3` expose rod-specific adjoint, inverse-adjoint, tangent, tangent-derivative, and bundled-operator APIs.
- Private matrix-polynomial helpers remain implementation details; arclength scaling, prepared powers, and rod-segment bundles remain in `constant_strain`.

## Performance against `main`

Measurements use the same adapted benchmark harness in both worktrees so the `main` baseline can compile `rollout_to` with Python solver/save parameters treated as static. Settings: CPU, macOS 26.5.2 arm64, JAX 0.11.0, 64-bit mode, 1-second duration, solver dt 1e-4 s, save dt 0.01 s, segment counts 1/2/4, three steady-state executions per fresh process, and five fresh processes per revision. Baseline commit: `bd6cd940`. PlanarPCS and PCS retain the measurements at `aa886bbb`; the GVS rows were refreshed at the current implementation commit `e746090`. Times are medians across the five processes; negative changes are faster. Rollout compile time is estimated as first-call latency minus steady execution and is consequently noisier than steady execution for long rollouts.

The targeted branch hardening itself adds no material overhead on CPU/JAX 0.11.0: across 65,536-sample batched SO(3), SE(3), and quaternion kernels, steady runtimes changed by approximately -4% to +5%. The PCS nested gradient path changed compilation by +1.1% and warm runtime by -0.1%, while removing the non-finite gradient.

### One-second rollout timings

| System | Segments | main compile (ms) | after compile (ms) | Δ compile | main steady exec (ms) | after steady exec (ms) | Δ exec | after/main |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| PlanarPCS | 1 | 418.823 | 388.747 | -7.2% | 284.686 | 220.726 | -22.5% | 0.775x |
| PlanarPCS | 2 | 519.097 | 424.780 | -18.2% | 1200.711 | 861.060 | -28.3% | 0.717x |
| PlanarPCS | 4 | 523.262 | 421.897 | -19.4% | 2337.649 | 1504.520 | -35.6% | 0.644x |
| PCS | 1 | 458.873 | 370.662 | -19.2% | 1339.003 | 1087.558 | -18.8% | 0.812x |
| PCS | 2 | 599.242 | 566.476 | -5.5% | 2853.023 | 2112.749 | -25.9% | 0.741x |
| PCS | 4 | 509.279 | 453.553 | -10.9% | 7243.398 | 4705.703 | -35.0% | 0.650x |
| GVS | 1 | 538.683 | 545.711 | +1.3% | 3289.324 | 3230.340 | -1.8% | 0.982x |
| GVS | 2 | 545.043 | 501.973 | -7.9% | 8289.066 | 4892.176 | -41.0% | 0.590x |
| GVS | 4 | 359.454 | 423.742 | +17.9% | 17525.821 | 11932.641 | -31.9% | 0.681x |

### Complete non-rollout method-suite totals

These totals sum the matched compile and steady-state timings for all non-rollout methods at segment counts 1/2/4 within each system class.

| System | main compile (ms) | after compile (ms) | Δ compile | main steady exec (ms) | after steady exec (ms) | Δ exec |
|---|---:|---:|---:|---:|---:|---:|
| PlanarPCS | 2598.080 | 2411.344 | -7.2% | 0.734 | 0.672 | -8.4% |
| PCS | 3097.498 | 2995.833 | -3.3% | 1.366 | 1.165 | -14.7% |
| GVS | 9167.956 | 9829.874 | +7.2% | 4.974 | 4.939 | -0.7% |

## Validation

- Nested `grad(vmap(jvp(...)))` regressions cover SO(3), SE(3), and quaternion conversions; the former strict xfails in PCS and SoftRobot defaults now pass as ordinary tests.
- Focused branch validation passes: 203 Lie/geometry tests, 15 spatial-default tests, and the full PCS suite (111 tests).
- `tests/lie_algebra` plus the generic matrix-polynomial tests: 169 passed.
- All 81 GVS system tests pass with the fused kernel.
- Broader numerical/system regression selection: 663 passed and 1 expected failure.
- Changed-code Ruff lint and formatting checks pass.
- Zensical documentation build passes.
- Python compilation passes.


docs/api/utilities/index.md
```diff
@@ -26,16 +27,34 @@ Pose encodings, rotation representation conversions, and geometric errors.
...
+- Constant-strain adjoint, inverse-adjoint, tangent, and tangent derivative
+- Bundled evaluation when several related operators are needed
+- Separate planar and spatial operator namespaces
 
 ## Quick Links
 
 - [Parameters API Reference](parameters.md) - Parameter management utilities
 - [Geometry API Reference](geometry.md) - Pose, rotation, and error utilities
-- [Lie Algebra API Reference](lie-algebra.md) - Lie group and constant-strain utilities
+- [Numerics API Reference](numerics.md) - Safe numerical primitives
```
mstoelzle, line 57: Please reflect on the best order here.

docs/api/overview.md
```diff
@@ -68,8 +68,10 @@ The [Actuation](actuation/index.md) module provides utilities for modeling and w
 The [Utilities](utilities/index.md) module includes:
 
 - **[Parameters](utilities/parameters.md)**: Parameter handling, validation, and default configurations for soft robot systems
```
mstoelzle, line 69: Please reflect on the best order for this list

src/soromox/utils/lie_algebra/se3.py
```diff
@@ -6,14 +6,197 @@
...
+
+_REDUCED_POLYNOMIAL_DEGREE = 4
+
+
+def _left_jacobian_series_threshold(dtype: jnp.dtype) -> Array:
+    """Return the dtype-aware dimensionless SE(3) Jacobian cutoff."""
+    return jnp.asarray(jnp.finfo(dtype).eps ** (1.0 / 14.0), dtype=dtype)
+
+
+def _left_jacobian_series_arguments(
```
mstoelzle, line 45: Please increase the docstrings and make them more comprehensive for all left jacobian functions (not just this one, not just in se3, but also in se2)

docs/api/utilities/index.md
mstoelzle: Thanks — I changed the utilities order to Parameters → Numerics → Geometry → Lie Algebra → Constant Strain, matching the top-level API overview and placing finite-value conventions before geometry and higher-level Lie/rod operators.

docs/api/overview.md
mstoelzle: I reviewed the overview ordering and kept Parameters → Numerics → Geometry → Lie Algebra → Constant Strain. The utilities index now uses the same sequence, so the two navigation views are consistent.

src/soromox/utils/lie_algebra/se3.py
mstoelzle: Addressed in 53913c8: expanded all six public left-Jacobian docstrings in se2.py and se3.py with the angular-first conventions, input shapes, threshold semantics, directional-derivative meaning, and shared-work bundle behavior.

src/soromox/control/reference_trajectory.py
```diff
@@ -43,10 +44,9 @@ def interp_fn(t: Array) -> Array:
         t0 = ts[idx - 1]
         t1 = ts[idx]
 
-        # Compute interpolation weight
-        # Handle case where t0 == t1 (avoid division by zero)
+        # Compute interpolation weight, guarding against duplicated samples
         dt = t1 - t0
-        alpha = jnp.where(dt > 0, (t - t0) / dt, 0.0)
+        alpha = safe_divide(t - t0, dt, 0.0)
```
vdperfetta01: This change no longer accounts for the possibility of dt<0. If, by construction of t1 and t0, this scenario would never happen, please ignore the comment.

vdperfetta01 (commented):
I went through the code and documentation, without spotting any significant issue. The implementation looks coherent. Only one implementation aspect needs to be double-checked

src/soromox/utils/lie_algebra/constant_strain/se3.py
```diff
@@ -0,0 +1,475 @@
...
+
+
+def _scaled_powers_from_prepared(
+    prepared: _PreparedAdjointPowers, s: Array
+) -> tuple[list[Array], list[Array]]:
+    """Scale prepared ``A**k`` and directional powers by ``s**k``."""
+    scale = jnp.ones((), dtype=prepared.powers.dtype)
+    powers: list[Array] = []
+    dot_powers: list[Array] = []
+    for order in range(5):
```
mohammedtarnini: I see _REDUCED_POLYNOMIAL_DEGREE defined above and is equal to 4. Maybe the for loop should be "in range(_REDUCED_POLYNOMIAL_DEGREE + 1)" instead of "in range(5)".

mohammedtarnini (approved):
The codes look fine overall. The structure is clear, and the implementation appears consistent and well organized. No significant issues were identified in the code.

src/soromox/control/reference_trajectory.py
mstoelzle: Addressed in fdfa2cc. I retained the finite-safe division for duplicate timestamps and added an explicit `dt > 0` guard, so non-positive intervals now use `alpha = 0` instead of interpolating. I also added a regression test for negative intervals; the focused reference-trajectory and SE(3) constant-strain suites pass (54 tests).

src/soromox/utils/lie_algebra/constant_strain/se3.py
mstoelzle: Addressed in fdfa2cc. The prepared-power loop now uses `range(_REDUCED_POLYNOMIAL_DEGREE + 1)` instead of the hard-coded `range(5)`, keeping the loop derived from the shared polynomial-degree constant. The focused SE(3) constant-strain suite passes.

PR #151, opened 2026-08-17
Migrate PlanarHSA to numeric PCS implementation

mstoelzle:
## Summary

- Migrate `PlanarHSA` from generated symbolic expressions to a JAX-native numerical implementation.
- Make `PlanarHSA` inherit from `PlanarPCS` and reuse its planar Lie-group kinematics, Jacobians, quadrature, and analytic dynamics infrastructure.
- Place the implementation in `src/soromox/systems/pcs/planar_hsa.py`; keep `src/soromox/systems/pcs/pcs.py` dedicated to spatial `PCS`.
- Remove the symbolic derivation/expression systems and legacy `soromox.parameters` package, update parameter files to the `xi_ref` schema, and move the PlanarHSA simulation example under `examples/simulation/pcs/`.
- Add cached HSA geometry/material quantities, shared end-effector kinematics, a gravity-only assembly path, segment-tip reuse, and batched Coriolis derivatives.
- Remove the obsolete motor-to-end-effector Jacobian demo.

## Validation

- `tests/systems/test_planar_hsa.py tests/systems/test_system_lengths.py`: 26 passed after the module relocation.
- The broader HSA/PCS/2D-coherence/citation regression set passed: 162 tests.
- Strict MkDocs build passed after updating the API reference to `soromox.systems.pcs.planar_hsa.PlanarHSA`.
- Ruff, Python compilation, JIT/autodiff smoke tests, renderer compatibility, energy/force identities, and positive-semidefinite inertia checks passed.
- An untracked symbolic-parity harness compared six deterministic scenarios plus hysteresis cases against the detached `main` baseline. Kinematics and Jacobians matched to approximately 1e-14; dynamics matched within calibrated quadrature tolerances. Five-point Gauss--Legendre quadrature was retained after convergence checks.

## PlanarHSA benchmark

Measured in separate fresh processes on Apple Silicon, macOS 26.5.2, Python 3.12.11, JAX 0.11.0, with float64 enabled. Both implementations used the same one-segment/two-rod `fpu_control` model and the same representative configuration, velocity, motor input, and external force. “First call” includes JAX compilation plus execution; “warm median” is the median of 30 synchronized calls.

| Operation | Old symbolic first | New numeric first | Old symbolic warm | New numeric warm |
| Model construction | 1.021 s | 2.399 s | — | — |
| End-effector forward kinematics | 42.7 ms | 56.9 ms | 14.1 µs | 15.0 µs |
| End-effector Jacobian | 37.4 ms | 118.7 ms | 13.5 µs | 44.9 µs |
| Gravitational force | 117.5 ms | 192.9 ms | 13.6 µs | 51.4 µs |
| Forward dynamics | 3.394 s | 0.426 s | 54.6 µs | 77.5 µs |
| Full kinematics/dynamics workload* | 474.6 ms | 364.9 ms | 47.5 µs | 71.9 µs |

*The full workload computes end-effector kinematics and Jacobian, inertia, gravity, elastic force, and damping in one JIT-compiled function.

The numeric implementation substantially reduces first-call compilation time for forward dynamics (about 8.0x in this run) and for the full workload (about 1.3x), while its warmed dynamics calls remain slower because they evaluate the analytic quadrature/differential model at runtime instead of dispatching to pre-generated symbolic expressions.


src/soromox/systems/pcs/planar_hsa.py
```diff
@@ -0,0 +1,1762 @@
...
+        )
+        return g_tip @ platform_end @ poses.planar_pose_to_transform(self.end_effector_offset)
+
+    def _forward_end_effector_from_xi(self, xi: Array) -> Array:
+        _, g_tips = self._segment_bases_and_tips(xi)
+        return poses.planar_pose_from_transform(
+            self._end_effector_transform_from_tip(g_tips[-1]), eps=self.global_eps
+        )
+
+    def _kinematic_xi(self, q: Array) -> Array:
```
mstoelzle: It doesn't seem this one is actually necessary. It could be removed.

src/soromox/systems/pcs/planar_hsa.py
```diff
@@ -0,0 +1,1762 @@
...
+
+    def beta_inv(self, pxi: Array) -> Array:
+        pxi = jnp.asarray(pxi).reshape(self.num_segments, self.num_rods_per_segment, 3)
+        virtual = jnp.mean(pxi, axis=1)
+        virtual = virtual.at[:, 2].set(
+            virtual[:, 2] - jnp.mean(self.rod_offset * pxi[:, :, 0], axis=1)
+        )
+        return virtual.reshape(self.num_strains)
+
+    def ref_strains(self) -> Array:
```
mstoelzle: Is this one necessary? Do we also have it in `PlanarPCS` or `PCS`? If not, I would suggest to remove it.

src/soromox/systems/pcs/planar_hsa.py
```diff
@@ -0,0 +1,1762 @@
...
+    def beta(self, vxi: Array) -> Array:
+        vxi = jnp.asarray(vxi).reshape(self.num_segments, 3)
+        physical = jnp.broadcast_to(
+            vxi[:, None, :], (self.num_segments, self.num_rods_per_segment, 3)
+        )
+        return physical.at[:, :, 2].set(
+            physical[:, :, 2] + self.rod_offset * physical[:, :, 0]
+        )
+
+    def beta_inv(self, pxi: Array) -> Array:
```
mstoelzle: Please add extensive doctrings everywhere - in particular to public methods.

src/soromox/systems/pcs/planar_hsa.py
```diff
@@ -0,0 +1,1762 @@
...
+
+    def _build_rod_strain_mapping(self) -> Array:
+        """Build the per-rod map from virtual to physical strains."""
+        mapping = jnp.broadcast_to(
+            jnp.eye(3, dtype=self.L.dtype),
+            (self.num_segments, self.num_rods_per_segment, 3, 3),
+        )
+        return mapping.at[..., 2, 0].set(self.rod_offset)
+
+    def _rod_strain_mapping(self) -> Array:
```
mstoelzle: This method doesn't seem necessary...

src/soromox/systems/pcs/planar_hsa.py
```diff
@@ -0,0 +1,1762 @@
...
+            axis=-2,
+        )
+        blocks = jnp.einsum("...ji,...jk,...kl->...il", mapping, Shat_rod, mapping)
+        return blk_diag(jnp.sum(blocks, axis=1))
+
+    def _nominal_stiffness_full(self) -> Array:
+        """Return the cached full HSA stiffness matrix."""
+        return self.K_full
+
+    def Shat(self) -> Array:
```
mstoelzle: This method doesn't seem necessary... Remember, we don't need to preserve backwards compatibility with respect to the previous implementation.

src/soromox/systems/pcs/planar_hsa.py
```diff
@@ -0,0 +1,1762 @@
...
+                jnp.stack(
+                    [zeros, zeros, self.nominal_axial_stiffness], axis=-1
+                ),
+            ],
+            axis=-2,
+        )
+        blocks = jnp.einsum("...ji,...jk,...kl->...il", mapping, Shat_rod, mapping)
+        return blk_diag(jnp.sum(blocks, axis=1))
+
+    def _nominal_stiffness_full(self) -> Array:
```
mstoelzle: This method doesn't seem necessary... Remember, we don't need to preserve backwards compatibility with respect to the previous implementation.

src/soromox/systems/pcs/planar_hsa.py
mstoelzle: Removed this wrapper. PlanarHSA now calls self.strain(q) directly and inherits the common strain implementation from PlanarPCS.

src/soromox/systems/pcs/planar_hsa.py
mstoelzle: Removed ref_strains. The model now uses the inherited PlanarPCS xi_ref attribute directly.

src/soromox/systems/pcs/planar_hsa.py
mstoelzle: Added comprehensive docstrings for the constructor, public kinematics/Jacobian/IK/dynamics methods, and the remaining protected numerical hooks.

src/soromox/systems/pcs/planar_hsa.py
mstoelzle: Removed the _rod_strain_mapping accessor. Actuation now uses the cached rod_strain_mapping field directly.

src/soromox/systems/pcs/planar_hsa.py
mstoelzle: Removed the legacy Shat method. HSA force assembly now uses the cached K_full matrix directly.

src/soromox/systems/pcs/planar_hsa.py
mstoelzle: Removed the unused nominal-stiffness wrapper. The cached K_full matrix is built and refreshed directly.

PR #156, opened 2026-08-19
Support compiled parameter updates

mstoelzle:
## Motivation

PR #152 exposed a real limitation in the parameter-update API: rebuilding a GVS model from updated parameters inside a JAX transformation could execute Python value validation and static-layout reconstruction on traced arrays, leading to a `ConcretizationTypeError`. Although the motivating example updated a base orientation, the underlying problem was broader than that field and broader than GVS.

Parameter identification, optimization, batched evaluation, and differentiable simulation all benefit from treating a model's numeric parameters as a dynamic PyTree argument while retaining one already-constructed model as the source of static structure. Before this PR, that pattern was not consistently supported across the system, actuator, passive-element, and caller-owned material implementations.

This PR establishes and tests a general fixed-layout parameter-update contract rather than adding a special case for the original GVS example.

## Update contract

A constructed system owns the static model structure. Its complete parameter PyTree may be passed as a dynamic argument to `with_params(...)` inside `eqx.filter_jit`:

```python
@eqx.filter_jit
def evaluate(params, q):
    candidate = robot.with_params(params)
    return candidate.potential_energy(q)
```

Calls reuse the same compiled executable when the parameter PyTree retains:

- the same tree structure;
- the same leaf shapes and dtypes;
- the same optional `None`/array layout; and
- the same model and component topology.

Numeric model, environment, actuator, passive-element, and isotropic-material values remain differentiable and can participate in `grad`, `value_and_grad`, and `vmap`.

Changes to static structure still require reconstruction and normally recompilation. Examples include segment or channel counts, active-strain and GVS basis layouts, quadrature or padding, joint/component types, chamber counts, routing spans, McKibben joint-pair topology, and selection between optional material fields.

### Two-stage validation

Parameter validation now distinguishes between two contexts:

- **Eager updates** perform complete structural and value validation, preserving the existing error reporting for invalid physical parameters.
- **Traced updates** perform validation based on statically observable properties such as types, PyTree structure, and array shapes. They do not attempt to convert runtime tracer values to Python booleans or integers.

`BaseSystemParams` owns the composed `validate()` and dispatched `validate_for_update()` methods. Every parameter subclass implements tracer-safe `validate_structure()` and eager-only `validate_values()` hooks; structure-aware families use matching compatibility hooks. This makes the eager-versus-traced boundary explicit by construction.

Compiled callers remain responsible for keeping runtime candidates inside their physical domains. Constrained parameterizations or eager validation of externally supplied inputs are recommended before entering an optimization or simulation loop.

### Functional cache refresh

Each concrete `with_params(...)` implementation continues to rebuild its parameter-dependent runtime representation immutably. The original system is unchanged, while the returned system receives refreshed geometry, mass, stiffness, damping, reference, or other cached arrays as appropriate for that model.

The `SoftRobot` documentation now makes this contract explicit and clarifies that `precompute()` is a constructor-time, potentially mutating hook rather than the mechanism for immutable compiled updates. Component parameter delegates return `Self`, preserving the concrete robot type; their parameter arguments use the component-level `BaseParams` type.

### Static component topology

Some components contain numeric parameters together with topology that must remain fixed for a compiled model:

- I-SUPPORT chamber updates reuse the actuator's precomputed routing spans instead of reconstructing routing topology from traced values.
- McKibben updates retain the installed joint-pair indices while allowing the remaining numeric transmission and actuator values to change.
- Threadlike routing and articulated tendon validation retain their eager topology checks while allowing compatible numeric leaves to be traced.

Topology-changing eager updates are still rejected and require component/robot reconstruction.

### Stable GVS runtime layout

GVS cache refreshes preserve the integer dtype of cross-section geometry indices. This prevents a same-layout parameter update from changing a PyTree leaf dtype and therefore triggering an avoidable recompilation.

### Isotropic material layouts from #139

The caller-owned `IsotropicMaterialParams` uses the same explicit validation split. Construction and replacement call `validate_for_update()`; `validate_structure()` checks the active optional-field selection and array layout, while `validate_values()` provides the finite-value and physical-domain checks used by eager validation.

The shear-modulus versus Poisson-ratio choice and damped versus undamped choice deliberately produce four different PyTree structures. Each structure compiles separately, while repeated numeric updates within that structure reuse the compiled executable. This preserves the new material API from #139 without exception-based tracer detection inside the material validator.

## Test strategy

Regression coverage was added before and alongside the implementation changes. The tests exercise:

- immutable eager updates and refreshed runtime arrays;
- same-layout compiled updates with a trace counter proving no retrace;
- differentiation through updated parameter PyTrees;
- rejection of shape or topology changes that require reconstruction;
- system body parameters;
- actuator parameters;
- passive-element parameters;
- the previously problematic GVS and I-SUPPORT/McKibben paths; and
- shear/Poisson and damped/undamped isotropic-material layouts on PCS, PlanarPCS, and GVS.

System-specific coverage lives in each concrete system's test module. Component-level coverage lives in the joint, threadlike-actuation, and shared continuum-component tests.

## Implications and tradeoffs

- Numeric parameter sweeps and optimization loops can update compatible parameters within compiled code without reconstructing the static model or compiling once per numeric value.
- The model returned by `with_params(...)` remains immutable and has consistent refreshed caches.
- Value-dependent Python exceptions cannot be produced from arbitrary runtime tracers. This is an inherent JAX tracing boundary, not a relaxation of eager validation.
- Structural updates remain deliberately outside the compiled update contract.
- Retaining precomputed topology means traced topology leaves are not a mechanism for changing routing or connectivity; reconstruct the component instead.
- Optional material selections remain structural, so changing between shear and Poisson input or enabling/disabling damping requires a different compilation.
- The implementation gives every parameter class explicit structural and value validation hooks and adds broader tests, but avoids moving model-specific cache logic into `SoftRobot`, where it could not be implemented correctly for all systems.

## Documentation

The user guide documents the compiled update pattern, its eager-versus-traced validation boundary, and the distinction between dynamic parameters and static structure. The parameter, continuum-component, and actuation API pages link the relevant concepts, and the changelog records the new behavior.

## Stack and related work

This PR originally built on #139, whose shear/Poisson and optional-damping material layouts are now part of `main`. The branch is current with `main`.

It supersedes the focused GVS fix in #152 by addressing the same failure mode through a general parameter-update contract across the library.

## Validation

Validation on the current head:

- Shared component, actuation, and typed-parameter suites: **133 passed**
- Articulated, pendulum, HSA, I-SUPPORT, and default-system suites: **142 passed**
- Full PCS, PlanarPCS, and GVS suites: **323 passed**
- Cross-system compiled/eager parameter-update selection: **29 passed** (**421 deselected**)
- Ruff lint and touched-file formatting checks: **passed**
- Zensical documentation build: **passed**

### DIFF
--- a/src/soromox/systems/params.py
+++ b/src/soromox/systems/params.py
@@ -12,11 +12,16 @@ from dataclasses import fields
 from typing import Any, ClassVar, Literal
 
 import equinox as eqx
-from jax import Array
+from jax import Array, core, tree_util
 from jax import numpy as jnp
 from jax.errors import ConcretizationTypeError, TracerBoolConversionError
 
 DEFAULT_GRAVITY_MAGNITUDE = 9.81
+
+
+def _contains_tracer(tree: Any) -> bool:
+    """Return whether a parameter PyTree is currently under a JAX transform."""
+    return any(isinstance(leaf, core.Tracer) for leaf in tree_util.tree_leaves(tree))
 
 
 def _validate_finite_array(
@@ -132,7 +137,7 @@ class BaseSystemParams(eqx.Module):
                 value,
                 is_leaf=lambda leaf: leaf is None,
             )
-        updated.validate()
+        updated.validate_for_update()
         return updated
 
     def _normalize_replacement(self, name: str, value: Any) -> Any:
@@ -147,6 +152,17 @@ class BaseSystemParams(eqx.Module):
             shapes, values, or component relationships.
         """
         return None
+
+    def validate_structure(self) -> None:
+        """Validate the shape/type contract observable during JAX tracing."""
+        self.validate()
+
+    def validate_for_update(self) -> None:
+        """Validate fully in eager code and structurally under JAX transforms."""
+        if _contains_tracer(self):
+            self.validate_structure()
+        else:
+            self.validate()
 
     def validate_against_structure(self, structure: Any) -> None:
         """Validate parameters against static construction choices.
--- a/src/soromox/systems/params.py
+++ b/src/soromox/systems/params.py
@@ -12,7 +12,7 @@ from dataclasses import fields
 from typing import Any, ClassVar, Literal, final
 
 import equinox as eqx
-from jax import Array, core, tree_util
+from jax import Array, core, ensure_compile_time_eval, tree_util
 from jax import numpy as jnp
 
 DEFAULT_GRAVITY_MAGNITUDE = 9.81
@@ -160,11 +160,12 @@ class BaseSystemParams(eqx.Module):
 
     @final
     def validate_for_update(self) -> None:
-        """Validate fully in eager code and structurally under JAX transforms."""
+        """Validate values when concrete and structure when dynamically traced."""
+        self.validate_structure()
         if _contains_tracer(self):
-            self.validate_structure()
-        else:
-            self.validate()
+            return
+        with ensure_compile_time_eval():
+            self.validate_values()
 
     def validate_structure_compatibility(self, structure: Any) -> None:
         """Validate statically observable compatibility with a model structure."""
@@ -193,12 +194,16 @@ class BaseSystemParams(eqx.Module):
 
     @final
     def validate_for_update_against_structure(self, structure: Any) -> None:
-        """Validate structure compatibility at the active JAX trace level."""
+        """Validate concrete values and traced structure against a model."""
+        self.validate_structure()
         if _contains_tracer(self):
-            self.validate_structure()
             self.validate_structure_compatibility(structure)
-        else:
-            self.validate_against_structure(structure)
+            return
+        with ensure_compile_time_eval():
+            self.validate_values()
+        self.validate_structure_compatibility(structure)
+        with ensure_compile_time_eval():
+            self.validate_value_compatibility(structure)
 
 
 class BaseSoftRobotParams(BaseSystemParams):
--- a/src/soromox/systems/params.py
+++ b/src/soromox/systems/params.py
@@ -9,12 +9,11 @@ __all__ = [
 ]
 
 from dataclasses import fields
-from typing import Any, ClassVar, Literal
+from typing import Any, ClassVar, Literal, final
 
 import equinox as eqx
 from jax import Array, core, tree_util
 from jax import numpy as jnp
-from jax.errors import ConcretizationTypeError, TracerBoolConversionError
 
 DEFAULT_GRAVITY_MAGNITUDE = 9.81
 
@@ -27,15 +26,11 @@ def _contains_tracer(tree: Any) -> bool:
 def _validate_finite_array(
     name: str, value: Array, expected_shape: tuple[int, ...]
 ) -> Array:
-    """Validate array shape and concrete finite values when available."""
+    """Validate an eager array's shape and finite values."""
     array = jnp.asarray(value)
     if array.shape != expected_shape:
         raise ValueError(f"{name} must have shape {expected_shape}, got {array.shape}.")
-    try:
-        all_finite = bool(jnp.isfinite(array).all())
-    except (ConcretizationTypeError, TracerBoolConversionError):
-        return array
-    if not all_finite:
+    if not bool(jnp.isfinite(array).all()):
         raise ValueError(f"{name} must contain only finite values.")
     return array
 
@@ -54,9 +49,7 @@ def validate_planar_base_pose(name: str, value: Array) -> None:
         None.
 
     Raises:
-        ValueError: If the shape is not ``(3,)`` or any concrete entry is
-            non-finite. During JAX tracing, only the static shape check is
-            performed.
+        ValueError: If the shape is not ``(3,)`` or any entry is non-finite.
     """
     _validate_finite_array(name, value, (3,))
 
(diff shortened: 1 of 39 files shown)


mstoelzle:
Hi @matteodv99tn,

Would you be willing to try the implementation in this PR against the GVS/JIT workflow that motivated #152, and check whether it resolves the concretization issue you encountered?

In particular, it would be helpful to test your representative parameter-update and potential-force/equilibrium evaluation path with the parameter PyTree passed through `eqx.filter_jit`. The intended contract is that numeric updates with unchanged PyTree structure, shapes, dtypes, and static model topology reuse the compiled executable.

If the behavior now addresses your use case, could you please leave a review afterwards? If it does not, please share the failing example or traceback here so we can incorporate it before merging.

Thank you again for raising the original issue and for helping validate the broader solution.

src/soromox/systems/params.py
```diff
@@ -148,6 +153,17 @@ def validate(self) -> None:
         """
         return None
 
+    def validate_structure(self) -> None:
```
Michele-Martini: What is the difference between validate() and validate_structure? It appears the latter calls the former with no other change.

matteodv99tn:
@Michele-Martini I agree with your comment; I also found that default implementation confusing at first.

My understanding of the intended design is:

* `validate_structure()` must be safe during JAX tracing and should only check statically observable properties, such as types, shapes, PyTree layout, and fixed topology.
* `validate()` performs complete eager validation, including value-dependent conditions such as finiteness, positivity, symmetry, and admissible ranges.
* `validate_for_update()` detects whether the parameter PyTree contains tracers and dispatches to the appropriate validation level.

Having `validate_structure()` delegate to `validate()` by default initially seems to undermine this distinction. I believe the reason is backward compatibility: several existing `BaseSystemParams` subclasses already have tracer-safe `validate()` implementations. 
Classes whose value checks are not inherently tracer-safe instead override both methods, like in `ISupportParams`.

Therefore, the current implementation appears correct for the existing classes, but it is safe by convention rather than by construction. A future subclass could implement a value-dependent `validate()` and forget to override `validate_structure()`, causing traced updates to fail. There is also a potential recursion trap if such a `validate()` calls `self.validate_structure()` without overriding it.

A clearer long-term design could separate structural and value validation explicitly:

```python
class BaseSystemParams:
    def validate_structure(self) -> None:
        pass

    def validate_values(self) -> None:
        pass

    def validate(self) -> None:
        self.validate_structure()
        self.validate_values()

    def validate_for_update(self) -> None:
        if _contains_tracer(self):
            self.validate_structure()
        else:
            self.validate()
```

This makes the contract explicit, although migrating all existing validators may be beyond the scope of this PR.

If that refactoring is not intended here, I would at least clarify the fallback in the docstring, and add a fail-safe implementation that catches tracer-conversion errors and replace them with a more actionable message:

```python
def validate_structure(self) -> None:
    """Validate statically observable parameter structure.

    The default delegates to `validate()` for backward compatibility with
    validators that are already trace-safe. Subclasses containing
    value-dependent Python checks must override this method.
    """
    try:
        self.validate()
    except (ConcretizationTypeError, TracerBoolConversionError) as exc:
        raise RuntimeError(
            f"{type(self).__name__}.validate() is not trace-safe; "
            "override validate_structure()."
        ) from exc
```


I will try to test feature in my own library ASAP, but from reviewing the code (especially looking at the unit tests), I think that everything shall work just fine.

mstoelzle:
Hi @matteodv99tn, thank you for the detailed analysis and suggestion. I agree that making the distinction safe by construction is clearer.

I implemented the proposed refactoring without retaining the backward-compatible fallback:

- `BaseSystemParams.validate()` now composes `validate_structure()` and `validate_values()`.
- `validate_for_update()` runs only structural validation during tracing and full validation otherwise.
- All existing parameter classes and usages have been migrated to the two explicit hooks.
- Structure-dependent GVS checks were similarly separated into structural and value compatibility hooks.
- Regression tests verify that traced updates skip value-dependent checks while still rejecting invalid shapes and structure.

I also updated the documentation to describe this contract explicitly. Thanks again for identifying the convention-based weakness. If you have an opportunity to test the updated branch in your library, your feedback would still be very helpful.

matteodv99tn:
Hi @mstoelzle,

Sorry, but I took my time to test the the library separately with my use-case as baseline.
This exposed some blind-spots in current testing strategy, as the following pattern still fails:

```python
@eqx.filter_jit
def evaluate(base_pose, q):
    return system.update_params(base_pose=base_pose).potential_force(q)
```

The current tests pass a complete parameter PyTree through `with_params`, so all nested parameter arrays are traced.
In the case above, only `base_pose` is traced and unchanged nested arrays remain closed over. Their value validation can then execute inside the enclosing JAX trace and raise a tracer conversion error.

Since I don't have write access to the repo, I will open a separate PR with focused regression tests (and proposed fixes) for this GVS path and the same partial-update pattern in threadlike, McKibben, and articulated tendon components.

mstoelzle:
Hi @matteodv99tn,

Thanks for testing this against your use case and for adding the focused regressions in #165. You were right: the existing tests missed partial updates where only a top-level field is traced while unchanged nested arrays remain closed over.

I’ve incorporated the regression coverage into #156 in [`a71894a`](https://github.com/tud-phi/soromox/commit/a71894a74f653983aa5662650828b3bc05532670) and added the revised implementation in [`5bfd97a`](https://github.com/tud-phi/soromox/commit/5bfd97a91b5970ead6484b16ee41442ee836a6a8).

Instead of relying on exception suppression or skipping all component-to-robot validation whenever any parameter is traced, the new implementation:

- always runs tracer-safe structural validation;
- validates closed concrete values at JAX trace time using `ensure_compile_time_eval`;
- defers only checks that depend on dynamic tracer leaves; and
- keeps component-to-robot structural and numeric checks separate, so traced bounds do not suppress closed routing checks.

I also consolidated construction and update validation behind a single tracer-aware `validate_for_robot()` dispatcher.

The focused local validation passes: 77 component tests, seven GVS compiled-update tests, Ruff, Python compilation, and the documentation build. I’ve therefore closed #165 in favor of #156.

Thanks again for identifying the missing partial-update case and providing the reproducer.

matteodv99tn:
I independently tested the script on my machine, and everything works just fine now.

Thank you so much for your effort in writing and maintaining the library!

matteodv99tn (approved):

PR #158, opened 2026-08-19
Fix Issue #154: Sec V.d Multi-start

### LINKED ISSUE #154: Recover and document Section Vd six-batch optimization initialization
## Context

[PR #68](https://github.com/tud-phi/soromox/pull/68) added six-batch MATLAB result archives for the Section Vd control-gain comparison, but did not include the process that produced those batches. The earlier MATLAB script in that history only loads and plots existing results; it does not define or execute the batch optimization.

## Information needed

Please recover and document:

- the six initial `Kp`, `Ki`, and `Kd` gain sets for each controller;
- all random seeds and the randomization or sampling procedure;
- the exact execution method used to produce each run (scripts, commands, environment, and whether runs were independent or vectorized);
- how per-run histories and best trajectories were selected and aggregated;
- the intended batch archive schema, including array axes and the relationship between every recorded loss and gain state.

This should be coordinated with [#128](https://github.com/tud-phi/soromox/issues/128) (best-batch selection) and [#129](https://github.com/tud-phi/soromox/issues/129) (loss/gain history alignment). Once the provenance is documented, a separate change can restore reproducible six-batch optimization.

Michele-Martini:
Restores the six-start ("multishoot") gain optimization behind the published Section Vd
results, and documents what could and could not be recovered about the original run.

Closes #154. Builds on #149 and targets its branch; **merge #149 first**.

## Why

Section Vd is currently unreproducible. The committed archives and the plotter are
six-batch; the generators are single-start. Neither can produce the other's data, so the
paper's headline numbers cannot be regenerated.

## What #154 asked for

Recovered from the legacy MATLAB archives:

- **Execution** — vectorized multi-start, not independent runs and not multiple shooting:
  `vmap` over both `opt_vars` and `opt_state`, one optimizer per start, gradients never
  averaged. The rollout is a single forward integration with no segment or defect variables.
- **Selection** — `best_batch = argmin_b min_i history_loss[i, b]`, which evaluates to `3`
  for *both* methods in the committed data. That coincidence is what hid #128.
- **Aggregation** — `docs/research.md` reports 62% / 57% "relative to their initial median
  values". `1 - min(loss[-1]) / median(loss[0])` gives **62.62%** and **57.82%**, an exact
  match, which pins the metric. The README carries a runnable snippet.
- **Schema** — array axes documented in the README.

Not recoverable, and stated as such:

- **The six initial gain sets and their seed.** The archives hold no gain history at all.
  The nearest surviving script has `batch_size = 1`, `num_iters = 3`, a `0.5 s` horizon
  rather than `5 s`, and a `PRNGKey(35)` that is split but never consumed — its multi-start
  branch never executed. Its setpoint `[-0.055, -0.094, 0.118]` does match the synergistic
  archive verbatim, so the results came from an **uncommitted variant** of that script.
- Structurally they never could have been: in that script the loss at index `j` is measured
  *before* the update whose parameters land at index `j`, so index `0` is already one step
  past the initialization (this is #129). The new schema therefore stores `init_*` directly.

The provenance section is marked **provisional** — a machine that may hold the original
generator becomes available in the week of 2026-08-24. A `legacy_uniform` scheme reproduces
the stale script's dormant sampling box so a recovered generator can be checked against it,
and `init_scheme` in every archive makes runs self-identifying.

## Changes

**Initialization** — new `code/secvd_init.py`. Start 0 is always the nominal set, so a
single-start run stays a strict subset of a batched one. Starts 1..B-1 scale the nominal by
one log-uniform factor per gain, so each start is describable by three numbers, the form
#154 asks for. New flags: `--batch-size` (default 6), `--init-seed`, `--init-scheme`,
`--init-spread`.

**Loop** — `gain_optimization_loop.py` vmaps the gradient function, `optimizer.init` and
`optimizer.update` together, inside the loop rather than in the callers, so the evaluation
cannot be batched while the optimizer is left unbatched. Two consequences: `multi_transform`
label trees keep matching the unbatched parameter tree, and `clip_by_global_norm` clips each
start independently. A start that goes non-finite freezes at its *current* parameters (not
its last-good ones, which would produce a finite loss again and resurrect it) and the rest
continue.

**Schema v2** — `secvd_results.py`. The batch axis appears only where data varies per start;
`t_ts` and the references are shared and lose their axis of one. Adds `batch_size`,
`init_Kp/Ki/Kd`, `init_seed/scheme/spread`, `history_finite_mask`, `best_batch`, and
per-start `best_iteration`. Selection indices are supplied by the loop and re-derived by the
validator, so the two cannot silently disagree.

**Removes `history_qd_ts` / `history_u_ts`.** The B=1, five-iteration placeholder is already
28.9 MB; at 100 iterations and six starts these two fields alone reach ~2 GB. The loop keeps
each start's iteration-zero rollout and a running per-start best instead, which is all the
archive and figure need.

**#128 fixed** — each panel and the renderer take the best start from *their own* archive.

## Two latent bugs found

Both would have broken the first real `B>1` run:

- **`JAX_PLATFORMS=gpu` is not a valid value.** jax 0.10 knows `cpu`, `tpu`, `cuda`; `gpu`
  raises. #149 wrote the user-facing name straight into the variable, which never fired
  because `B` was always 1, so `auto` always chose `cpu`. Auto-GPU now requests `cuda,cpu`
  and warns on fallback; explicit `--device gpu` requests `cuda` and fails loudly.
- **JAX ignores `JAX_PLATFORMS` once imported** (verified). Threading `--batch-size` made the
  CLI import `secvd_init`, which imported jax, which would have silently defeated #149's
  device selection. `secvd_init` now imports jax lazily, guarded by a subprocess test.

## Files

| file | change |
| --- | --- |
| `code/secvd_init.py` | **new** — seeded multi-start initialization |
| `code/secvd_results.py` | schema v2, selection helpers, validation |
| `code/gain_optimization_loop.py` | vmapped loop, per-start freeze, running best |
| `code/secvd_optimization.py` | new flags, JAX platform mapping, CPU fallback |
| `code/control_gain_optimization_with_{collocated,synergistic}.py` | wired to `B=6` |
| `code/plot_control_gain_optimization.py` | min/max bands, per-archive `best_batch` |
| `code/render_control_gain_optimization_animations.py` | renders `best_batch`; validates every start |
| `README.md` | provenance, schema, CLI, plotting sections |
| `tests/paper_results/secVd/test_init_sampling.py` | **new** — initialization tests |

14 files, +1864 / -612.

## Tests

**69 passed, 2 xfailed.** Ruff check and format clean. Every test is cheap — toy losses and
synthetic archives, no rollouts.

### `test_init_sampling.py` (new) — the initialization is reproducible

| test | guards |
| --- | --- |
| `test_start_zero_is_always_the_untouched_nominal` | a single-start run stays a strict subset of a batched one, for both schemes and both batch sizes |
| `test_default_batch_matches_the_legacy_archive_width` | the default `B` stays 6, the width of the legacy archives |
| `test_sampling_is_reproducible_and_seed_dependent` | a seed reproduces a run exactly, and a different seed actually moves the starts |
| `test_samples_stay_inside_their_scheme_box` | `log_uniform_v1` stays in `[nominal/spread, nominal*spread]`; `legacy_uniform` stays in the recovered box |
| `test_each_start_is_one_scalar_factor_per_gain` | each start remains describable by three numbers (the form #154 asks for), and an anisotropic nominal keeps its shape — blocks a "simplification" to per-component sampling |
| `test_gain_keys_are_consumed_in_a_fixed_order` | reordering the input dict cannot reshuffle which PRNG key each gain draws |
| `test_requests_that_cannot_produce_finite_gains_are_rejected` | bad batch size, unknown scheme, degenerate spread, missing gains, and a zero nominal under the multiplicative scheme |
| `test_absolute_box_scheme_accepts_a_zero_nominal` | that restriction is specific to the multiplicative scheme |
| `test_description_lists_every_start` | the printed initialization table covers all starts |

### `test_result_schema.py` — the archive cannot misrepresent the run

| test | guards |
| --- | --- |
| `test_batch_axis_appears_only_where_data_varies_per_start` | per-start fields carry the axis, shared fields do not, and the multi-gigabyte per-iteration rollout histories stay absent |
| `test_metadata_needed_to_reproduce_the_run_round_trips` | seed, scheme, spread, batch size and selection survive a save/load — the fields whose absence made the legacy results unrecoverable |
| `test_schema_v1_and_legacy_archives_are_rejected_clearly` | stale archives fail loudly instead of being misread |
| `test_inconsistent_archives_are_rejected` | six mutations — wrong `batch_size`, wrong `best_batch`, wrong or out-of-range `best_iteration`, non-finite results, non-finite unmasked loss. Selection is re-derived on load, never trusted |
| `test_a_frozen_start_keeps_the_archive_valid` | masked entries are allowed to be non-finite; that is what the mask is for |
| `test_start_that_never_produced_a_finite_iterate_is_rejected` | such a start has no trajectory worth archiving |
| `test_collocated_requires_its_configuration_reference` | `q_des_ts` is mandatory for the collocated method |
| `test_best_batch_is_derived_per_archive` | **#128** at the primitive level: two methods, two indices |
| `test_improvement_metric_matches_the_recovered_paper_formula` | the metric verified to reproduce the published 62% / 57% |

### `test_gain_optimization_loop.py` — the numerical core

| test | guards |
| --- | --- |
| `test_every_start_is_optimized_independently` | no averaging across starts; each descends on its own |
| `test_history_pairs_each_loss_with_the_gains_it_was_measured_at` | **#129** per start: re-evaluates the loss at every recorded iterate, and pins entry 0 to the untouched initialization — the exact off-by-one that lost the legacy initialization |
| `test_a_diverging_start_freezes_but_keeps_the_iterates_it_earned` | one failure neither aborts the run nor retroactively discards that start's earlier valid iterations |
| `test_a_start_invalid_from_its_first_evaluation_is_excluded` | the distinct case that has nothing to contribute |
| `test_only_two_rollout_snapshots_are_kept_per_start` | no per-iteration rollout accumulator; `init_aux` is iteration zero and `best_aux` is each start's own best |
| `test_single_start_still_works` | `B=1` is not a broken special case |
| `test_invalid_sizes_are_rejected` | zero iterations or zero starts |

### `test_optimization_device.py` — device selection

New alongside the four existing cases:

| test | guards |
| --- | --- |
| `test_jax_platform_names_are_the_ones_jax_accepts` | `gpu` is never written to `JAX_PLATFORMS`; only `cuda` names the CUDA backend |
| `test_only_an_automatic_gpu_choice_falls_back` | auto requests `cuda,cpu`; an explicit `--device gpu` requests `cuda` and does not silently relocate |
| `test_unknown_device_has_no_platform_mapping` | unknown names fail rather than passing through |
| `test_batch_size_from_the_command_line_drives_device_selection` | `--batch-size` is read early enough to affect the backend |
| `test_cli_and_init_modules_do_not_import_jax` | subprocess check that the CLI layer stays JAX-free, since JAX ignores `JAX_PLATFORMS` once imported |

### `test_plot_and_renderer.py` — figure and animation

| test | guards |
| --- | --- |
| `test_each_panel_uses_its_own_best_batch` | **#128**, plotter: archives peaking at different starts must not share one index |
| `test_renderer_animates_the_archives_own_best_start` | **#128**, renderer: never assume the best start is index 0 |
| `test_plot_loss_draws_a_band_only_when_there_are_several_starts` | `B=1` gets no fabricated spread |
| `test_frozen_starts_do_not_drag_the_loss_band` | masked entries are excluded, not plotted as gaps |

The 2 xfails are `strict=True` on `test_committed_placeholder_pose_matches_forward_kinematics`:
the committed placeholders are schema-v1, so they fail the moment the archives are
regenerated, forcing the marker's removal rather than letting it rot.

## Caveats

- **Archives are not regenerated in this PR.** The committed placeholders remain schema-v1
  and the comparison figure is stale. Regeneration waits on the 2026-08-24 machine check, in
  case the original seeds surface and change what should be run.
- **Cost.** A GPU probe of the collocated generator at **B=2** measured 407 s for
  compile-plus-first-iteration and **385 s steady per iteration**, so 100 iterations is
  roughly **11 h per method**. B=6 was not measured; the batch axis is expected to be nearly
  free because the 50 001 sequential solver steps dominate and the GPU parallelizes only
  across the batch — but that is an extrapolation. Whether CPU beats GPU here is also
  unmeasured and plausible, given how small the per-step matrices are.
- **Regenerated numbers will not reproduce 62% / 57%**, independently of the lost seeds:
  #149 unified both methods on `u_constant = [0.5, 0.2, 0.1]` and moved the collocated
  saturation from `gamma = 10` to `gamma = 100 1/m`. `docs/research.md` is deliberately left
  untouched until real results exist.
- Expect archives around 80-100 MB at `B=6` (LFS-tracked), up from the legacy 23 MB, since
  per-start trajectories scale with `B`.


mstoelzle:
Closing this PR as it conflicts with the merged #139. The next PR should be based off `origin/main` or even better on top of #156, as #156 will be merged first.

PR #163, opened 2026-08-23
Accelerate PlanarPCS, PCS, and GVS forward dynamics

mstoelzle:
## Motivation

`PlanarPCS`, `PCS`, and `GVS` previously assembled forward dynamics through several avoidable intermediate quantities. Depending on the system, the hot path materialized pointwise poses, full-width Jacobians, dense selector products, coadjoint matrices, or a complete Coriolis matrix even though forward dynamics only needs the contracted force `C(q, qd) @ qd`.

This PR therefore targets four goals:

- reduce warm forward-dynamics latency for single-environment CPU and multi-environment GPU workloads;
- avoid materializing dynamics intermediates that are not part of the public result;
- keep one maintainable dynamics-assembly path without CPU/GPU or segment-count heuristics; and
- bound GVS tracing work as the number of links grows.

### PlanarPCS and PCS

- Propagate gravity directly in each segment's local frame instead of carrying absolute quadrature poses through dynamics assembly.
- Assemble only the active causal Jacobian prefix needed by the current segment.
- Apply diagonal spatial mass matrices directly and use matrix-free coadjoint actions.
- Accumulate inertia, `C(q, qd) @ qd`, and gravity together without constructing the full Coriolis matrix.
- Retain a trace-time segment loop so each segment uses its exact causal-prefix width.
- Use joint-compatible `_segment_dof_ends` naming so future joints between links do not require another structural rename.

### GVS

- Gather generalized coordinates into the existing fixed-size joint/link blocks.
- Evaluate joint transforms and cell-local Magnus/tangent terms in parallel with `vmap`.
- Advance the causal state with one fixed-shape outer segment `lax.scan` and one nested cell scan.
- Keep the propagated Jacobian in active coordinates and insert local joint/link bases with masked scatters, avoiding both dense selectors and a padded Jacobian state.
- Carry the Jacobian, `Jdot @ qd`, spatial velocity, local gravity, inertia, convective force, and gravity force through the outer recurrence.
- Use zero quadrature weights to mask padded cells and advance one final zero-weight cell to obtain each segment tip.

The outer segment body is traced once, independently of each link's concrete joint DOFs, basis DOFs, and quadrature count. It does not introduce a platform or model-size dispatch.

### Shared inertia solve

- Move the internal inertia solve to `SoftRobot` so the system implementations share one numerical backend.
- Use Cholesky on CPU for the small symmetric-positive-definite systems and retain the generic dense solve on accelerators.
- Keep `_solve_inertia` private because it is an implementation detail rather than a stable public dynamics API.

This is the only CPU/accelerator dispatch introduced by the PR; it is not combined with segment-count or DOF heuristics.

## Considered alternatives

- **Fully fusing kinematic recurrence and quadrature accumulation:** the initial PCS prototype reduced intermediate arrays but was slower in both single-environment CPU and multi-environment GPU measurements, so it was reverted.
- **Using `lax.scan` or `lax.fori_loop` for PlanarPCS/PCS segments:** rolled, partially unrolled (1, 2, 4, and 8), and fully unrolled forms were benchmarked. Their fixed-width state lost the exact causal-prefix advantage of the trace-time loop, including in the CPU cases that motivated the experiment.
- **Specializing every GVS link at trace time:** using each link's exact joint/link DOF count and quadrature count delivered the highest warm runtime in some CPU cases, but made compilation grow approximately linearly with link count and duplicated the link body in the compiled program.
- **Propagating a block-padded GVS Jacobian:** a literal fixed-shape outer scan substantially reduced compilation, but padded future-link blocks caused material CPU regressions, especially for heterogeneous bases. The retained implementation instead carries a full active-coordinate Jacobian and inserts only the current local basis.
- **Adding CPU/GPU plus segment-count heuristics:** this could select locally optimal loop forms, but would make behavior hardware-dependent and increase maintenance. The retained dynamics assembly has one code path; only the shared dense-solve backend dispatches by platform.

## Correctness

The public API and equations of motion are unchanged: `dynamics_terms(q, qd)` still returns `(M, Cqd, G)`, and `forward_dynamics` consumes those terms together with actuation, elasticity, damping, and external forces.

Across the benchmark grids, the maximum relative L2 differences from `origin/main` are:

| System | Maximum relative L2 difference |
| --- | ---: |
| PlanarPCS | `5.20e-10` |
| PCS | `7.74e-10` |
| GVS | `2.44e-9` |

The GVS path additionally agrees with the public inertia, Coriolis, and gravity methods and remains finite under both forward- and reverse-mode automatic differentiation for heterogeneous joint/basis models.

## Performance against `origin/main`

All “PR” columns below measure the implementation at the current PR head; all “main” columns measure `origin/main`. No intermediate implementation from this branch is used as a baseline.

Measurements use JAX FP64 with inputs resident on the selected device. CPU measurements use one environment on an Intel Core Ultra 9 285K. GPU measurements use an RTX 5090 and report batched resident-device execution. Baseline and PR functions were compiled separately, warmed, synchronized with `block_until_ready`, and invoked in rotating order. Compile time is `lower(...).compile()` wall time; runtime is the median warm forward-dynamics latency. Negative compile deltas are faster, while runtime speedups greater than 1 are faster.

### PlanarPCS

CPU, one environment:

| Segments | main compile (s) | PR compile (s) | Δ compile | main runtime (ms) | PR runtime (ms) | Runtime speedup |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 0.313 | 0.220 | -29.8% | 0.0154 | 0.0118 | 1.31x |
| 4 | 0.317 | 0.321 | +1.2% | 0.0465 | 0.0373 | 1.25x |
| 8 | 0.325 | 0.393 | +21.0% | 0.0807 | 0.0609 | 1.32x |
| 12 | 0.320 | 0.528 | +64.9% | 0.1264 | 0.0953 | 1.33x |
| 16 | 0.339 | 0.599 | +76.6% | 0.2118 | 0.1414 | 1.50x |
| 32 | 0.333 | 0.903 | +171.5% | 1.5495 | 0.6798 | 2.28x |

GPU:

| Segments | Environments | main compile (s) | PR compile (s) | Δ compile | main runtime (ms) | PR runtime (ms) | Runtime speedup |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 4 | 64 | 1.018 | 0.789 | -22.5% | 0.3847 | 0.3285 | 1.17x |
| 4 | 256 | 1.041 | 0.893 | -14.3% | 0.6940 | 0.4949 | 1.40x |
| 8 | 64 | 1.368 | 1.361 | -0.5% | 0.6443 | 0.5150 | 1.25x |
| 8 | 256 | 1.356 | 1.469 | +8.3% | 1.2899 | 0.7921 | 1.63x |
| 16 | 64 | 2.183 | 3.042 | +39.3% | 1.4328 | 1.0067 | 1.42x |
| 16 | 256 | 2.329 | 3.013 | +29.4% | 3.1455 | 1.3590 | 2.31x |

At 16 segments and batch 256, temporary GPU memory decreases from 459 MB on `origin/main` to 87 MB on this PR.

### PCS

CPU, one environment:

| Segments | main compile (s) | PR compile (s) | Δ compile | main runtime (ms) | PR runtime (ms) | Runtime speedup |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 0.376 | 0.275 | -26.9% | 0.0335 | 0.0219 | 1.53x |
| 4 | 0.428 | 0.435 | +1.6% | 0.1057 | 0.0780 | 1.36x |
| 8 | 0.453 | 0.490 | +8.2% | 0.2124 | 0.1644 | 1.29x |
| 12 | 0.418 | 0.578 | +38.3% | 0.4169 | 0.3133 | 1.33x |
| 16 | 0.426 | 0.713 | +67.3% | 0.9831 | 0.5261 | 1.87x |
| 32 | 0.490 | 1.177 | +140.1% | 17.4480 | 2.1842 | 7.99x |

GPU:

| Segments | Environments | main compile (s) | PR compile (s) | Δ compile | main runtime (ms) | PR runtime (ms) | Runtime speedup |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 4 | 64 | 1.526 | 1.307 | -14.3% | 0.7254 | 0.6069 | 1.20x |
| 4 | 256 | 1.618 | 1.281 | -20.8% | 1.1865 | 0.7755 | 1.53x |
| 8 | 64 | 2.382 | 2.674 | +12.3% | 1.1595 | 0.8214 | 1.41x |
| 8 | 256 | 2.515 | 2.778 | +10.4% | 2.8923 | 1.4652 | 1.97x |
| 16 | 64 | 3.400 | 5.523 | +62.4% | 3.5916 | 1.9735 | 1.82x |
| 16 | 256 | 3.722 | 5.054 | +35.8% | 9.7907 | 3.8442 | 2.55x |

At 16 segments and batch 256, temporary GPU memory decreases from 1.73 GB on `origin/main` to 270 MB on this PR.

### GVS

“Homogeneous” uses fixed joints and a constant six-DOF strain basis per link. “Mixed” cycles through revolute/planar/helical joints, monomial/Legendre/Fourier bases, different basis orders and active DOF counts, and different quadrature counts.

CPU, one environment:

| Basis | Segments | DOF | main compile (s) | PR compile (s) | Δ compile | main runtime (ms) | PR runtime (ms) | Runtime speedup |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| homogeneous | 1 | 6 | 0.435 | 0.455 | +4.7% | 0.0748 | 0.0504 | 1.49x |
| homogeneous | 2 | 12 | 0.357 | 0.555 | +55.7% | 0.1507 | 0.0937 | 1.61x |
| homogeneous | 4 | 24 | 0.371 | 0.571 | +54.1% | 0.3133 | 0.1514 | 2.07x |
| homogeneous | 8 | 48 | 0.393 | 0.558 | +42.1% | 0.6494 | 0.3421 | 1.90x |
| homogeneous | 12 | 72 | 0.403 | 0.530 | +31.4% | 1.1155 | 0.6143 | 1.82x |
| homogeneous | 16 | 96 | 0.417 | 0.537 | +28.8% | 1.9919 | 1.0017 | 1.99x |
| mixed | 1 | 9 | 0.444 | 0.484 | +8.9% | 0.0976 | 0.0598 | 1.63x |
| mixed | 2 | 19 | 0.422 | 0.533 | +26.2% | 0.1994 | 0.1157 | 1.72x |
| mixed | 3 | 36 | 0.427 | 0.540 | +26.5% | 0.3464 | 0.1939 | 1.79x |
| mixed | 4 | 47 | 0.440 | 0.541 | +22.9% | 0.4632 | 0.2660 | 1.74x |
| mixed | 6 | 79 | 0.472 | 0.561 | +18.9% | 1.0009 | 0.6475 | 1.55x |

GPU, 64 environments:

| Basis | Segments | DOF | main compile (s) | PR compile (s) | Δ compile | main runtime (ms) | PR runtime (ms) | Runtime speedup |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| homogeneous | 1 | 6 | 1.052 | 1.281 | +21.7% | 0.6825 | 0.3945 | 1.73x |
| homogeneous | 2 | 12 | 0.983 | 1.354 | +37.7% | 1.5330 | 0.8287 | 1.85x |
| homogeneous | 4 | 24 | 1.190 | 1.501 | +26.1% | 2.9424 | 1.4006 | 2.10x |
| homogeneous | 8 | 48 | 1.740 | 1.678 | -3.6% | 5.8006 | 2.6205 | 2.21x |
| mixed | 1 | 9 | 1.117 | 1.305 | +16.8% | 0.9456 | 0.5356 | 1.77x |
| mixed | 2 | 19 | 1.193 | 1.653 | +38.6% | 2.1180 | 1.1250 | 1.88x |
| mixed | 3 | 36 | 1.868 | 2.061 | +10.3% | 3.3596 | 1.6966 | 1.98x |
| mixed | 4 | 47 | 1.750 | 2.080 | +18.9% | 4.2376 | 1.9996 | 2.12x |
| mixed | 6 | 79 | 2.822 | 2.470 | -12.5% | 8.1190 | 3.6996 | 2.19x |

At six mixed segments and batch 64, temporary GPU memory decreases from 293 MiB on `origin/main` to 48 MiB on this PR.

Over the measured `origin/main` and current-PR grids, GVS compile-time slopes are approximately:

| Scenario | main slope (s/segment) | PR slope (s/segment) |
| --- | ---: | ---: |
| CPU, homogeneous | 0.0015 | 0.0016 |
| CPU, mixed | 0.0069 | 0.0129 |
| GPU, homogeneous | 0.107 | 0.056 |
| GPU, mixed | 0.341 | 0.224 |

Warm runtime improves in every measured PlanarPCS, PCS, and GVS scenario relative to `origin/main`. Compilation is a tradeoff for larger PlanarPCS/PCS CPU models and some larger GPU models, while GVS compilation remains close to the scaling of `origin/main` and becomes faster than `origin/main` for the largest measured GPU models.

## Validation

- `ruff check` and `ruff format --check` pass for the changed code.
- The complete `PlanarPCS`, `PCS` and `GVS` system suites pass.
- Focused GVS public-matrix/dynamics tests pass after the final explicit scan-result unpacking: `2 passed`.
- Forward- and reverse-mode GVS differentiation tests pass for one-, two-, and three-link heterogeneous models: `6 passed`.
- The passive-threadlike GVS forward-dynamics regression passes: `1 passed`.
- The current GitHub Actions build job passes; the multi-version Python test matrix runs on the pushed PR head.


src/soromox/systems/pcs/pcs.py
```diff
@@ -3083,6 +3091,96 @@ def jacobian_at_s(s_local_ij: Array) -> tuple[Array, Array]:
 
         return Ws_inner, g_ps, J_ps, Jd_or_Jd_qd_ps
 
+    @eqx.filter_jit
+    def _dynamics_integration_kinematics(
+        self, q: Array, qd: Array
+    ) -> tuple[Array, Array, Array, Array]:
+        """Return quadrature kinematics for dynamics without absolute poses."""
```
mstoelzle: Please make the docstring more extensive (for all private methods that you are editing) and aligned in format with the rest of the file and package

src/soromox/systems/pcs/pcs.py
```diff
@@ -3083,6 +3091,96 @@ def jacobian_at_s(s_local_ij: Array) -> tuple[Array, Array]:
...
+        xi = self.strain(q).reshape(self.num_segments, 6)
+        xid = (self.B_xi @ qd).reshape(self.num_segments, 6)
+        B_segments = self.B_xi.reshape(self.num_segments, 6, self.num_dofs)
```
mstoelzle: Why not call this `B_xi_segments` to be more consistent?

tools/benchmarks/benchmark_pcs_dynamics_candidate.py
mstoelzle, line 1: Please remove this file from git

src/soromox/systems/pcs/pcs.py
mstoelzle: Addressed in f4917dc. I expanded the docstrings for all private PCS methods touched by this PR, including behavior, arguments, return values, and array shapes, using the existing package format.

tools/benchmarks/benchmark_pcs_dynamics_candidate.py
mstoelzle: Addressed in f4917dc. The benchmark candidate file has been removed from Git.

src/soromox/systems/pcs/pcs.py
mstoelzle: Addressed in f4917dc. I renamed the segment strain basis consistently to B_xi_segments and the per-segment value to B_xi_i throughout the affected kinematics helpers.

vdperfetta01 (commented):
Really nice performance! The code and documentation look coherent and well-described. One question about the PR description: the validation section states that everything was successfully tested for the GVS case, but results are also presented above for the PlanarPCS/PCS case. If I understand correctly, those tests were run as well... should we explicitly mention that in the validation section too?

mohammedtarnini:
I guess I am a bit late as the PR was already merged and closed. Just wanted to say incredible performance boost you got there! And I just had one question, why only coadjoint_action? cant this method of doing action without storing matrices be also done to other functions in se2 and se3 and help enhance performance even further?

mstoelzle:
Thanks, Mohammed! This was an excellent question. I investigated the other matrix-free Lie-group actions systematically in #166, including direct small-adjoint, adjoint, inverse-adjoint, left-Jacobian, and matrix-RHS actions on single-environment CPU and batched GPU workloads.

The direct SE(3) small-adjoint action was the only sufficiently portable improvement, so #166 applies it to the contracted PCS and GVS dynamics recurrences. The other variants produced some substantial local gains, but also platform- or model-size-dependent regressions. The PR includes the complete benchmark results and considered alternatives:
https://github.com/tud-phi/soromox/pull/166

PR #165, opened 2026-08-24
Codex/jittable parameter updates

matteodv99tn:
## Summary

Adds focused regressions for JIT-compiled partial parameter updates and proposes fixes for the discovered failures.

## Problem

Existing compiled-update tests pass a complete parameter PyTree dynamically. They do not cover a closed-over model where only one top-level field is traced, such as `update_params(base_pose=...)`. Unchanged nested parameter arrays can then trigger tracer conversion failures during validation.

## Coverage

- GVS base-pose update followed by `potential_force`
- Threadlike actuator and passive component updates
- McKibben actuator bounds update
- Articulated tendon actuator and passive component updates

This PR contains regression tests and proposed fixes, as anticipated in https://github.com/tud-phi/soromox/pull/156#issuecomment-5393249187.

At minimum, the regression tests should be retained. The accompanying fixes are a proposal and may be taken as-is, revised, or replaced by an implementation that better fits SoRoMoX.

## Bonus content

<details>
<summary>Standalone python script that creates a passive GVS system and evaluates its generalised forces as function of both base pose and Lagrangian coordinates.</summary>

```python
import jax
import jax.numpy as jnp
import equinox as eqx

from soromox.systems.components.joints import JointSpec
from soromox.systems.components.links import LinkSpec
from soromox.systems.gvs.core import GVS
from soromox.systems.gvs.specs import GVSSegment, StrainBasisSpec

link = LinkSpec.circular(
    length=1.0,
    radius=0.05,
    density=1000.0,
    reference_strain=[1.0, 1.0, 1.0, 1.0, 1.0, 1.0],
    young_modulus=1e8,
    poisson_ratio=0.3,
    # stiffness=10.0*jnp.eye(6),
)
joint = JointSpec.fixed()
strain_basis = StrainBasisSpec(
    type="monomial",
    strain_selector=[True, True, True, False, False, False],
    basis_order=(1, 1, 2, 0, 0, 0),
)
segment = GVSSegment(
    link=link,
    joint=joint,
    basis=strain_basis,
    num_gauss_points=5,
)

system = GVS.from_segments((segment, ),
                           actuators=(),
                           gravity=jnp.array([0.0, -9.81, 0.0]))

@jax.jit
def evaluate(base, q):
    return system.update_params(base_pose=base).potential_force(q)

qs = jnp.zeros((system.num_dofs), )
base1 = jnp.array([1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0])
base2 = jnp.array([0.5, 0.5, 0.5, 0.5, 0.0, 0.0, 0.0])
assert base1.shape == (7, )
assert base2.shape == (7, )

print(evaluate(base1, qs))
print(evaluate(base2, qs))

```
</details>

### DIFF
--- a/src/soromox/systems/gvs/core.py
+++ b/src/soromox/systems/gvs/core.py
@@ -843,7 +843,6 @@ class GVS(SoftRobot):
         Raises:
             ValueError: If link params do not contain one entry per segment.
         """
-        link.validate_for_update()
         if link.length.shape != (self.num_segments,):
             raise ValueError(
                 "link params must contain one entry per segment, "
--- a/src/soromox/systems/params.py
+++ b/src/soromox/systems/params.py
@@ -8,12 +8,14 @@ __all__ = [
     "validate_quaternion_base_pose",
 ]
 
+from contextlib import suppress
 from dataclasses import fields
 from typing import Any, ClassVar, Literal, final
 
 import equinox as eqx
 from jax import Array, core, tree_util
 from jax import numpy as jnp
+from jax.errors import ConcretizationTypeError, TracerBoolConversionError
 
 DEFAULT_GRAVITY_MAGNITUDE = 9.81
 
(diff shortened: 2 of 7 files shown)


mstoelzle:
Thanks @matteodv99tn for isolating these partial-update cases and opening this PR. I incorporated the regression coverage into PR #156 in [`a71894a`](https://github.com/tud-phi/soromox/commit/a71894a74f653983aa5662650828b3bc05532670), and implemented the revised validation strategy in [`5bfd97a`](https://github.com/tud-phi/soromox/commit/5bfd97a91b5970ead6484b16ee41442ee836a6a8).

The implementation now validates closed concrete values at JAX trace time, keeps structural checks active for dynamic tracer leaves, and separates component-to-robot structural checks from concrete value checks. This also avoids suppressing unrelated routing validation merely because another component parameter is traced.

Since PR #156 now contains both the regressions and the corresponding fix, I am closing this PR in favor of #156.

PR #171, opened 2026-08-25
Add optional Coriolis-free dynamics paths

mstoelzle:
## Motivation

Some simulation and optimization workloads deliberately use a reduced dynamics
model that neglects Coriolis and centrifugal forces. Soromox previously offered
no coherent way to request this approximation: callers could manually modify a
dense equation-of-motion solve, but system-specific `forward_dynamics` paths and
the rollout APIs would still compute the physical velocity-bias terms.

The goal of this PR is therefore to provide one opt-in interface across every
`SoftRobot` family while making the reduced path genuinely cheaper. The existing
physical dynamics remain the default.

This PR builds on the device/affinity/warmup benchmark infrastructure merged in
#170. Its diff against current `main` contains the Coriolis API, optimized
dynamics paths, coverage, and Coriolis-specific benchmark support.

## API

- Add the static Equinox model setting `consider_coriolis: bool` to `SoftRobot`.
- Default it to `True`, preserving current behavior and physical accuracy.
- Accept the setting in direct constructors and shared `from_links` /
  `from_segments` construction helpers:

  ```python
  robot = PCS.from_links(links, consider_coriolis=False)
  ```

- Preserve the setting across immutable `with_params` and parameter-update
  workflows while keeping it outside physical parameter PyTrees.
- Keep the canonical method name `dynamics_terms`; no singular alias is added.
- Let `rollout_to`, `rollout_closed_loop_to`, and
  `rollout_discrete_closed_loop_to` inherit the configured model through their
  existing `forward_dynamics` calls, without changing the generic
  `DynamicalSystem` signatures.

The setting is static rather than a per-call method argument because each mode
has a materially different compiled recurrence. Equinox/JAX can eliminate the
unused branch and produce a smaller executable; changing the setting therefore
creates a separate optimized JIT executable instead of carrying a runtime
conditional through every integration step.

## Dynamics semantics

When `consider_coriolis=False`:

- `dynamics_terms(q, qd)` returns `(M, zeros_like(qd), G)`;
- every `forward_dynamics` implementation omits the Coriolis force; and
- damping, velocity-dependent actuation, gravity, elasticity, and external
  generalized forces continue to use their normal inputs and equations.

An explicit `coriolis_matrix(q, qd)` call remains physically exact for
diagnostics. Kinetic and total energy values and derivatives are also unchanged;
the custom energy JVP uses the explicit physical Coriolis API where the identity
requires it rather than interpreting the dynamics approximation as a change in
kinetic energy.

### Shared and articulated systems

- The base `SoftRobot` and `Pendulum` paths skip `coriolis_matrix` entirely when
  the approximation is selected.
- ArticulatedSoftRobot and McKibben UMArm bypass the ABA velocity-bias,
  motion-cross, and force-cross recurrences, while retaining the real velocity
  for damping and actuation.
- Correct the ABA gravity initialization for non-identity base orientations by
  transforming world gravity into the base frame before the acceleration
  recurrence.

### Continuum systems

- PlanarPCS and PCS use position/Jacobian/gravity-only recurrences and avoid
  strain rates, tangent directional derivatives, `Jdot @ qdot`, spatial
  velocities, and coadjoint momentum assembly.
- GVS uses Jacobian-only joint and Magnus helpers plus a reduced recurrence that
  carries Jacobians, gravity, inertia, and gravity force only.
- PlanarHSA reuses its pose/Jacobian geometry path and skips Jacobian-derivative
  and convective assembly.
- Specialized PCS/GVS systems inherit the same setting and optimized path.

Small assembly differences remain inline. Dedicated private helpers are used
only where removing Coriolis changes the recurrence state and required inputs
substantially; forcing those paths through one scan would retain unnecessary
arrays and derivatives. The serial Jacobian/gravity propagation shared by PCS
and PlanarPCS lives on `SoftRobot`, with independent scans because measurements
showed that combining both carries slightly regressed CPU performance.

## Correctness

Parameterized regression coverage verifies that:

- omitted/default construction matches explicit `consider_coriolis=True`;
- disabled dynamics preserve `M` and `G`, return an exactly zero `Cqd`, and match
  a manual no-Coriolis forward solve;
- explicit Coriolis matrices and energy values, gradients, and JVPs are
  unchanged;
- all three rollout variants honor the configured model;
- derivative and convective helpers are not invoked on reduced paths;
- JIT specialization, immutable parameter updates, damping, actuation, gravity,
  elasticity, and external-force behavior remain valid; and
- inherited McKibben, pressure-actuated PCS, and shared continuum-component
  construction paths preserve the setting.

## Performance

Measurements use JAX FP64 on the CPU backend of an Intel Core Ultra 9 285K.
The process and all JAX worker threads are pinned with `taskset -c 1` to one
5.7 GHz performance core; no efficiency core or GPU backend participates.
Continuum systems use five Gauss points per segment. Each separately compiled
case receives a one-second synchronized warmup followed by 20 synchronized
samples, and the median warm latency is reported. Rollouts integrate 2 ms with
`solver_dt=0.1 ms` and one final saved sample.

Every cell below is `enabled / disabled`: the compile-time ratio followed by the
median warm-runtime ratio. A ratio greater than 1 means the Coriolis-disabled
path is faster. Results are reported separately for every measured link/segment
count.

| System | Links / segments | `dynamics_terms` compile / warm | `forward_dynamics` compile / warm | `rollout_to` compile / warm |
| ArticulatedSoftRobot | 1 | 1.53x / 1.12x | 1.48x / 1.87x | 1.30x / 1.71x |
| ArticulatedSoftRobot | 4 | 1.86x / 1.57x | 1.30x / 1.16x | 1.02x / 1.63x |
| ArticulatedSoftRobot | 8 | 1.76x / 1.84x | 1.30x / 1.40x | 1.16x / 1.68x |
| ArticulatedSoftRobot | 16 | 1.76x / 2.62x | 1.29x / 1.27x | 1.31x / 1.98x |
| PlanarPCS | 1 | 2.00x / 1.11x | 2.10x / 1.07x | 1.27x / 1.37x |
| PlanarPCS | 4 | 1.89x / 1.27x | 1.77x / 1.28x | 1.55x / 1.77x |
| PlanarPCS | 8 | 1.77x / 1.29x | 1.67x / 1.33x | 1.54x / 1.67x |
| PlanarPCS | 16 | 1.65x / 1.46x | 1.73x / 1.51x | 1.58x / 1.57x |
| PCS | 1 | 1.95x / 1.67x | 1.78x / 1.85x | 1.56x / 2.74x |
| PCS | 4 | 1.86x / 1.33x | 1.76x / 1.35x | 1.50x / 1.80x |
| PCS | 8 | 1.96x / 1.52x | 1.77x / 1.51x | 1.53x / 1.59x |
| PCS | 16 | 1.84x / 1.48x | 1.79x / 1.39x | 1.47x / 1.40x |
| GVS | 1 | 1.77x / 1.28x | 1.70x / 1.19x | 1.51x / 1.73x |
| GVS | 4 | 1.79x / 1.53x | 1.75x / 1.53x | 1.51x / 1.17x |
| GVS | 8 | 1.75x / 1.50x | 1.69x / 1.51x | 1.50x / 1.18x |
| GVS | 16 | 1.73x / 1.37x | 1.52x / 1.36x | 1.30x / 1.33x |

Warm-runtime geometric means across the four sizes are:

| System | `dynamics_terms` | `forward_dynamics` | `rollout_to` | All methods |
| ArticulatedSoftRobot | 1.71x | 1.40x | 1.75x | 1.61x |
| PlanarPCS | 1.28x | 1.29x | 1.59x | 1.38x |
| PCS | 1.49x | 1.51x | 1.82x | 1.60x |
| GVS | 1.42x | 1.39x | 1.33x | 1.38x |

An XLA cost audit at 16 segments corroborates the timings. The reduced PlanarPCS
`dynamics_terms` executable uses approximately 17% fewer FLOPs, 25% less memory
traffic, and 37% fewer transcendental operations; GVS uses approximately 13%
fewer FLOPs, 36% less memory traffic, and 38% fewer transcendental operations.

The benchmark command, per-size table, and raw artifact paths are also retained
in the development benchmarking guide.

## Validation

- Ruff check and format — passed for all changed Python files.
- PlanarPCS system suite — 123 passed.
- PCS system suite — 120 passed.
- GVS system suite — 88 passed.
- PlanarHSA system suite — 23 passed.
- ArticulatedSoftRobot and Pendulum suites — 65 passed.
- SoftRobot defaults, pressure-actuated PCS, shared continuum construction, and
  McKibben targeted suites — passed.
- Benchmark registry, benchmark CLI, and all rollout variants — 38 passed after
  rebasing onto the `main` commit containing #170.
- `zensical build` — passed.
- CPU-only Coriolis benchmark smoke test pinned to performance core 1 — passed.

No GPU backend was queried or initialized during the benchmark runs.

### DIFF
--- a/docs/api/systems/index.md
+++ b/docs/api/systems/index.md
@@ -32,6 +32,22 @@ The `DynamicalSystem` class is the fundamental base class for all dynamical syst
 
 The `SoftRobot` class extends `DynamicalSystem` with interfaces specific to soft robots, including methods for forward kinematics and Jacobians parameterized by arc-length or position along the robot backbone.
 
+All concrete soft-robot systems accept the static model setting
+`consider_coriolis`, which defaults to `True`. Construct a model with
+`consider_coriolis=False` to omit Coriolis and centrifugal forces from
+`dynamics_terms`, `forward_dynamics`, and rollouts while leaving the explicit
+`coriolis_matrix` diagnostic and all energy methods physically exact. Because
+the setting is static, enabled and disabled models compile to separate optimized
+JAX executables.
+
+```python
+robot = PCS.from_links(links, consider_coriolis=False)
+```
+
+The configured setting is inherited by `rollout_to`, `rollout_closed_loop_to`,
+and `rollout_discrete_closed_loop_to`; the generic rollout signatures do not
+change.
+
 ::: soromox.systems.SoftRobot
     options:
       show_root_heading: true
--- a/docs/development/benchmarking.md
+++ b/docs/development/benchmarking.md
@@ -34,6 +34,7 @@ python tools/benchmarks/benchmark_system_methods.py \
   --device cpu \
   --systems articulated_soft_robot pendulum planar_pcs pcs \
   --segment-counts 1 3 5 7 \
+  --coriolis-modes enabled disabled \
   --duration 2.0 \
   --solver-dt 5e-4 \
   --warmup-duration 1 \
@@ -54,6 +55,9 @@ python tools/benchmarks/benchmark_system_methods.py \
   CPU frequency scaling and runtime worker threads before collecting samples.
 - `--execution-repeats`: number of synchronized calls whose median is reported
   after the cold run and optional warmup.
+- `--coriolis-modes`: benchmark enabled and/or disabled static model settings.
+  Each result row records `consider_coriolis`, and plots use distinct lines for
+  both modes.
 - `--json` / `--csv`: export raw results for regression tracking.
 - `--plot` / `--show-plot`: render Matplotlib summaries (compile vs. exec time).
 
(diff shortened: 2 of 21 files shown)


Michele-Martini (approved):
The PR enables optional opt-out of the Coriolis term in the EOM of the robot classes, showing increase of computational speed in all cases. The code appears structured and organized, with targeted changes.

mstoelzle:
After reviewing the trade-off, we believe that the current performance gains do not outweigh the added code complexity and maintenance burden. Supporting this option requires dedicated, specialized implementations for many methods and kernels, which would make the codebase more challenging to maintain. We are therefore closing this PR for now.

PR #173, opened 2026-08-26
Add shape-generic Warp backend for batched GVS dynamics

mstoelzle:
## Summary

This PR adds an optional, shape-generic Warp execution backend for GVS dynamics while preserving the true GVS equations and the existing differentiable JAX implementation.

The production interface is now the existing `GVS.dynamics_terms(q, qd, backend=...)` method rather than a separate batched API. It accepts either one state `(num_dofs,)` or a batch `(batch_size, num_dofs)`. The model-level `backend` setting is deliberately general enough to govern additional accelerated GVS operations in the future:

- `backend="auto"` (default): use Warp for GPU primal execution and JAX on CPU;
- `backend="warp"`: explicitly request the Warp primal path;
- `backend="jax"`: explicitly request the reference JAX path;
- any forward- or reverse-mode differentiation: use JAX, even when the primal policy selects Warp.

There is intentionally no batch-size, basis-order, or model-size crossover in the GPU policy. Such thresholds are hardware-dependent. GPU execution consistently selects Warp; CPU auto-dispatch consistently selects JAX.

The Warp implementation supports:

- spatially varying strain bases, including basis order greater than zero;
- two-point Magnus integration along every link cell;
- fixed, revolute, prismatic, helical, cylindrical, planar, spherical, free, and mixed joint chains;
- runtime-varying segment count, local/global DoF counts, quadrature count, basis order, and batch size;
- FP64 assembly of the inertia matrix `B`, contracted Coriolis term `C @ qd`, and gravity force `G`;
- scalar calls, explicit batched calls, and `jax.vmap` without degenerating into one batch-one Warp launch pipeline per environment;
- normal use from `GVS.forward_dynamics` and therefore from the existing rollout APIs.

Warp remains optional through `soromox[warp]`; importing Soromox does not eagerly require it.

## Motivation and constraints

GVS dynamics is not a conventional rigid-body tree algorithm or a nodal cable/rod solver. Each segment propagates pose-dependent Lie quantities, Jacobians, Jacobian-rate contractions, velocity, and gravity through a serial chain while distributed inertia and force contributions are integrated at quadrature points. Higher-order strain bases vary spatially, and general joints add SE(3) propagation between links.

This creates two practical GPU problems in the existing batched JAX implementation:

1. the generated XLA graph grows with model structure and contains many small Lie-algebra, recurrence, and assembly operations;
2. the ordered recurrence inside one environment offers less coarse parallelism than a rigid-body/contact workload, especially at small batches.

The implementation therefore keeps the GVS algorithm intact and reorganizes its execution around the parallelism that actually exists:

- environments provide the outer grid;
- rows, columns, active DoF pairs, joints, and cells provide cooperative work within an environment;
- segment/cell dependencies remain physically ordered;
- invariant topology and numerical operands are flattened and cached once;
- generated Warp source is independent of the concrete robot shape.

The data organization takes inspiration from MuJoCo Warp and Newton: flat persistent model arrays, precomputed topology maps, predictable output ownership, bounded runtime-shaped kernels, and cooperative reuse. It does **not** copy their rigid-body, constraint, or flexible-cable dynamics algorithms. The two-point Magnus link integration, general-joint Lie propagation, distributed quadrature, and GVS equations remain the source of truth.

## Branch contents relative to `main`

After rebasing, the Warp-specific commits in this PR are:

1. `4440dd7` — adds the shape-generic benchmark harness, investigated kernel families, and phase-by-phase report.
2. `b4a7b0b` — records the measured CPU regressions and CPU dispatch decision.
3. `7286fa3` — integrates the first persistent, shape-generic Warp backend into `GVS`.
4. `1140105` — compacts invariant basis data, moves repeated quantities out of runtime kernels, hardens general-joint/model-map handling, and makes GPU auto-dispatch consistently choose Warp.
5. `4809c38` — adds cooperative joint preparation, recurrence-value reuse, weighted-mass caching, upper-triangular inertia assembly, and the selected high-DoF launch configuration.
6. `ea439fa` — unifies scalar and batched execution under `GVS.dynamics_terms`, adds model-level backend configuration, custom batching, transform-aware JAX differentiation, and automatic forward-dynamics integration.

### Unified API and backend policy

`GVS` has a static `backend: ExecutionBackend` field. `ExecutionBackend` is exported from both `soromox.systems.gvs` and `soromox.systems`. The name describes an execution policy rather than a GVS-specific algorithm and leaves room for the same setting to control future accelerated operations.

`GVS.dynamics_terms()` validates scalar or batched states and resolves an optional per-call override against the model setting. JAX and Warp keep separate compiled implementation boundaries behind this small dispatcher. The dispatcher itself is intentionally not wrapped in `eqx.filter_jit`: differentiation and batching rules must select their implementation before a forward-only Warp primal is staged. The selected JAX assembly and Warp batch pipeline remain JIT compiled.

The former `dynamics_terms_batched()` compatibility wrapper was removed because there were no production callers and the unified method is clearer. Existing code can write either:

```python
B, Cqd, G = model.dynamics_terms(q, qd)
B_batch, Cqd_batch, G_batch = model.dynamics_terms(q_batch, qd_batch)
```

### Transform-aware batching

The Warp primitive has scalar-call semantics at the public boundary, with a `jax.custom_batching.custom_vmap` rule underneath. When `jax.vmap` maps independent `q` and `qd` states, the rule gathers the mapped axis and invokes **one true batched Warp pipeline**. It does not emit `batch_size` independent batch-one pipelines.

This is important for the existing simulation stack: callers can continue to use normal JAX transformations and the GVS implementation still exposes all environments together to the GPU. Explicit two-dimensional calls use the same batching rule and kernel path.

Batching model parameters themselves is rejected because the persistent model arrays describe one robot topology; batching environment states for that model is fully supported.

### Transform-aware differentiation

Warp is used only for primal forward execution. `evaluate_terms` is an Equinox custom-JVP function whose derivative rule evaluates the existing JAX `_assemble_dynamics_terms` implementation. Reverse mode is obtained by JAX transposing that JVP, so both forward- and reverse-mode transformations use the differentiable JAX equations rather than attempting to differentiate the forward-only Warp FFI.

`GVS.forward_dynamics` has the same policy at the complete equation level. A small static custom-JVP bridge surrounds the compiled forward-dynamics implementation; its derivative rule explicitly selects `backend="jax"` for term assembly, forces, and the inertia solve. The bridge is static because `eqx.filter_custom_jvp` is a callable transform rather than a Python method descriptor.

Consequences:

- ordinary GPU simulation uses Warp automatically;
- `jax.jvp`, `jax.linearize`, `jax.grad`, `jax.jacrev`, and reverse-mode transformations use JAX automatically;
- callers do not need to remember to mutate the model backend before differentiating;
- primal outputs retain the Warp numerical path outside transformations.

### Forward dynamics and rollouts

`GVS.forward_dynamics` now resolves the configured backend internally. Therefore existing Diffrax-based `rollout_to`/closed-loop rollout calls reach the selected dynamics backend without a parallel Warp-specific rollout API.

No manual `wp.capture_*` call is embedded in `GVS`. JAX/XLA owns command-buffer/CUDA-graph capture and replay of the compiled computation. The custom batching rule is what makes this useful: each vector-field evaluation presents one batch-shaped Warp primitive instead of a mapped collection of batch-one FFI calls.

Nsight Systems traces of the production methods establish the capture granularity. Ten warm `GVS.forward_dynamics` queries used one graph executable and exactly ten `cuGraphLaunch` calls; each Warp preprocessing/chain kernel and each dense-solve kernel executed ten times. `GVS.rollout_to` also replays captured graphs, but Diffrax/Tsit5 is not captured as one monolithic graph: two 100-step rollouts produced 2,834 graph launches and 1,214 forward-dynamics evaluations. The graphs therefore cover command-buffer fragments inside the solver stages rather than an entire rollout call.

### Persistent model representation

`GVS.precompute()` builds runtime arrays used by the shape-generic kernels:

- padded joint/link local-to-global and global-to-local coordinate maps;
- the active serial coordinate prefix after every segment;
- compact link-basis row selectors and values at both Magnus evaluation points;
- cell widths and length-scaled basis values;
- inner quadrature mass diagonals and preweighted mass diagonals;
- base-frame gravity;
- row/column maps for the upper triangle of the global inertia matrix.

The GVS basis has at most one active spatial row per local basis column. Packing the row and its scalar values avoids repeatedly loading dense `6 x max_dof` basis matrices without assuming constant strain. `update_params()` recomputes all dependent cached operands when model parameters change.

### General-joint and cell preprocessing

One warp-sized block owns each `(environment, segment)` joint. Six lanes form joint strain and strain rate from the padded basis; the block computes shared SE(3) exponential/left-Jacobian coefficients and cooperatively emits inverse adjoint terms, local tangent terms, tangent-rate contraction, and joint velocity.

Joint behavior is encoded by the existing padded basis, so generated source does not specialize on a Python joint class. The same kernel supports all current joint families and mixed chains.

Each link cell evaluates the true two-point Magnus expression and its directional derivative, then emits compact local tangents and Lie quantities consumed by the recurrence. Runtime loop bounds preserve higher-order spatial variation. The production path never assumes `B_Z1 == B_Z2`, a vanishing Magnus commutator, or fixed joints.

### Persistent active-prefix chain

One cooperative block owns one environment and advances segments/cells in physical order while parallelizing matrix/vector work within each step. Two scratch states alternate as recurrence buffers, avoiding a separate JAX operation or launch at every segment boundary.

For a serial GVS chain, segment `s` can only depend on coordinates introduced at or before it. The implementation derives that prefix from the actual joint/link coordinate maps and reduces dominant assembly work from `S * N^2` toward `sum_s N_s^2`. The guard falls back to the full DoF count if a model map is not a contiguous active prefix; it does not assume fixed joints or a particular strain order.

The retained kernel also:

- computes `J @ qd` once per recurrence state and shares it;
- uses preweighted mass diagonals;
- assembles only the upper triangle of symmetric inertia and mirrors it;
- writes final `B`, `C @ qd`, and `G` plus bounded double-buffer scratch;
- uses 128 lanes for models up to 64 DoFs and 192 lanes above 64 DoFs, based on measured occupancy/throughput sweeps.

### Compilation behavior

Segment count, basis order, quadrature count, joint family, global/local DoFs, and batch size are runtime dimensions. Warp compiles a small fixed set of modules rather than regenerating unrolled CUDA source for every robot.

True empty-cache measurements reached the first completed Warp result in 0.901 s for a four-segment order-1 revolute model and 0.942 s for a 16-segment mixed-joint model; both shapes produced identical Warp module hashes. With Warp modules cached, lower plus XLA compilation was 55 ms and 63 ms respectively, compared with about 1.02 s and 1.08 s for JAX. The new dispatch/custom-transform layer does not introduce shape-specialized Warp generation.

## Why this implementation is faster

The speedup is cumulative:

1. matrix-free Lie actions avoid temporary `6 x 6` matrices when only contractions are needed;
2. sparse topology, basis structure, gravity, quadrature weights, and mass terms are packed once;
3. each joint/cell Lie quantity is computed by one natural owner and reused;
4. the persistent chain removes per-segment JAX/FFI recurrence boundaries;
5. active-prefix processing removes structurally zero global-DoF work early in the chain;
6. recurrence velocities are shared rather than recomputed;
7. weighted masses reduce inner-loop arithmetic and loads;
8. upper-triangle assembly nearly halves inertia output-pair work;
9. the 128/192-lane selection exposes useful pair parallelism without the resource loss measured at 256 lanes;
10. custom batching preserves the large batch presented by callers and rollouts.

Relative to the production backend before the final cooperative/cache/triangular changes (`1140105`), the retained kernel changes reduced runtime by approximately 20–30% for order 1 at the priority batches and about 47.5% for order 5 with nine Gauss points.

## Important alternatives evaluated

The benchmark report contains 23 principal implementation families plus focused follow-ups. The main negative results are retained because they explain the final boundaries.

- **Per-cell launches and per-segment CUDA graphs:** reduced launch overhead and beat JAX at moderate batches, but materialized recurrence states globally and approached parity at batch 1024.
- **One serial thread per environment:** removed launches but discarded row/column parallelism and created register/local-memory pressure; it was an order of magnitude slower than cooperative mappings.
- **One block per environment/segment:** useful at small/medium batches, but synchronization and segment-tip handoff lost throughput after saturation.
- **One owner per `6 x N` Lie output:** recomputed the same strain, Magnus coefficients, and SE(3) exponential up to 144 times per cell in the original shape.
- **Fully monolithic Lie + recurrence + quadrature:** failed the compile-feasibility gate; even a smaller recurrence/quadrature form did not finish Warp compilation within six minutes.
- **Constant-strain/fixed-joint specialization:** fast for exact PCS-like order-zero/fixed-joint models, but invalid for general GVS and still took about 31.6 s through the first shape-specialized result.
- **Joint propagation left in JAX:** worked, but preserved a size-dependent JAX graph and prevented the final persistent chain. Moving it to Warp enabled general mixed-joint fusion.
- **One global owner per final output:** deterministic and atomic-free, but reread the entire chain for each output entry and was slower at every batch.
- **Fusing joint propagation separately into every segment block:** duplicated block-local joint work and did not repay the added scheduling/synchronization.
- **Unguarded whole-chain persistence:** initially underoccupied the RTX 5090 at batch 64. It became successful only after active-prefix sparsity, shared recurrence values, weighted masses, and triangular inertia reduced work per block.
- **Cooperative cell layouts:** some helped batch 64 but regressed batch 256 because small cells did not amortize block scheduling, shared memory, and synchronization. Cooperative joint preprocessing was retained because joints contain enough shared SE(3) work.
- **Warp dense solve variants:** helped only at the non-priority large-batch end or only below 64 DoFs. The production backend therefore accelerates term assembly and retains the mature JAX dense solve.
- **Column-owned inertia and 256-lane blocks:** increased repeated quadrature reads or reduced occupancy. Upper-triangle pair ownership with 128/192 lanes was consistently better.

The major remaining bottleneck is the serial per-environment chain plus dense `N x N` inertia assembly. At small batches there are fewer environment blocks than SMs; at high order the active pair count and quadrature work dominate. Further gains should preserve runtime-sized modules and target local joint-tangent storage, cell/recurrence data movement, and multi-block high-order assembly rather than returning to shape-unrolled monolithic kernels.

## Updated benchmarks

Environment: RTX 5090, CUDA 13 driver, Warp 1.16.0, JAX 0.11.0, FP64. Results are synchronized interleaved medians. The current production API was used for both paths (`backend="jax"` versus `backend="warp"`), and every Warp result was compared with JAX before timing.

### GPU dynamics terms: 4 segments, basis order 1, revolute joints (52 DoFs)

| Environments | JAX | Warp | Speedup |
|---:|---:|---:|---:|
| 1 | 1.108 ms | 0.502 ms | 2.21x |
| 64 | 1.711 ms | 0.582 ms | 2.94x |
| 256 | 2.311 ms | 0.630 ms | 3.67x |
| 1024 | 6.007 ms | 1.497 ms | 4.01x |

Maximum absolute term error: `3.44e-12`.

### GPU dynamics terms: 4 segments, basis order 5, 9 Gauss points, revolute joints (148 DoFs)

This is the important coupled high-order case: increasing strain order also increases quadrature resolution.

| Environments | JAX | Warp | Speedup |
|---:|---:|---:|---:|
| 1 | 1.623 ms | 1.467 ms | 1.11x |
| 64 | 3.051 ms | 1.610 ms | 1.89x |
| 256 | 5.913 ms | 2.067 ms | 2.86x |
| 1024 | 21.886 ms | 7.265 ms | 3.01x |

Maximum absolute term error: `1.02e-12`.

### CPU dynamics terms: one environment

| Model | JAX | Warp | Warp regression |
|---|---:|---:|---:|
| order 1, 52 DoFs | 0.270 ms | 0.438 ms | 1.62x slower |
| order 5 / 9 points, 148 DoFs | 0.864 ms | 1.563 ms | 1.81x slower |

This is why `auto` selects JAX on CPU. The regression is documented rather than hidden behind a GPU crossover policy.

### GPU production `GVS.forward_dynamics`

These calls use the complete public method: Warp/JAX dynamics terms, elastic and damping forces, external/actuation handling, and the existing JAX dense inertia solve.

| Basis / quadrature | Environments | JAX | Warp | Speedup |
|---|---:|---:|---:|---:|
| order 1 | 1 | 1.325 ms | 0.685 ms | 1.93x |
| order 1 | 64 | 1.769 ms | 0.708 ms | 2.50x |
| order 1 | 256 | 2.442 ms | 0.909 ms | 2.69x |
| order 1 | 1024 | 6.456 ms | 1.917 ms | 3.37x |
| order 5 / 9 points | 1 | 2.156 ms | 2.015 ms | 1.07x |
| order 5 / 9 points | 64 | 4.139 ms | 2.776 ms | 1.49x |
| order 5 / 9 points | 256 | 7.560 ms | 3.679 ms | 2.05x |
| order 5 / 9 points | 1024 | 27.109 ms | 12.615 ms | 2.15x |

The inertia solve amplifies small term-ordering differences in absolute units because these test states produce large accelerations. Maximum scale-normalized errors remained below `1.7e-10` for order 1 and `1.3e-9` for order 5.

### GPU production `GVS.rollout_to`

These measurements call the standard Diffrax-based `GVS.rollout_to`, vectorized over environments exactly as other system-method benchmarks are. They use the default Tsit5 solver, constant `1e-5 s` solver steps, 100 requested steps (`1e-3 s` duration), and only initial/final state saves. A `1e-4 s` step was rejected because this stiff test model produced non-finite trajectories in both implementations.

| Basis / quadrature | Environments | JAX | Warp | Speedup |
|---|---:|---:|---:|---:|
| order 1 | 1 | 652.843 ms | 285.776 ms | 2.28x |
| order 1 | 64 | 979.368 ms | 363.113 ms | 2.70x |
| order 1 | 256 | 1,350.015 ms | 406.848 ms | 3.32x |
| order 5 / 9 points | 1 | 1,022.919 ms | 955.234 ms | 1.07x |
| order 5 / 9 points | 64 | 2,359.880 ms | 1,496.849 ms | 1.58x |
| order 5 / 9 points | 256 | 4,345.825 ms | 2,034.396 ms | 2.14x |

Maximum scale-normalized trajectory errors remained below `4.7e-10` for order 1 and `2.2e-10` for order 5.

### CPU production methods: one environment

| Method | Model | JAX | Warp | Warp regression |
|---|---|---:|---:|---:|
| `forward_dynamics` | order 1 | 0.342 ms | 0.462 ms | 1.35x slower |
| `forward_dynamics` | order 5 / 9 points | 0.891 ms | 1.589 ms | 1.78x slower |
| `rollout_to` | order 1 | 86.219 ms | 177.472 ms | 2.06x slower |
| `rollout_to` | order 5 / 9 points | 454.034 ms | 719.335 ms | 1.58x slower |

These complete-method results reinforce the JAX-on-CPU auto-dispatch decision.

### CUDA graph evidence

Nsight Systems 2025.5.2 traces used CUDA graph node tracing around warmed production calls at order 1, batch 64.

| Production call | Profiled calls | `cuGraphLaunch` | Forward-kernel instances | Interpretation |
|---|---:|---:|---:|---|
| `GVS.forward_dynamics` | 10 | 10 | 10 | one replayed graph per query |
| `GVS.rollout_to` (100 steps, Tsit5) | 2 | 2,834 | 1,214 | captured stage fragments, not one graph per rollout |

Thus graph capture works as intended for standalone forward queries. It is active and beneficial inside `rollout_to`, but full-rollout capture remains a separate optimization opportunity.

### Compilation scaling

| Segments | Cached Warp lower + XLA compile | JAX lower + XLA compile | Empty-cache Warp through first result |
|---:|---:|---:|---:|
| 4 | 55 ms | 1.02 s | 0.901 s |
| 16 | 63 ms | 1.08 s | 0.942 s |

The same Warp module hashes are used at both shapes, so CUDA generation does not grow with the number of links.

## Correctness and validation

- Every low-level benchmark checks `B`, `C @ qd`, and `G` against JAX before timing; production-method benchmarks compare complete outputs or trajectories.
- Current GPU term errors are approximately `1e-12` to `4e-12`.
- Mixed-joint validation covering every supported joint family stayed below `9.17e-12`.
- Tests verify scalar and explicit-batch calls, real Warp execution, and heterogeneous/length-scaled models.
- A custom-vmap test verifies that mapped scalar calls produce one batch-shaped Warp invocation.
- JVP, linearization, and reverse-mode tests verify that derivative transforms use JAX and do not stage Warp.
- Complete `forward_dynamics` JVP/gradient tests verify the same policy through forces and the inertia solve.
- Production `forward_dynamics` and `rollout_to` benchmarks reject non-finite outputs and report both absolute and scale-normalized differences.
- Parameter-update tests verify that compact bases, weighted masses, and gravity caches track model changes.
- The complete GVS CPU test module passed (`95 passed`); focused post-dispatch CPU and GPU Warp/transform tests also passed.
- `ruff check` and `git diff --check` pass for all committed production and test files.

## Known limitations

- Warp currently requires FP64 `q` and `qd`.
- Warp accelerates primal term assembly; derivative transformations intentionally execute the JAX implementation.
- The persistent kernel targets the current serial-chain GVS topology and does not introduce a rigid-body tree.
- The production forward-dynamics path retains the JAX dense inertia solve.
- CPU auto-dispatch intentionally remains on JAX.

## Recommendation

Use the unified API with the default `backend="auto"`: JAX on CPU, Warp on GPU, and JAX whenever derivatives are requested. This gives one hardware policy at the model boundary, preserves standard `jax.vmap` and rollout usage, and avoids exposing benchmark-era kernel option names in production code.

The resulting implementation supports spatially varying GVS strain and all current joint families, is 1.89–3.67x faster than JAX at the priority GPU batches for the measured order-1 and coupled order-5 models, and retains a sub-second, nearly flat first-result compilation law across four and 16 segments.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -128,6 +128,10 @@ dependencies = [
 cuda13 = [
     "jax[cuda13]>=0.10.0; sys_platform == 'linux'",
     "torch>=2.13.0; sys_platform == 'linux'",
+    "warp-lang>=1.16.0; sys_platform == 'linux'",
+]
+warp = [
+    "warp-lang>=1.16.0",
 ]
 dev = [
     "bump2version",
@@ -245,6 +249,7 @@ all = [
     "tox",
     "trimesh",
     "viser",
+    "warp-lang>=1.16.0",
 ]
 
 # List URLs that are relevant to your project
--- a/src/soromox/systems/__init__.py
+++ b/src/soromox/systems/__init__.py
@@ -53,6 +53,7 @@ from soromox.systems.system_state import EnvironmentState, SystemState
 from .articulated import ArticulatedSoftRobot, McKibbenActuatedUMArm
 from .gvs import (
     GVS,
+    GVSBackend,
     GVSSegment,
     StrainBasisSpec,
 )
@@ -94,6 +95,7 @@ __all__ = [
     "McKibbenActuatedUMArmParams",
     # gvs systems
     "GVS",
+    "GVSBackend",
     "GVSParams",
     "GVSStructure",
     "GVSSegmentStructure",
--- a/src/soromox/systems/__init__.py
+++ b/src/soromox/systems/__init__.py
@@ -53,7 +53,7 @@ from soromox.systems.system_state import EnvironmentState, SystemState
 from .articulated import ArticulatedSoftRobot, McKibbenActuatedUMArm
 from .gvs import (
     GVS,
-    GVSBackend,
+    ExecutionBackend,
     GVSSegment,
     StrainBasisSpec,
 )
@@ -94,8 +94,8 @@ __all__ = [
     "McKibbenActuatedUMArm",
     "McKibbenActuatedUMArmParams",
     # gvs systems
+    "ExecutionBackend",
     "GVS",
-    "GVSBackend",
     "GVSParams",
     "GVSStructure",
     "GVSSegmentStructure",
--- a/src/soromox/systems/gvs/__init__.py
+++ b/src/soromox/systems/gvs/__init__.py
@@ -7,12 +7,12 @@ from soromox.systems.gvs.structures import (
     GVSStructure,
 )
 
-from .core import GVS, GVSBackend
+from .core import GVS, ExecutionBackend
 from .specs import GVSSegment, StrainBasisSpec
 
 __all__ = [
+    "ExecutionBackend",
     "GVS",
-    "GVSBackend",
     "GVSSegment",
     "StrainBasisSpec",
     "GVSParams",
--- a/src/soromox/systems/gvs/__init__.py
+++ b/src/soromox/systems/gvs/__init__.py
@@ -7,11 +7,12 @@ from soromox.systems.gvs.structures import (
     GVSStructure,
 )
 
-from .core import GVS
+from .core import GVS, GVSBackend
 from .specs import GVSSegment, StrainBasisSpec
 
 __all__ = [
     "GVS",
+    "GVSBackend",
     "GVSSegment",
     "StrainBasisSpec",
     "GVSParams",
--- a/src/soromox/systems/gvs/_dynamics.py
+++ b/src/soromox/systems/gvs/_dynamics.py
@@ -1,0 +1,98 @@
+"""Transform-aware execution of GVS dynamics terms."""
+
+from __future__ import annotations
+
+from typing import Any, Protocol
+
+import equinox as eqx
+import jax
+import jax.numpy as jnp
+from jax import Array
+
+DynamicsTerms = tuple[Array, Array, Array]
+
+
+class DynamicsModel(Protocol):
+    """Structural interface required by the backend dispatcher."""
+
+    num_dofs: int
+
+    def _assemble_dynamics_terms(self, q: Array, qd: Array) -> DynamicsTerms: ...
+
+
+@eqx.filter_jit
+def _execute_batch(model: DynamicsModel, q: Array, qd: Array) -> DynamicsTerms:
+    """Execute one shape-generic Warp launch pipeline for a state batch."""
+
+    if q.shape[0] == 0:
+        raise ValueError("The Warp backend requires a non-empty batch.")
+    if q.dtype != jnp.float64 or qd.dtype != jnp.float64:
+        raise TypeError(
+            "The Warp GVS backend currently requires float64 q and qd; "
+            f"got {q.dtype} and {qd.dtype}."
+        )
+
+    try:
+        from soromox.systems.gvs._warp.backend import dynamics_terms
+    except ModuleNotFoundError as error:
+        if error.name == "warp":
+            raise ImportError(
+                "The Warp GVS backend requires the optional 'warp-lang' "
+                "dependency. Install it with `pip install soromox[warp]`."
+            ) from error
+        raise
+
+    if jax.default_backend() == "gpu":
+        lanes_per_block = 192 if model.num_dofs > 64 else 128
+    else:
+        lanes_per_block = 1
+    return dynamics_terms(model, q, qd, lanes_per_block=lanes_per_block)
+
+
+@jax.custom_batching.custom_vmap
+def _execute_primal(model: DynamicsModel, q: Array, qd: Array) -> DynamicsTerms:
+    """Give the batched Warp implementation scalar-call semantics."""
+
+    batched = _execute_batch(model, q[None, :], qd[None, :])
+    return jax.tree.map(lambda value: value[0], batched)
+
+
+@_execute_primal.def_vmap
+def _vmap_rule(
+    axis_size: int,
+    in_batched: tuple[Any, bool, bool],
+    model: DynamicsModel,
+    q: Array,
+    qd: Array,
+) -> tuple[DynamicsTerms, tuple[bool, bool, bool]]:
+    """Map independent environments to one batched Warp execution."""
+
+    model_batched, q_batched, qd_batched = in_batched
+    if any(jax.tree.leaves(model_batched)):
+        raise ValueError("Batching over GVS model parameters is not supported.")
+    if not q_batched:
+        q = jnp.broadcast_to(q, (axis_size, *q.shape))
+    if not qd_batched:
+        qd = jnp.broadcast_to(qd, (axis_size, *qd.shape))
+    return _execute_batch(model, q, qd), (True, True, True)
+
+
+@eqx.filter_custom_jvp
+def evaluate_terms(model: DynamicsModel, q: Array, qd: Array) -> DynamicsTerms:
+    """Evaluate Warp dynamics while retaining differentiable JAX semantics."""
+
+    return _execute_primal(model, q, qd)
+
+
+@evaluate_terms.def_jvp
+def _jvp_rule(
+    primals: tuple[DynamicsModel, Array, Array],
+    tangents: tuple[Any, Array | None, Array | None],
+) -> tuple[DynamicsTerms, DynamicsTerms]:
+    """Use the JAX assembly for both forward- and reverse-mode derivatives."""
+
+    return eqx.filter_jvp(
+        lambda model, q, qd: model._assemble_dynamics_terms(q, qd),
+        primals,
+        tangents,
+    )
--- a/src/soromox/systems/gvs/_warp/__init__.py
+++ b/src/soromox/systems/gvs/_warp/__init__.py
@@ -1,0 +1,1 @@
+"""Optional forward-only Warp backend for batched GVS dynamics."""
--- a/src/soromox/systems/gvs/_warp/backend.py
+++ b/src/soromox/systems/gvs/_warp/backend.py
@@ -10,7 +10,6 @@ from jax import Array
 from soromox.systems.gvs._warp.joint import scalable_joint_terms
 from soromox.systems.gvs._warp.persistent import scalable_persistent_chain
 from soromox.systems.gvs._warp.scalable import scalable_cell_terms
-from soromox.utils.lie_algebra import se3
 
 if TYPE_CHECKING:
     from soromox.systems.gvs.core import GVS
@@ -81,22 +80,16 @@ def _cell_terms(
     max_dof = model.max_dof
     work_items = batch_size * num_segments * num_cells
     matrix_rows = work_items * SPATIAL_DIM
-    basis_z1 = model.B_Z1
-    basis_z2 = model.B_Z2
-    if model.scale_rotational_basis_by_length:
-        scales = model.segment_lengths[:, None, None]
-        basis_z1 = basis_z1.at[:, :, :3].divide(scales)
-        basis_z2 = basis_z2.at[:, :, :3].divide(scales)
-    cell_widths = model.integration_points[:, 1:] - model.integration_points[:, :-1]
     outputs = scalable_cell_terms(
         q_link.reshape(batch_size * num_segments, max_dof),
         qd_link.reshape(batch_size * num_segments, max_dof),
-        basis_z1.reshape(num_segments * num_cells * SPATIAL_DIM, max_dof),
-        basis_z2.reshape(num_segments * num_cells * SPATIAL_DIM, max_dof),
+        model.scaled_B_Z1_values.reshape(num_segments * num_cells, max_dof),
+        model.scaled_B_Z2_values.reshape(num_segments * num_cells, max_dof),
+        model.link_basis_rows,
         model.xi_ref_Z1.reshape(num_segments * num_cells, SPATIAL_DIM),
         model.xi_ref_Z2.reshape(num_segments * num_cells, SPATIAL_DIM),
         model.segment_lengths,
-        cell_widths.reshape(num_segments * num_cells),
+        model.cell_widths.reshape(num_segments * num_cells),
         jnp.asarray([num_cells], dtype=jnp.int32),
         jnp.asarray([0], dtype=jnp.int32),
         output_dims={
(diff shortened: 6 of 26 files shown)


mstoelzle:
Superseded by #177, which combines the complete GVS Warp backend with the PlanarPCS and PCS Warp specializations in one main-targeted PR.

PR #174, opened 2026-08-27
Optimize JAX GVS dynamics with bounded prefix buckets

mstoelzle:
## Summary

This PR transfers the useful, differentiable parts of #173 into the existing JAX GVS dynamics path without adding a second backend:

- reuse the body velocity already carried by the segment/cell recurrence instead of recomputing `J @ qd`;
- store each link strain-basis column as its one active spatial row plus two scalar Magnus-point values;
- reduce `B`, `C @ qd`, and `G` only over bounded active-DOF prefixes;
- select among at most K reusable prefix widths inside the existing outer `lax.scan`; and
- make K user-configurable as `num_dynamics_prefix_buckets`, with K=4 as the default.

The production implementation contains one dynamics path only. The previous full-width reduction and the exact/full-prefix segment-unrolled implementation are not retained as alternatives. This preserves the fixed recurrence graph and avoids changing compilation-time scaling from bounded control flow to one separately traced body per segment.

The implementation remains pure JAX, supports forward- and reverse-mode differentiation, and does not change the GVS equations, joint families, strain bases, or quadrature scheme.

## Branch strategy

The phase-by-phase implementation and benchmark history was first preserved and pushed as `codex/pr173-jax-evaluation`. The exact unrolled prefix implementation is additionally preserved on `codex/pr173-jax-active-prefix-snapshot` at `38ad507b90`.

This PR comes from the clean branch `codex/jax-dynamics-prefix-buckets`, created directly from `origin/main` (`683fb66`). It contains a single production commit and only two changed repository files: the GVS implementation and focused GVS tests. Benchmark scripts and reports used during the investigation are deliberately not included.

## Motivation and design constraint

The original fixed-shape JAX implementation already uses one outer segment `lax.scan` and one inner cell `lax.scan`, which gives it a valuable property: the recurrence/control-flow graph does not grow with segment count. Its main runtime inefficiency is different. At segment `s`, the Jacobian can only depend on the active coordinates introduced up to `s`, but the three global quadrature reductions still process all final system DOFs.

The exact active-prefix experiment removed all of that padded work and achieved the best isolated runtime. However, it used a Python loop with one concretely sized body per segment. The generated StableHLO and compiler time consequently grew linearly with segment count. For the high-order CPU case, compiler time increased by about 0.45 seconds per additional segment and reached about 15.3 seconds at 32 segments, versus about 2.3 seconds for the original scan.

The selected design keeps the scan and introduces only a fixed number of reusable reduction shapes. It therefore trades a small amount of remaining padded arithmetic for a bounded graph.

## Public API and parameter name

Both `GVS(...)` and `GVS.from_segments(...)` now accept:

```python
num_dynamics_prefix_buckets: int = 4
```

The value must be a positive non-boolean integer. The effective branch count is capped by the segment count. Because it is an Equinox static field, changing it intentionally produces a separately compiled executable.

`num_dynamics_prefix_buckets` was selected because it identifies:

1. that this is a count (`num_`);
2. that it affects only dynamics compilation/execution; and
3. that each branch represents an active-coordinate prefix bucket.

The alternatives were less precise: `branch_budget` exposed an implementation detail, `num_reduction_buckets` was too broad, and `num_active_dof_buckets` did not identify the dynamics path.

K=1 or K=2 can be appropriate when cold compilation dominates. K=8 can be useful for a long-lived, throughput-dominated executable. K=4 was the best tested general Pareto point and is therefore the default.

### Compact link-basis operands

Every GVS link basis column has at most one structurally active spatial row. `precompute()` now stores:

- the row index of each padded local basis column;
- the scalar basis value at the first Magnus point; and
- the scalar basis value at the second Magnus point.

The compact cell helper reconstructs strain and strain rate from those row/value operands and computes the same two-point Magnus increment, tangent, tangent directional derivative, `B_Magnus`, and `B_Magnus_dot` as the dense helper. Padded columns are masked. Fourier widths and optional rotational length scaling are handled explicitly.

The cached scalar values are refreshed by `with_params()` when segment lengths or other dynamic parameters change.

### Recurrence-velocity reuse

The segment/cell recurrence already propagates body velocity. The contracted Jacobian-derivative recurrence now consumes that value directly at joints and cells instead of repeatedly forming `J_previous @ qd`. This is algebraically identical because the recurrence maintains `velocity_previous == J_previous @ qd`.

### Equal-count active-prefix buckets

Construction records the cumulative active-coordinate prefix after each segment. The ordered segments are divided into at most K equally populated groups. Each group ends at an actual active prefix, so its reduction branch slices the quadrature Jacobians to that width, performs the three contractions, and pads the result back to the final active-coordinate shape.

Only the global reductions are inside `lax.switch`. Joint evaluation, Magnus cell evaluation, and both causal recurrences remain in their single reusable scan bodies.

This is intentionally different from branching on local segment DOF or `(DOF, Gauss points)`. A segment's global prefix depends on every preceding segment, not only on its own local shape. Grouping equal local shapes in model order can still help bucket placement indirectly, but it does not justify duplicating the cell evaluator.

## Measurement method

- JAX 0.11.0, Python 3.12.13, FP64.
- CPU: Intel Core Ultra 9 285K. Every CPU benchmark was pinned to high-performance logical core 1 with `taskset -c 1` and `OMP_NUM_THREADS=1`; the benchmark aborted unless affinity was exactly `(1,)`.
- GPU: NVIDIA GeForce RTX 5090, 32 GB, driver 595.84.
- GVS segments: 1, 4, 8, 16, and 32.
- Coupled strain-order/Gauss-point pairs: 0/5, 1/5, 3/7, and 5/9.
- CPU: one environment. GPU: batches 1, 64, 256, and 1024; the 32-segment endpoint was measured through batch 256.
- Runtime: synchronized medians after 8--10 warm-up calls, normally over 30 CPU or 15 GPU samples.
- Compilation-scaling cells: fresh Python/JAX processes. Lowering, `lowered.compile()`, first execution, StableHLO text size, operation count, `case` count, and `while` count were recorded separately.
- Candidate values and JVP directional derivatives were compared with the original fixed-width scan before timing.

## Pareto optimization of the branch budget

The representative order-5 / 9-Gauss-point results below are fresh-process compiler seconds / warmed term-assembly milliseconds.

| Device/workload | Full width | K=2 | K=4 | K=8 | Exact prefix |
|---|---:|---:|---:|---:|---:|
| CPU, S=16, B=1 | 1.97 / 13.04 | 2.07 / 9.49 | 2.18 / 8.34 | 2.48 / 8.11 | 7.90 / 5.88 |
| GPU, S=16, B=256 | 3.35 / 146.98 | 3.58 / 112.00 | 4.54 / 96.39 | 7.17 / 89.28 | 14.99 / 57.97 |
| GPU, S=16, B=1024 | 4.39 / 576.79 | 5.21 / 429.40 | 6.64 / 369.35 | 10.77 / 346.14 | 19.83 / 219.88 |
| GPU, S=32, B=256 | 4.40 / 1004.56 | 5.13 / 752.84 | 6.83 / 646.87 | 10.65 / 606.36 | not run |

K=4 captures about 59--68% of the exact-prefix runtime gain in these large cases. K=8 improves runtime only another 4--7% while roughly doubling incremental compile cost. K=16 was only another 2--3% faster than K=8. K=4 is therefore the knee of the measured compile/runtime frontier.

### StableHLO graph law

| Segments | CPU full width | CPU K=4 | CPU exact | GPU B=256 full width | GPU B=256 K=4 | GPU B=256 exact |
|---:|---:|---:|---:|---:|---:|---:|
| 4 | 2,933 | 3,147 | 3,260 | 3,121 | 3,347 | 3,480 |
| 8 | 2,933 | 3,147 | 4,048 | 3,121 | 3,347 | 4,364 |
| 16 | 2,933 | 3,147 | 5,624 | 3,121 | 3,347 | 6,132 |
| 32 | 2,933 | 3,147 | 8,776 | 3,121 | 3,347 | not run |

The exact path adds 197 CPU or 221 batched-GPU StableHLO operations per segment. K=4 adds a fixed 214 CPU or 226 batched-GPU operations once all branches exist and then plateaus. Both full-width and K=4 retain two `while` operations; K=4 adds one bounded `case`.

StableHLO text still grows with tensor dimensions and embedded constants. That is shape growth, not recurrence-body duplication: operation count and control-flow topology stay fixed.

### Fresh-process CPU compiler time

Cells are full width / K=4 / exact seconds.

| Segments | order 1 / GQ 5 | order 5 / GQ 9 |
|---:|---:|---:|
| 4 | 1.878 / 2.043 / 2.697 | 1.922 / 2.110 / 2.846 |
| 8 | 1.867 / 2.052 / 4.350 | 1.969 / 2.162 / 4.488 |
| 16 | 1.967 / 2.118 / 7.858 | 1.988 / 2.260 / 7.898 |
| 32 | 1.937 / 2.106 / 15.255 | 2.294 / 2.468 / 15.326 |

Linear compiler slopes over 4--32 segments:

| Path | order 1 | order 5 |
|---|---:|---:|
| Full width | 0.0026 s/segment | 0.0132 s/segment |
| K=4 | 0.0023 s/segment | 0.0128 s/segment |
| Exact prefix | 0.4502 s/segment | 0.4472 s/segment |

K=4 preserves the full-width CPU scaling law. At 32 segments it compiles in 2.11--2.47 seconds, versus 15.26--15.33 seconds for exact prefix.

### Fresh-process GPU compiler time

High-order cells are full width / K=4 / exact seconds.

| Segments | batch 1 | batch 256 | batch 1024 |
|---:|---:|---:|---:|
| 4 | 1.74 / 3.55 / 3.80 | 2.64 / 4.06 / 4.50 | 2.81 / 4.89 / 5.82 |
| 8 | 1.72 / 3.25 / 6.45 | 2.88 / 4.30 / 8.26 | 3.09 / 4.70 / 8.84 |
| 16 | 1.68 / 3.30 / 12.30 | 3.15 / 4.54 / 14.99 | 4.42 / 6.64 / 19.83 |
| 32 | 1.88 / 3.27 / not run | 4.36 / 6.83 / not run | not run |

At batch 1, K=4 is flat from 4--32 segments while exact prefix adds about 0.71 seconds per segment. At 16 segments, K=4 compiles 67--73% faster than exact prefix depending on batch size.

K=4 compiler time still grows at high order and large batch despite a constant 3,347-operation graph. The same shape/layout/GEMM code-generation effect exists in the full-width scan. It is fundamentally different from the exact path's per-segment graph growth.

## Warmed runtime

Representative dynamics-term assembly:

| Workload | Full width | K=4 | Speed-up |
|---|---:|---:|---:|
| CPU, S=16, order 5/GQ 9, B=1 | 12.65 ms | 8.04 ms | 36.4% |
| CPU, S=32, order 5/GQ 9, B=1 | 100.95 ms | 68.04 ms | 32.6% |
| GPU, S=16, order 5/GQ 9, B=256 | 146.76 ms | 94.32 ms | 35.7% |
| GPU, S=16, order 5/GQ 9, B=1024 | 578.00 ms | 367.33 ms | 36.4% |
| GPU, S=32, order 5/GQ 9, B=256 | 1.007 s | 0.664 s | 34.1% |

Across the pinned-CPU order/Gauss sweep, S=16 improved 24.9--36.4% and S=32 improved 28.6--33.2%. GPU batch-1 includes regressions at some small shapes (up to 15.3% for the measured S=4/order-5 case); the target multi-environment workloads improve consistently once branch overhead is amortized.

Complete forward dynamics retains smaller but material gains because the dense solve dilutes term-assembly savings:

| Segments | order / GQ | CPU B=1 | GPU B=64 | GPU B=256 |
|---:|:---|---:|---:|---:|
| 4 | 5 / 9 | 32.2% | 10.4% | 17.2% |
| 8 | 1 / 5 | 16.6% | 2.2% | 10.2% |
| 8 | 5 / 9 | 28.5% | 10.8% | 24.1% |
| 16 | 1 / 5 | 27.3% | 7.3% | 13.6% |
| 16 | 5 / 9 | 31.2% | 20.4% | 26.8% |

At S=16/order-5, complete forward dynamics falls from 14.46 to 9.95 ms on CPU and from 193.17 to 141.38 ms on GPU B=256.

### Heterogeneous segment shapes

Alternating order-3/GQ-7 and order-5/GQ-9 links, and the same shapes grouped contiguously, were both evaluated:

| Layout | CPU B=1 | GPU B=256 |
|---|---:|---:|
| Alternating | 8.79 -> 5.63 ms (35.9%) | 103.59 -> 68.97 ms (33.4%) |
| Equal shapes grouped | 8.85 -> 5.28 ms (40.3%) | 103.54 -> 65.79 ms (36.5%) |

Grouping equal local shapes is modestly useful as a model-ordering property, but a separate compiled evaluator branch per shape is not.

## PlanarPCS and PCS transfer evaluation

Velocity reuse and compact selector bases were implemented experimentally and measured separately for both PCS systems over S=1/4/8/16, GQ=5/9, CPU B=1, and GPU B=1/256/1024. CPU runs used 100 synchronized samples on core 1. GPU runs used 100 samples; critical compact-basis cases were repeated in three independent fresh processes.

Numerical differences were exactly zero for PlanarPCS and at most `5.3e-23` for PCS.

Pinned-CPU medians (baseline / velocity only / compact only):

| System/workload | Runtime | Compiler time |
|---|---:|---:|
| PlanarPCS, S=4, GQ=5 | 0.030 / 0.031 / 0.034 ms | 1.028 / 1.011 / 1.014 s |
| PlanarPCS, S=16, GQ=9 | 0.127 / 0.127 / 0.121 ms | 2.720 / 2.735 / 2.719 s |
| PCS, S=4, GQ=5 | 0.050 / 0.049 / 0.048 ms | 1.514 / 1.471 / 1.460 s |
| PCS, S=16, GQ=9 | 0.389 / 0.390 / 0.388 ms | 3.004 / 3.019 / 2.990 s |

Across the CPU grid, median velocity-reuse speed-up was 0.4% for PlanarPCS and 1.7% for PCS. Compact-basis medians were 4.3% and 0.5%, but the absolute differences were only microseconds and included regressions.

Three-run GPU medians for compact basis at critical large shapes:

| System/workload | Runtime change | Compiler change |
|---|---:|---:|
| PlanarPCS, S=8, GQ=5, B=256 | 0.580 -> 0.617 ms (-6.0%) | 1.349 -> 1.162 s (+16.1%) |
| PlanarPCS, S=16, GQ=5, B=1024 | 3.050 -> 2.895 ms (+5.4%) | 2.565 -> 1.877 s (+36.7%) |
| PlanarPCS, S=16, GQ=9, B=1024 | 4.129 -> 4.080 ms (+1.2%) | 2.706 -> 2.737 s (-1.2%) |
| PCS, S=8, GQ=5, B=256 | 1.125 -> 1.263 ms (-10.9%) | 2.002 -> 1.898 s (+5.5%) |
| PCS, S=16, GQ=5, B=1024 | 9.486 -> 9.255 ms (+2.5%) | 4.146 -> 2.925 s (+41.7%) |
| PCS, S=16, GQ=9, B=1024 | 15.738 -> 15.625 ms (+0.7%) | 7.117 -> 7.290 s (-2.4%) |

Velocity reuse added only 8--11 StableHLO operations and compact basis added 55--67, independent of segment count, so neither worsened the existing PCS graph law. However, velocity reuse was neutral overall on GPU (median -0.2% for PlanarPCS and -0.3% for PCS), and compact basis was inconsistent. Its large compile benefit was mostly confined to 5-Gauss-point, 16-segment GPU cases and disappeared at 9 Gauss points.

Neither PCS experiment is retained. They add representation/update-path complexity without a portable runtime win. The narrow compact-basis compiler result remains useful evidence for possible future work focused specifically on large, low-quadrature PCS models.

## Alternatives evaluated and rejected

The preserved evaluation history contains each phase. The important negative results are:

### Exact per-segment active prefixes

This gives the best small/medium runtime but adds 197--221 StableHLO operations per segment and about 0.45 CPU compiler seconds per segment. It is the fundamental compilation-law regression this PR avoids.

### Bucket placement policies

- Uniform prefix widths are reasonable but less aligned with the serial segment distribution.
- A nominal padding-cost-optimal policy reduced arithmetic on paper but was 1--4% slower on GPU, apparently because its irregular GEMM widths were less efficient.
- Equal segment counts were simple, robust, and selected.

### Grouping by local DOF or `(DOF, Gauss points)`

Separate local evaluator branches grew StableHLO from about 3,151 to 4,726--4,891 operations and were neutral to 5% slower. Equal local DOF does not imply equal global active prefix, so it does not address the dominant reduction padding.

### Bucketing the recurrence transforms

Also specializing the 6-by-N recurrence work added 91--171 StableHLO operations and regressed runtime by about 3--7% on CPU and 1--13% on GPU.

### Dynamic tiled reductions

Fixed tiles of 32/64/128 preserved a bounded graph but created many small GEMMs and regressed CPU/GPU runtime by 19--45%.

### Larger branch budgets

K=8 is on the Pareto table but not the default. K=16 bought only 2--3% over K=8 while doubling branch count.

### Other arithmetic/cache ideas

Preweighted quadrature/cached transforms, matrix-free adjoints, triangular inertia assembly, and fused evaluation were individually neutral or slower in A/B tests and are not retained.

## Correctness and validation

- K=1, default K=4, and K greater than the segment count produce matching `B`, `C @ qd`, and `G`.
- JVP directional derivatives match across those public bucket settings.
- Experimental K=2/4/8 policies and exact prefix were also compared with the original fixed-width values and JVPs before timing.
- Compact operands are refreshed by parameter updates.
- The full existing GVS suite passed: `88 passed`.
- New bucket configuration, value equivalence, JVP equivalence, and validation tests passed: `5 passed`.
- `tests/systems/test_planar_pcs.py`, `tests/systems/test_pcs.py`, and `tests/systems/test_pcs_2d_vs_3d_coherence.py`: `263 passed`.
- `ruff check src/soromox/systems/gvs/core.py tests/systems/test_gvs.py`.
- `git diff --cached --check`.

Per the requested scope, the unrelated repository-wide suite was not rerun.

## Generalization and tuning guidance

The bounded branch count and constant StableHLO operation count after K is saturated are structural properties and should generalize across CPUs and GPUs.

The exact K=4 runtime Pareto point is less portable. CPU vector/cache behavior, GPU `case` dispatch, GEMM shape efficiency, compiler version, batch size, and strain/quadrature order all move the crossover. The default should therefore be understood as the best balanced value on the measured Core Ultra 9 285K / RTX 5090 environment, not as a universal optimum.

The user-configurable static parameter is the portability mechanism:

- prefer K=1/2 when cold compilation and small models dominate;
- keep K=4 for the measured general balance; and
- consider K=8 for long-lived, large-batch/high-order executables after benchmarking the target hardware.

This keeps policy explicit without maintaining multiple dynamics algorithms or reintroducing segment-dependent graph growth.

### DIFF
--- a/src/soromox/systems/gvs/core.py
+++ b/src/soromox/systems/gvs/core.py
@@ -102,8 +102,14 @@ class GVS(SoftRobot):
         max_num_integration_points: Maximum number of integration points per link (= max_num_gauss_points + 2).
         num_dofs: Total number of DOFs for the robot (sum of joint + link DOFs).
         num_padded_dofs: Theoretical maximum number of DOFs (num_segments * 2 * max_dof).
-        num_dynamics_prefix_buckets: Maximum number of active-DOF prefix
-            widths used by the dynamics reductions.
+        num_dynamics_prefix_buckets: Upper bound on the number of specialized
+            active-DOF widths compiled for the dynamics reductions. A segment
+            can depend only on its own and preceding segments' DOFs, so ordered
+            segments are grouped into at most this many buckets and each bucket
+            reduces only up to its largest required DOF prefix. Higher values
+            remove more zero-padded arithmetic, but add compiled branches and
+            can increase JAX compilation time. The default is 4; a value of 1
+            uses one full-width reduction branch.
         active_dof_map: Basis matrix mapping active DOFs into the full padded
             joint/link coordinate space.
         scale_rotational_basis_by_length: If True, apply scaling to the angular component of the strain basis matrix for improved numerical stability.
@@ -266,9 +272,14 @@ class GVS(SoftRobot):
                 model.
             passive_elements: Optional passive element or tuple of passive
                 elements. Pass ``None`` or an empty tuple to disable them.
-            num_dynamics_prefix_buckets: Maximum number of active-DOF prefix
-                widths compiled for the dynamics reductions. Larger values can
-                reduce padded arithmetic at the cost of additional compilation.
+            num_dynamics_prefix_buckets: Upper bound on the number of
+                specialized active-DOF widths compiled for dynamics. Ordered
+                segments are grouped into at most this many buckets, and each
+                bucket reduces only up to the largest causal DOF prefix needed
+                by one of its segments. Higher values remove more zero-padded
+                arithmetic, but add compiled branches and can increase JAX
+                compilation time. The default is 4; a value of 1 uses one
+                full-width reduction branch.
             **kwargs: Additional keyword arguments forwarded to
                 :class:`BaseContinuumSoftRobot`.
 
@@ -389,8 +400,12 @@ class GVS(SoftRobot):
             max_num_gauss_points: Optional common quadrature padding size.
             scale_rotational_basis_by_length: Whether to normalize rotational
                 strain-basis rows by link length.
-            num_dynamics_prefix_buckets: Maximum number of active-DOF prefix
-                widths compiled for the dynamics reductions.
+            num_dynamics_prefix_buckets: Upper bound on the number of
+                specialized active-DOF widths compiled for dynamics. Segments
+                are grouped in order, and each bucket reduces only up to its
+                largest required causal DOF prefix. Higher values trade more
+                compiled branches for less zero-padded arithmetic. The default
+                is 4; a value of 1 uses one full-width reduction branch.
             **kwargs: Additional keyword arguments forwarded to the GVS
                 constructor, such as actuators or passive elements.
 
@@ -790,7 +805,30 @@ class GVS(SoftRobot):
     def _compact_link_basis_operands(
         self, segment_lengths: Array
     ) -> tuple[Array, Array, Array]:
-        """Pack the structurally one-row columns of each link strain basis."""
+        """Pack the structurally one-row columns of every link strain basis.
+
+        Every active generalized-coordinate column in a link basis belongs to
+        exactly one of the six strain rows. Storing that row index and its value
+        avoids dense ``(6, self.max_dof)`` matrix products in the dynamics path.
+        Columns used only for padding receive row index ``-1``.
+
+        When rotational basis scaling is enabled, the values for strain rows
+        ``0:3`` are divided by the supplied link lengths. Passing lengths into
+        this helper keeps them dynamic JAX parameters: :meth:`with_params`
+        recomputes these compact values whenever a user updates link lengths.
+
+        Args:
+            segment_lengths: Current physical link lengths, shape
+                ``(self.num_segments,)``.
+
+        Returns:
+            Tuple ``(basis_rows, B_Z1_values, B_Z2_values)``. ``basis_rows`` has
+            shape ``(self.num_segments, self.max_dof)`` and maps active columns
+            to strain rows. Each values array has shape
+            ``(self.num_segments, self.max_num_integration_points - 1,
+            self.max_dof)`` and contains the corresponding scalar basis value
+            at a Magnus quadrature point.
+        """
         active = self.basis_active_params.astype(jnp.int32)
         orders = self.basis_order_params.astype(jnp.int32)
         standard_widths = active * (orders + 1)
@@ -1658,33 +1696,74 @@ class GVS(SoftRobot):
         xi_ref_Z1: Array,
         xi_ref_Z2: Array,
     ) -> tuple[Array, Array, Array, Array, Array, Array]:
-        """Evaluate one cell from compact one-row-per-column basis operands."""
-        valid = basis_rows >= 0
-        safe_rows = jnp.maximum(basis_rows, 0)
-        values_Z1 = jnp.where(valid, B_Z1_values, 0.0)
-        values_Z2 = jnp.where(valid, B_Z2_values, 0.0)
-        row_indicators = (
-            jax.nn.one_hot(safe_rows, 6, dtype=q_i.dtype).T * valid[None, :]
+        """Compute local ``J`` and ``Jdot`` terms from compact link bases.
+
+        This is algebraically equivalent to
+        :meth:`_magnus_jacobian_time_derivative_step_terms`, but exploits the
+        GVS basis structure that every active generalized-coordinate column has
+        a value in exactly one of the six strain rows. ``basis_rows`` and the
+        two value vectors reconstruct only the row selectors needed by the
+        second-order Magnus expansion; no dense link-basis matrices are formed.
+
+        The returned terms support the same local product-rule recurrence for
+        propagating a body Jacobian and its time derivative. Rotational basis
+        values are already length-scaled when that model option is enabled.
+
+        Args:
+            length: Physical length of the current link. Shape ``()``.
+            H: Normalized full or partial cell width. Shape ``()``.
+            q_i: Padded link generalized coordinates, shape
+                ``(self.max_dof,)``.
+            qd_i: Padded link generalized velocities, shape
+                ``(self.max_dof,)``.
+            basis_rows: Strain-row index for each compact basis column, shape
+                ``(self.max_dof,)``. Padding columns are marked with ``-1``.
+            B_Z1_values: Scalar basis value for each column at the first Magnus
+                quadrature point, shape ``(self.max_dof,)``.
+            B_Z2_values: Scalar basis value for each column at the second Magnus
+                quadrature point, shape ``(self.max_dof,)``.
+            xi_ref_Z1: Reference strain at the first quadrature point, shape
+                ``(6,)``.
+            xi_ref_Z2: Reference strain at the second quadrature point, shape
+                ``(6,)``.
+
+        Returns:
+            Tuple ``(g_step, T_step, Td_step, Ad_step_inv, B_Magnus,
+            B_Magnus_dot)``. ``g_step`` has shape ``(4, 4)``; ``T_step``,
+            ``Td_step``, and ``Ad_step_inv`` have shape ``(6, 6)``;
+            ``B_Magnus`` and ``B_Magnus_dot`` have shape
+            ``(6, self.max_dof)``.
+        """
+        # The compact arrays retain ``self.max_dof`` columns for fixed JAX
+        # shapes. A row of -1 marks a padding column with no strain-basis entry.
+        active_basis_columns = basis_rows >= 0
+        safe_basis_rows = jnp.maximum(basis_rows, 0)
+        masked_B_Z1_values = jnp.where(active_basis_columns, B_Z1_values, 0.0)
+        masked_B_Z2_values = jnp.where(active_basis_columns, B_Z2_values, 0.0)
+        basis_row_selectors = (
+            jax.nn.one_hot(safe_basis_rows, 6, dtype=q_i.dtype).T
+            * active_basis_columns[None, :]
         )
 
-        xi_Z1 = row_indicators @ (values_Z1 * q_i) + xi_ref_Z1
-        xi_Z2 = row_indicators @ (values_Z2 * q_i) + xi_ref_Z2
-        xid_Z1 = row_indicators @ (values_Z1 * qd_i)
-        xid_Z2 = row_indicators @ (values_Z2 * qd_i)
+        xi_Z1 = basis_row_selectors @ (masked_B_Z1_values * q_i) + xi_ref_Z1
+        xi_Z2 = basis_row_selectors @ (masked_B_Z2_values * q_i) + xi_ref_Z2
+        xid_Z1 = basis_row_selectors @ (masked_B_Z1_values * qd_i)
+        xid_Z2 = basis_row_selectors @ (masked_B_Z2_values * qd_i)
 
         ad_xi_Z1 = se3.small_adjoint(xi_Z1)
         ad_xi_Z2 = se3.small_adjoint(xi_Z2)
         comm_coeff = jnp.sqrt(3) * (length**2) * H**2 / 12
         Magnus = length * (H / 2) * (xi_Z1 + xi_Z2) + comm_coeff * (ad_xi_Z1 @ xi_Z2)
-        B_Magnus = length * (H / 2) * row_indicators * (values_Z1 + values_Z2)[
-            None, :
-        ] + comm_coeff * (
-            ad_xi_Z1[:, safe_rows] * values_Z2[None, :]
-            - ad_xi_Z2[:, safe_rows] * values_Z1[None, :]
+        B_Magnus = length * (H / 2) * basis_row_selectors * (
+            masked_B_Z1_values + masked_B_Z2_values
+        )[None, :] + comm_coeff * (
+            ad_xi_Z1[:, safe_basis_rows] * masked_B_Z2_values[None, :]
+            - ad_xi_Z2[:, safe_basis_rows] * masked_B_Z1_values[None, :]
         )
         B_Magnus_dot = comm_coeff * (
-            se3.small_adjoint(xid_Z1)[:, safe_rows] * values_Z2[None, :]
-            - se3.small_adjoint(xid_Z2)[:, safe_rows] * values_Z1[None, :]
+            se3.small_adjoint(xid_Z1)[:, safe_basis_rows] * masked_B_Z2_values[None, :]
+            - se3.small_adjoint(xid_Z2)[:, safe_basis_rows]
+            * masked_B_Z1_values[None, :]
         )
         Magnusd = B_Magnus @ qd_i
 
--- a/src/soromox/systems/gvs/core.py
+++ b/src/soromox/systems/gvs/core.py
@@ -65,6 +65,25 @@ from soromox.utils.numerics import safe_divide, safe_norm, safe_normalize
 
 __all__ = ["GVS"]
 
+_DEFAULT_NUM_DYNAMICS_PREFIX_BUCKETS = 4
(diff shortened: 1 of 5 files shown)


src/soromox/systems/gvs/core.py
```diff
@@ -83,6 +102,8 @@ class GVS(SoftRobot):
         max_num_integration_points: Maximum number of integration points per link (= max_num_gauss_points + 2).
         num_dofs: Total number of DOFs for the robot (sum of joint + link DOFs).
         num_padded_dofs: Theoretical maximum number of DOFs (num_segments * 2 * max_dof).
+        num_dynamics_prefix_buckets: Maximum number of active-DOF prefix
+            widths used by the dynamics reductions.
```
mstoelzle: Please describe in more detail. The user wouldn't know what we are referring here to

src/soromox/systems/gvs/core.py
```diff
@@ -742,6 +787,44 @@ def xi_reffn(x: Array) -> Array:
...
+        active_columns = (columns >= offsets[:, :, None]) & (
+            columns < (offsets + widths)[:, :, None]
+        )
+        rows = jnp.max(jnp.where(active_columns, row_numbers, -1), axis=1)
+
+        B_Z1 = self.B_Z1
+        B_Z2 = self.B_Z2
+        if self.scale_rotational_basis_by_length:
+            scales = segment_lengths[:, None, None, None]
+            B_Z1 = B_Z1.at[:, :, :3].divide(scales)
```
mstoelzle, line 859: Please double check if the segment do not need to be static parameters if the scaling of the strain basis is activated. If this is not the case, please double check if we are correctly updating the strain bases when the user updates the robot length parameters. Also, check if this aspect is covered by tests.

src/soromox/systems/gvs/core.py
```diff
@@ -1544,6 +1646,56 @@ def _magnus_jacobian_time_derivative_step_terms(
...
+        H: Array,
+        q_i: Array,
+        qd_i: Array,
+        basis_rows: Array,
+        B_Z1_values: Array,
+        B_Z2_values: Array,
+        xi_ref_Z1: Array,
+        xi_ref_Z2: Array,
+    ) -> tuple[Array, Array, Array, Array, Array, Array]:
+        """Evaluate one cell from compact one-row-per-column basis operands."""
```
mstoelzle: Please significantly expand the docstring and align it with the other google style docstrings.

src/soromox/systems/gvs/core.py
```diff
@@ -1544,6 +1646,56 @@ def _magnus_jacobian_time_derivative_step_terms(
...
+        valid = basis_rows >= 0
+        safe_rows = jnp.maximum(basis_rows, 0)
+        values_Z1 = jnp.where(valid, B_Z1_values, 0.0)
```
mstoelzle: This variable name is not very descriptive

src/soromox/systems/gvs/core.py
```diff
@@ -1544,6 +1646,56 @@ def _magnus_jacobian_time_derivative_step_terms(
...
+        q_i: Array,
+        qd_i: Array,
+        basis_rows: Array,
+        B_Z1_values: Array,
+        B_Z2_values: Array,
+        xi_ref_Z1: Array,
+        xi_ref_Z2: Array,
+    ) -> tuple[Array, Array, Array, Array, Array, Array]:
+        """Evaluate one cell from compact one-row-per-column basis operands."""
+        valid = basis_rows >= 0
```
mstoelzle: Please explain in a code comment why some rows would be invalid.

src/soromox/systems/gvs/core.py
mstoelzle: Expanded the public parameter documentation in 125ec6c. It now explains the causal DOF prefix, how ordered segments are bucketed, the runtime-versus-compilation tradeoff, the K=4 default, and that K=1 uses one full-width reduction branch.

src/soromox/systems/gvs/core.py
mstoelzle: Confirmed in 125ec6c that segment lengths should remain dynamic array leaves. with_params rebuilds the length-dependent compact Z1/Z2 operands from the updated lengths, so no static-parameter recompile is needed. I added test_scaled_compact_basis_tracks_dynamic_segment_length_updates, which checks the exact inverse-length update for rotational columns, unchanged translational columns, eager/JIT equality, and a single JIT trace across two different length vectors.

src/soromox/systems/gvs/core.py
mstoelzle: Expanded this in 125ec6c to a full Google-style docstring covering the compact representation, its equivalence to the dense helper, length-scaling expectations, every argument and shape, and all returned terms.

src/soromox/systems/gvs/core.py
mstoelzle: Renamed the local operands in 125ec6c to active_basis_columns, safe_basis_rows, masked_B_Z1_values, masked_B_Z2_values, and basis_row_selectors so their roles are explicit.

src/soromox/systems/gvs/core.py
mstoelzle: Added a code comment in 125ec6c explaining that fixed JAX shapes retain max_dof columns and row -1 marks a padding column that has no strain-basis entry.

vdperfetta01 (commented):
All good to me :)

mohammedtarnini (approved):
Amazing improvement! and codes look consistent.

Johannap1 (approved):

PR #177, opened 2026-08-27
Add Warp dynamics backends for GVS, PCS, and PlanarPCS

mstoelzle:
## Summary

This PR adds production Warp dynamics backends for `GVS`, `PCS`, and
`PlanarPCS`, while retaining JAX/XLA as the reference and differentiable
implementation. It also introduces the public `soromox.systems.execution`
package and a reusable Warp-native mechanics interface for future integrations,
including a separate Newton Cosserat-rod package.

The implementation is the result of shape-generic GVS experiments, profiling
with Nsight Systems and Nsight Compute, and subsequent specialization of the
successful ideas to the requested five-Gauss-point PlanarPCS and PCS setting.
Investigation-only benchmark sources are intentionally not tracked in this PR.

User-facing behavior is documented in the new
[`Execution Backends`](docs/user-guide/execution-backends.md) guide and is
introduced naturally in the Quick Start. The API reference mirrors the public
package under [`docs/api/execution`](docs/api/execution/index.md), including a
dedicated [`docs/api/execution/warp`](docs/api/execution/warp/index.md)
hierarchy for the GVS and PCS integration APIs. Highlighted notes make clear
that ordinary SoRoMoX users should select a backend on a system and use its
normal methods rather than launch low-level kernels directly.

## User-facing execution policy

Supported systems expose a static `backend` setting:

- `"auto"`: Warp on GPU, JAX/XLA on CPU;
- `"warp"`: explicitly request Warp;
- `"jax"`: explicitly request the reference JAX implementation.

GPU execution has no model-order, batch-size, or GPU-model crossover heuristic.
When Warp is selected on a GPU, scalar and batched calls use Warp. PlanarPCS and
PCS use portable defaults of 128 and 192 threads per block. Advanced tuning is
grouped in the typed `PCSBackendParams` object instead of adding
backend-specific constructor keywords:

```python
from soromox.systems import PCSBackendParams

robot = PCS.from_links(
    links,
    backend="auto",
    backend_params=PCSBackendParams(warp_block_dim=256),
)
```

The backend currently affects:

- `dynamics_terms(q, qd)`;
- `forward_dynamics(t, y, ..., backend=...)`;
- `rollout_to(...)` and `rollout_closed_loop_to(...)` through forward dynamics.

It does not change direct calls to `inertia_matrix`, `coriolis_matrix`,
`gravitational_force`, kinematics, Jacobians, energy functions, constitutive
forces, actuation, or rendering. Those remain JAX operations.

Both `dynamics_terms` and `forward_dynamics` accept a per-call override. The
model-level setting is used when the override is omitted. `dynamics_terms`
accepts scalar states `(num_dofs,)` and batches `(batch_size, num_dofs)`.
A direct `jax.vmap(model.dynamics_terms)(q, qd)` is intercepted by a custom
batching rule and becomes one batch-shaped Warp pipeline rather than
`batch_size` one-environment launches.

Warp is currently a forward-only numerical implementation. Direct forward- or
reverse-mode differentiation of `model.dynamics_terms` and complete
`model.forward_dynamics` calls is routed automatically to the existing JAX
assembly. The custom-JVP boundary is therefore an execution-routing rule, not a
new analytical derivative rule.

## Execution package structure

`soromox.systems.execution` owns backend types and parameters, neutral model
protocols, dispatch, batching/autodiff routing, and optional Warp loading.
Concrete system modules depend on this layer; the execution layer does not
import concrete system classes, avoiding circular dependencies.

The refactor also centralizes complete forward-dynamics assembly in a shared
compiled implementation. Public `GVS`, `PCS`, and `PlanarPCS`
`forward_dynamics` methods remain small, expose the same per-call backend
override, and use an identical force convention and inertia solve.

Importing SoRoMoX or `soromox.systems.execution` does not import `warp`.
Warp-specific symbols are loaded lazily, so `warp-lang` remains optional for
normal JAX-only use. Warp 1.16 is the minimum because its JAX FFI support added
the CPU compatibility required by the GVS execution path. Warp is included in
the test and docs dependency groups so the relevant CI jobs exercise and render
the interface.

Common Warp code now contains only mechanics shared across families:

- SE(3) and spatial-algebra functions;
- storage/view helpers;
- general-joint kernels and cooperative joint kernels;
- a family-neutral joint-term orchestration function accepting arrays and maps
  rather than GVS operand objects.

The joint code therefore lives under `execution/warp/common`, ready for PCS
joint extensions without creating a dependency on GVS. Family-specific cell,
chain, operand, and executor code remains under `warp/gvs` or `warp/pcs`.

`PlanarHSA` continues to use the JAX backend. Its inherited, unavailable
PlanarPCS-only fields are initialized in one explicitly named helper so the
Equinox PyTree remains complete without interleaving placeholder fields with
HSA parameter construction.

All new public and private execution functions have Google-style docstrings.
Public operand, workspace, kernel, and launch contracts document shapes,
dtypes, mutation, buffer ownership, launch dimensions, stream handling, and
return values.

## Reusable Warp-native integration API

Two public family namespaces expose reusable mechanics:

### `soromox.systems.execution.warp.gvs`

- `GVSOperandSource`, `GVSOperands`, and `GVSPipelineShapes`;
- joint, cooperative-joint, cell, and persistent-chain kernels;
- allocation-free launch functions for each stage.

### `soromox.systems.execution.warp.pcs`

- `PCSOperandSource`, `PCSOperands`, and `PCSPipelineShapes`;
- dedicated planar and spatial local-operator kernels;
- dedicated planar and spatial persistent-chain kernels;
- matching allocation-free launch functions.

The PCS surface is symmetrical with GVS. An external Newton adapter can create
zero-copy Warp views over persistent model/state storage, allocate workspace
once from the shape contract, and launch the mechanics without instantiating a
SoRoMoX system or invoking JAX.

Public launchers operate only on caller-owned `warp.array` inputs and outputs,
perform no output allocation, expose launch dimensions and streams, and can be
captured by a surrounding Warp-native CUDA graph. Production SoRoMoX executors
use the same public contracts, preventing the reusable surface from drifting
away from the system-class implementation.

## GVS implementation

The GVS pipeline supports shape-generic chains, spatially varying strain bases,
general joint families, arbitrary fixed topology, and the true GVS recurrence.
Generated Warp source is independent of batch size, segment count, basis order,
quadrature count, and joint family:

1. runtime-shaped general-joint Lie terms, with a cooperative variant where it
   helps;
2. runtime-shaped link-cell Lie terms for spatially varying GVS strain fields;
3. a persistent block per environment that propagates segment states,
   assembles quadrature contributions, and writes only final `(B, C @ qd, G)`.

This retains recurrence state on-chip where practical, removes large
intermediate global-memory traffic, exposes environments as independent GPU
work, and prevents source/compile size from growing with links or basis order.

Alternatives evaluated along the way include monolithic blocks, one block per
segment, finer kernel splitting, more aggressive cooperative reductions,
topology-specialized generated source, fused inertia solves, graph-wrapped
rollouts, and several block geometries. Many lost because the ordered chain
recurrence limits useful parallelism: extra kernels add launch/global-memory
cost, while extra cooperation adds synchronization and register pressure.
Topology specialization improved isolated kernels but reintroduced the compile
scaling this design is intended to avoid.

## Five-point PlanarPCS and PCS implementation

The PCS families exploit their constant-per-segment strain and regular chain
more aggressively than general GVS:

1. a local-operator kernel evaluates segment transforms, tangent terms,
   velocities, and five-point quadrature-local quantities in parallel;
2. a persistent chain kernel performs ordered propagation and accumulates
   `(B, C @ qd, G)` without materializing every recurrence state globally.

PlanarPCS uses dedicated compact planar algebra; PCS uses dedicated spatial
algebra. They share execution contracts and genuinely common low-level helpers
without paying for the full GVS generality.

## Why the implementation is faster

The gains are primarily reductions in total work, launch overhead, and memory
traffic around the serial chain dependency:

- batch-shaped launches expose independent environments;
- persistent blocks retain recurrence state across segments;
- local Lie/quadrature work is fused enough to avoid intermediate launches and
  global traffic without excessive synchronization;
- only final assembled terms cross the Warp/JAX boundary;
- runtime topology operands keep compilation essentially independent of model
  length and batch size;
- reusable buffers make the raw launchers CUDA-graph compatible.

The principal remaining bottleneck is the irreducible ordered propagation along
each kinematic chain. More environments improve throughput, but more threads do
not remove that dependency.

## Benchmark tooling

The shared system-method benchmark and Section IVb rollout generator accept a
single `--backend auto|jax|warp` option. It is propagated through both ordinary
and Gauss-point factories for GVS, PCS, and PlanarPCS, including the GVS
basis-order sweep. JAX-only systems continue to construct normally during a
heterogeneous sweep.

JSON/CSV rows distinguish `device` (`cpu` or `gpu`) from the requested
`backend`, the `resolved_backend`, and `backend_applies`. The last field is
false for methods such as kinematics or direct inertia-matrix queries that do
not use dynamics backend dispatch. Controlled JAX/Warp comparisons should run
each backend in a separate process to isolate compilation and allocator state.

## GPU benchmarks

These results were rerun on the final production code after the execution-layer
refactor. Values are medians of 50 synchronized FP64 calls on an RTX 5090 after
20 warmups. JAX and Warp samples were alternated to reduce ordering and clock
bias. Each model has four segments; GVS order 1 uses five quadrature points,
GVS order 5 uses nine, and both PCS systems use five. Batch 1024 is included for
completeness but was not the main tuning target.

### GVS, basis order 1

| Operation | Batch | JAX (ms) | Warp (ms) | Speedup |
|---|---:|---:|---:|---:|
| terms | 1 | 1.086 | 0.506 | 2.15x |
| terms | 64 | 1.670 | 0.622 | 2.69x |
| terms | 256 | 2.224 | 0.735 | 3.02x |
| terms | 1024 | 5.291 | 1.381 | 3.83x |
| forward | 1 | 1.174 | 0.620 | 1.89x |
| forward | 64 | 1.761 | 0.738 | 2.39x |
| forward | 256 | 2.482 | 1.034 | 2.40x |
| forward | 1024 | 5.729 | 1.821 | 3.15x |

### GVS, basis order 5 / nine quadrature points

| Operation | Batch | JAX (ms) | Warp (ms) | Speedup |
|---|---:|---:|---:|---:|
| terms | 1 | 1.329 | 1.252 | 1.06x |
| terms | 64 | 2.566 | 1.515 | 1.69x |
| terms | 256 | 4.381 | 1.973 | 2.22x |
| terms | 1024 | 15.604 | 6.899 | 2.26x |
| forward | 1 | 1.755 | 1.629 | 1.08x |
| forward | 64 | 3.601 | 2.509 | 1.44x |
| forward | 256 | 5.752 | 3.325 | 1.73x |
| forward | 1024 | 19.993 | 11.540 | 1.73x |

### PlanarPCS, four segments, five Gauss points

| Operation | Batch | JAX (ms) | Warp (ms) | Speedup |
|---|---:|---:|---:|---:|
| terms | 1 | 0.340 | 0.372 | 0.92x |
| terms | 64 | 0.269 | 0.263 | 1.02x |
| terms | 256 | 0.291 | 0.284 | 1.02x |
| terms | 1024 | 0.749 | 0.554 | 1.35x |
| forward | 1 | 0.389 | 0.241 | 1.61x |
| forward | 64 | 0.455 | 0.307 | 1.48x |
| forward | 256 | 0.335 | 0.151 | 2.21x |
| forward | 1024 | 0.769 | 0.374 | 2.06x |

### PCS, four segments, five Gauss points

| Operation | Batch | JAX (ms) | Warp (ms) | Speedup |
|---|---:|---:|---:|---:|
| terms | 1 | 0.260 | 0.367 | 0.71x |
| terms | 64 | 0.476 | 0.495 | 0.96x |
| terms | 256 | 0.663 | 0.518 | 1.28x |
| terms | 1024 | 1.507 | 0.843 | 1.79x |
| forward | 1 | 0.421 | 0.301 | 1.40x |
| forward | 64 | 0.572 | 0.433 | 1.32x |
| forward | 256 | 0.755 | 0.456 | 1.66x |
| forward | 1024 | 1.496 | 0.696 | 2.15x |

## CPU benchmarks

The same final-code harness was run for one CPU environment. Warp does not
provide a useful CPU advantage and substantially regresses GVS, which is why
`backend="auto"` selects JAX on CPU.

| Model | Operation | JAX (ms) | Warp (ms) | Warp relative to JAX |
|---|---|---:|---:|---:|
| GVS order 1 / 5 points | terms | 0.232 | 0.415 | 1.79x slower |
| GVS order 1 / 5 points | forward | 0.382 | 0.645 | 1.69x slower |
| GVS order 5 / 9 points | terms | 0.587 | 1.445 | 2.46x slower |
| GVS order 5 / 9 points | forward | 0.638 | 1.533 | 2.40x slower |
| PlanarPCS / 5 points | terms | 0.037 | 0.039 | 1.04x slower |
| PlanarPCS / 5 points | forward | 0.034 | 0.035 | 1.02x slower |
| PCS / 5 points | terms | 0.075 | 0.077 | 1.03x slower |
| PCS / 5 points | forward | 0.072 | 0.071 | 1.01x faster |

## Compilation scaling

Runtime-shaped Warp source prevents compile time from growing with chain
length. Representative GVS measurements were:

| Model | Cached Warp load | Cold JAX compile | Cold Warp process |
|---|---:|---:|---:|
| 4 segments | 0.055 s | 1.02 s | 0.901 s |
| 16 segments | 0.063 s | 1.08 s | 0.942 s |

Five-point Warp estimates were approximately 0.05–0.19 s, compared with
0.57–0.92 s for PlanarPCS JAX and 0.92–1.55 s for PCS JAX depending on shape.
All observed Warp compilation paths remain below the 10 s target.

## CUDA graph behavior

SoRoMoX does not embed manual `wp.capture_*` ownership in a system. JAX/XLA
owns command-buffer/CUDA-graph capture around graph-compatible Warp FFI calls.
Nsight traces of production paths showed ten warmed `GVS.forward_dynamics`
queries producing ten `cuGraphLaunch` calls. Standard Diffrax rollouts capture
and replay stage fragments rather than one monolithic rollout graph.

The public raw launchers allocate nothing and can be captured by a surrounding
Warp/Newton integrator; graph creation, replay, and lifetime remain that
integrator's responsibility.

## Refactor performance guard

The final package/API refactor was compared directly with the previously
committed Warp implementation (`1db8fd1`) using dependent 100-call steady-state
GPU loops. Values below are final relative changes; negative is faster:

| System | Batch | Terms | Forward dynamics |
|---|---:|---:|---:|
| GVS | 64 | -0.36% | -0.92% |
| GVS | 256 | -0.02% | +0.76% |
| PCS | 64 | -3.65% | -0.88% |
| PCS | 256 | -1.30% | -0.08% |
| PlanarPCS | 64 | -3.49% | +1.33% |
| PlanarPCS | 256 | -1.92% | +1.54% |

The only upward differences are at most 1.54%, within normal run-to-run noise;
the structural refactor introduced no material performance regression.

## Validation and test structure

Tests mirror the package under `tests/systems/execution` and
`tests/systems/execution/warp`. Generic dispatch/transform behavior is tested
once in the execution suite; real family-specific JAX/Warp equivalence lives in
the Warp subtree rather than being duplicated in the system-class tests.

Coverage includes:

- optional/lazy Warp imports and loader failures;
- backend and typed parameter validation;
- neutral dependency direction and family isolation;
- operand-view and workspace contracts;
- public symbol discovery, launcher signatures, and documentation;
- scalar calls, explicit batches, and direct public `vmap`;
- proof that mapped scalar calls execute one runtime batch;
- direct gradients of `dynamics_terms` and gradients of complete
  `forward_dynamics` calls;
- numerical JAX/Warp equivalence for `dynamics_terms` and `forward_dynamics` on
  GVS, PCS, and PlanarPCS.

Latest local results:

- execution suite: **62 passed, 3 skipped**;
- real-GPU integration/equivalence suite: **17 passed**;
- full systems suite: **350 passed, 8 skipped**;
- Ruff, formatting, `git diff --check`, and the complete docs build: clean.

## Scope and limitations

- Warp is optional and current kernels require FP64.
- Warp provides forward execution; differentiation routes to JAX.
- PlanarPCS/PCS specialization targets five Gauss points.
- Ordinary users should use system methods; the raw launch API is for expert
  integrators.
- CUDA-graph ownership belongs to the surrounding integrator.
- Newton integration is deliberately not included and SoRoMoX does not acquire
  a Newton dependency. This PR establishes the reusable mechanics boundary for
  a separate adapter package.

## Commit structure relative to `main`

The branch is rebased onto current `origin/main`. Its commits cover the
investigation history, production GVS integration and optimization, unified
dispatch, five-point PCS specialization, removal of tracked benchmark
artifacts, execution-layer refactoring, and publication/hardening of the GVS
and PCS integration APIs and documentation. Commits already present on `main`
are not described as part of this PR.


src/soromox/systems/pcs/pcs.py
```diff
@@ -3293,24 +3460,66 @@ def mass_action(value: Array, diagonal: Array = mass_diagonal) -> Array:
 
         return inertia, coriolis_qd, gravity_force
 
-    @eqx.filter_jit
     def forward_dynamics(
         self, t: Array, y: Array, actuation_args: tuple | None = None
     ) -> Array:
+        """Compute the first-order state derivative for spatial PCS dynamics.
```
mstoelzle: This is the first-order time derivative, no? This is not a derivative with respect to `y`

src/soromox/systems/pcs/pcs.py
```diff
@@ -3293,24 +3460,66 @@ def mass_action(value: Array, diagonal: Array = mass_diagonal) -> Array:
...
+
+        The model's configured backend selects the primal implementation of
+        :meth:`dynamics_terms`. JAX transformations remain valid because the
+        terms dispatcher substitutes the differentiable JAX assembly whenever
```
mstoelzle: I think this note is putting too much effort on the validity of JAX transformations. The user should assume that this is the case anyway.

src/soromox/systems/pcs/pcs.py
```diff
@@ -3293,24 +3460,66 @@ def mass_action(value: Array, diagonal: Array = mass_diagonal) -> Array:
...
+        return evaluate_forward_dynamics(self, t, y, actuation_args)
+
+    @eqx.filter_jit
+    def _evaluate_forward_dynamics(
+        self,
+        t: Array,
+        y: Array,
+        actuation_args: tuple | None,
+        *,
+        backend: ExecutionBackend | None,
```
mstoelzle: Why not add the `backend` parameter to the `forward_dynamics` and eliminate `_evaluate_forward_dynamics`?

src/soromox/systems/pcs/planar_hsa.py
```diff
@@ -198,6 +201,14 @@ def __init__(
         self.young_stiffness_operator = None
         self.shear_stiffness_operator = None
         self.material_damping_operator = None
+        self.active_strain_indices = None
```
mstoelzle: Can we find a smarter way of initializing `PlanarPCS` attributes in `PlanarHSA` that are not actually used?

src/soromox/systems/pcs/planar_pcs.py
```diff
@@ -285,6 +307,9 @@ def __init__(
                 model.
             passive_elements: Optional passive element or tuple of passive
                 elements. Pass ``None`` or an empty tuple to disable them.
+            backend: Preferred execution backend for accelerated methods.
+            warp_block_dim: CUDA threads per block for the persistent Warp
```
mstoelzle: Should this be part of `ExecutionBackend` or a (typed)`backend_params` parameter? I am concerned that if we add more parameters and backends, things will get very messy.

src/soromox/systems/pcs/planar_pcs.py
```diff
@@ -678,6 +709,60 @@ def _precomputed_matrices(self) -> tuple[Array, Array, Array, Array, Array]:
         D_active = self.B_xi.T @ D_full @ self.B_xi
         return M_segments, K_full, K_active, D_full, D_active
 
+    def _dynamics_runtime_arrays(
+        self, M_segments: Array
+    ) -> tuple[Array, Array, Array, Array, Array, Array, Array, Array]:
+        """Build compact, state-independent operands for dynamics backends."""
```
mstoelzle: As discussed, make the docstrings more extensive according to the format and style established in the package.

docs/user-guide/execution-backends.md
```diff
@@ -0,0 +1,199 @@
+# Execution Backends
+
+SoRoMoX can use different numerical implementations for selected system
+operations while preserving the same model API. Backend selection currently
+applies to continuum-system dynamics, but the setting is intentionally named
```
mstoelzle: This is too much "developer talk". Make the docs more user-centric.

docs/user-guide/execution-backends.md
```diff
@@ -0,0 +1,199 @@
...
+            material_damping_coefficient=318.0,
+            reference_strain=[0.0, 1.0, 0.0],
+        )
+    ],
+    backend="auto",
+)
+```
+
+The system methods do not otherwise change. The configured backend is used by
+direct dynamics queries and by simulations that call forward dynamics.
```
mstoelzle: Be more specific and precise here. Namely, it is (currently) NOT used when calling the individual functions, such as `inertial_matrix`, `coriolis_matrix`, etc.

pyproject.toml
```diff
@@ -245,6 +249,7 @@ all = [
     "tox",
     "trimesh",
     "viser",
+    "warp-lang>=1.16.0",
```
mstoelzle, line 256: Why are you pinning a specific version? Is this because were using specific features / interfaces introduced in v1.16?

src/soromox/systems/execution/warp/common/se3.py
```diff
@@ -0,0 +1,362 @@
...
+
+
+Vec6d = wp.types.vector(length=6, dtype=wp.float64)
+SPATIAL_DIM = 6
+
+wp.set_module_options({"enable_backward": False})
+
+
+@wp.func
+def _ad_action(left: Vec6d, right: Vec6d) -> Vec6d:
```
mstoelzle, line 21: I am missing docstrings for all these functions. Again, please apply the format and style of the existing docstrings in the package (including a list of args and return values).

src/soromox/systems/execution/warp/common/storage.py
```diff
@@ -0,0 +1,67 @@
+# ruff: noqa: I001, UP018
```
mstoelzle, line 1: Again, docstrings are missing for all these functions and kernels.

src/soromox/systems/execution/warp/gvs/cell.py
```diff
@@ -0,0 +1,333 @@
...
+        row = basis_rows[segment, column]
+        if row >= 0:
+            value[row] += basis_values[cell_item, column] * velocity
+        column += 1
+    return value
+
+
+@wp.kernel(enable_backward=False)
+def cell_terms_kernel(
+    q_link: wp.array2d(dtype=wp.float64),
```
mstoelzle: Missing docstrings everywhere.

src/soromox/systems/execution/warp/gvs/executor.py
```diff
@@ -0,0 +1,219 @@
...
+        Padded local values with shape
+        ``(batch_size, num_segments, max_dof)``. Padding entries are zero.
+    """
+
+    safe_indices = jnp.maximum(local_to_global, 0)
+    gathered = jnp.take(values, safe_indices, axis=1)
+    return gathered * (local_to_global >= 0)[None, :, :]
+
+
+def _joint_terms(
```
mstoelzle: Joints might later also be used for the `PCS` class. Can we generalize this already ahead of time a bit better?

src/soromox/systems/execution/warp/common/joints.py
mstoelzle, line 1: Same comment as before: Joints might be later also be used for `PCS` systems. Therefore, should this move to `warp/common`?

src/soromox/systems/pcs/pcs.py
```diff
@@ -705,6 +736,57 @@ def _precomputed_matrices(self) -> tuple[Array, Array, Array, Array, Array]:
         D_active = self.B_xi.T @ D_full @ self.B_xi
         return M_segments, K_full, K_active, D_full, D_active
 
+    def _dynamics_runtime_arrays(
+        self, M_segments: Array
+    ) -> tuple[Array, Array, Array, Array, Array, Array, Array, Array]:
+        """Build compact, state-independent operands for dynamics backends."""
```
mstoelzle: Please make the docstring more extensive and properly document all parameters and return values.

tests/systems/test_gvs.py
```diff
@@ -1824,6 +1839,328 @@ def test_dynamics_terms_match_public_matrices(
...
+    qd = jax.random.normal(key_qd, (3, robot.num_dofs), dtype=jnp.float64) * 0.04
+
+    expected = jax.vmap(robot.dynamics_terms)(q, qd)
+    actual = robot.dynamics_terms(q, qd, backend="jax")
+
+    for actual_term, expected_term in zip(actual, expected, strict=True):
+        assert_allclose(actual_term, expected_term, rtol=RTOL, atol=ATOL)
+
+
+def test_warp_dispatch_batches_primals_and_differentiates_with_jax(
```
mstoelzle: Should this test not go into warp test folder? Please carefully reflect about this for each test where its optimal placement is.

pyproject.toml
```diff
@@ -128,6 +128,10 @@ dependencies = [
 cuda13 = [
     "jax[cuda13]>=0.10.0; sys_platform == 'linux'",
     "torch>=2.13.0; sys_platform == 'linux'",
+    "warp-lang>=1.16.0; sys_platform == 'linux'",
+]
+warp = [
```
mstoelzle, line 135: Warp also needs to be added as test dependency otherwise the CI will fail I think.

src/soromox/systems/pcs/pcs.py
mstoelzle: Fixed in d7c6934. The docstring now explicitly calls this the first-order time derivative and returns `[qd, qdd]`; it no longer suggests differentiation with respect to `y`.

src/soromox/systems/pcs/pcs.py
mstoelzle: Fixed in d7c6934. The public `forward_dynamics` docstrings now describe user-visible behavior and arguments without explaining internal JAX-transform validity.

src/soromox/systems/pcs/pcs.py
mstoelzle: Implemented in d7c6934. `forward_dynamics` now exposes the optional per-call `backend` override, and the duplicated family `_evaluate_forward_dynamics` methods were removed in favor of one shared compiled execution implementation.

src/soromox/systems/pcs/planar_hsa.py
mstoelzle: Addressed in d7c6934. The inherited placeholders are centralized in `_initialize_unavailable_planar_pcs_fields`, which documents why the fields must exist for the Equinox PyTree while keeping them out of the HSA parameter setup.

src/soromox/systems/pcs/planar_pcs.py
mstoelzle: Implemented in d7c6934. PCS-specific tuning is now grouped in the typed `PCSBackendParams`; `warp_block_dim` is no longer a top-level system constructor argument.

src/soromox/systems/pcs/planar_pcs.py
mstoelzle: Addressed in d7c6934. `_dynamics_runtime_arrays` now documents every input and each returned array, including shapes and semantics, using the package's Google-style format.

docs/user-guide/execution-backends.md
mstoelzle: Addressed in d7c6934. The guide now opens with the normal user workflow, backend choices, supported methods, and examples; implementation details are confined to the separate API reference.

docs/user-guide/execution-backends.md
mstoelzle: Clarified in d7c6934. The guide now lists exactly which calls use backend dispatch and explicitly says direct `inertia_matrix`, `coriolis_matrix`, `gravitational_force`, kinematics, Jacobian, energy, constitutive-force, actuation, and rendering calls remain JAX operations.

pyproject.toml
mstoelzle: Documented in d7c6934. Warp 1.16 is the minimum because that release introduced the JAX FFI CPU support used by this backend; the dependency comment and PR description now state this rationale.

src/soromox/systems/execution/warp/common/se3.py
mstoelzle: Addressed in d7c6934. All private and public SE(3) execution helpers now have Google-style docstrings with argument and return-value documentation.

src/soromox/systems/execution/warp/common/storage.py
mstoelzle: Addressed in d7c6934. Storage helpers and kernels now document inputs, returned views, in-place mutation, dtypes, shapes, and buffer ownership.

src/soromox/systems/execution/warp/gvs/cell.py
mstoelzle: Addressed in d7c6934. The GVS cell functions and kernels now have full Google-style docstrings, and an AST audit verifies all private execution functions include `Args` and `Returns` sections.

src/soromox/systems/execution/warp/gvs/executor.py
mstoelzle: Implemented in d7c6934. Joint orchestration is now family-neutral in `execution/warp/common/joint_terms.py` and accepts runtime arrays/maps rather than `GVSOperands`; the GVS executor only adapts its operands into that shared contract.

src/soromox/systems/execution/warp/common/joints.py
mstoelzle: Implemented in d7c6934. Both joint kernel modules were moved from `warp/gvs` to `warp/common`, while the public GVS namespace lazily re-exports the supported symbols.

src/soromox/systems/pcs/pcs.py
mstoelzle: Addressed in d7c6934. The spatial PCS runtime-array helper now documents the input and all returned arrays, including their shapes and roles.

tests/systems/test_gvs.py
mstoelzle: Reorganized in d7c6934. Generic dispatch/transform tests live under `tests/systems/execution`; real GVS/PCS/PlanarPCS JAX-Warp equivalence, batching, and gradient tests live under `tests/systems/execution/warp`; system test files retain only system API/cache behavior.

pyproject.toml
mstoelzle: Fixed in d7c6934. `warp-lang>=1.16.0` is now included in the test and docs dependency groups, and `uv.lock` was regenerated.

vdperfetta01 (commented):
impressive additions, but it looks all good to me

vdperfetta01 (approved):

PR #181, opened 2026-08-28
Add fused Warp kinematics kernels

mstoelzle:
## Summary

This PR adds fused Warp forward-kinematics and inertial-frame Jacobian
execution for `PlanarPCS`, `PCS`, and `GVS`. It is intentionally stacked on
#177 and reuses that PR's backend selection, lazy optional-dependency loading,
FP64 policy, custom batching/autodiff boundaries, and public Warp-native
integration conventions.

The public continuum kinematics methods now support:

- one configuration and one curvilinear abscissa;
- one configuration and an abscissa batch;
- an environment batch and one shared abscissa;
- an environment batch and shared abscissae;
- an environment batch and per-environment abscissae; and
- direct or nested `jax.vmap` forms of the same layouts.

The implementation covers pose-only, inertial-Jacobian-only, and fused
pose-plus-Jacobian calls. Dynamics, force assembly, body-frame Jacobians, and
arc-length/time derivatives remain on their existing paths.

## Stack and merge order

This branch contains focused commits on top of the rebased #177 branch
`codex/pcs-warp-kernels`. The intended order is:

1. merge #177;
2. retarget this PR to `main`; and
3. merge this PR after the resulting base-only diff disappears.

PRs #179 and #180 are already present in the stack through `main`. The rebase
retains #179's shared rotational-basis normalization helper and expanded
six-family regression, and uses #180's `*_abscissa_batched` naming throughout.

## Public API and execution policy

The following methods accept keyword-only
`backend: ExecutionBackend | None = None` on `PlanarPCS`, `PCS`, and `GVS`:

- `forward_kinematics`;
- `forward_kinematics_abscissa_batched`;
- `jacobian_inertialframe`;
- `jacobian_inertialframe_abscissa_batched`;
- `forward_kinematics_and_jacobian_inertialframe`; and
- `forward_kinematics_and_jacobian_inertialframe_abscissa_batched`.

The un-suffixed methods retain the package-wide single-point contract: one
generalized-coordinate vector `q[D]` and one scalar abscissa `s`. The explicit
`*_abscissa_batched` methods accept one `q[D]` and abscissae `s[N]`. Environment,
pairwise, and Cartesian batches are expressed with `jax.vmap`, consistent with
the rest of the public system API.

Backend selection follows #177 exactly:

- `"auto"` selects Warp on a supported GPU and JAX on CPU;
- `"jax"` selects the differentiable reference implementation; and
- `"warp"` explicitly selects Warp and validates model support and FP64.

Warp remains a forward-only primal accelerator. `jvp`, `jacfwd`, `jacrev`, and
`grad` route through the established JAX implementations.

## Batching and JAX transformation behavior

All Warp calls normalize to environments `q[E,D]` and samples `s[E,N]`.
Single-point and abscissa-batched public outputs are sliced back to their
declared shapes. Shared samples are broadcast without changing the public
result layout.

Custom batching rules ensure that:

- `vmap` over abscissae uses each family's specialized JAX abscissa traversal;
- `vmap` over configurations reaches one environment-batched Warp executor;
- pairwise `vmap(forward_kinematics)(q_batch, s_batch)` reaches one canonical
  Warp execution; and
- non-pairwise and nested environment-by-abscissa `vmap` forms also collapse
  into one double-batched Warp execution rather than repeated batch-one
  launches.

The corresponding public forms are:

```python
# One configuration, multiple abscissae: N
jax.vmap(robot.forward_kinematics, in_axes=(None, 0))(q, s_ps)

# Multiple configurations, one shared scalar abscissa: E
jax.vmap(robot.forward_kinematics, in_axes=(0, None))(q_ts, s)

# Pairwise configurations and scalar abscissae: E
jax.vmap(robot.forward_kinematics)(q_ts, s_ts)

# Cartesian configurations by shared abscissae: E x N
jax.vmap(
    robot.forward_kinematics_abscissa_batched,
    in_axes=(0, None),
)(q_ts, s_ps)

# Per-environment abscissa batches: E x N
jax.vmap(robot.forward_kinematics_abscissa_batched)(q_ts, s_tps)
```

The same custom batching rules cover `jacobian_inertialframe` and the fused
pose-plus-Jacobian methods. In particular, an abscissa `vmap` of the scalar JAX
Jacobian is redirected to `_jacobian_inertialframe_abscissa_batched`, avoiding
the naive point-by-point JAX traversal.

Samples may be unsorted or duplicated and may include the base, tip, or exact
segment/cell boundaries. Existing boundary ownership is preserved.

## Warp-native integration API

The public `soromox.systems.execution.warp.pcs` and
`soromox.systems.execution.warp.gvs` namespaces gain kinematics operand and
shape contracts plus allocation-free launchers for:

- forward kinematics;
- inertial-frame Jacobians; and
- fused forward kinematics and inertial-frame Jacobians.

The launchers accept caller-owned Warp input, output, and scratch arrays. They
perform no JAX calls, host synchronization, allocation, or device transfer and
remain reusable from downstream Warp-native code and CUDA graph capture.
Planar and spatial PCS kernels are split into dedicated public modules so each
family's low-level math and launch surface remain discoverable.

## PlanarPCS and PCS implementation

Both PCS families exploit constant strain per segment:

1. build segment strain, pose, and—when requested—body-Jacobian boundary
   states once per environment;
2. classify arbitrary samples and evaluate their partial segment in parallel;
3. rotate body Jacobians into the inertial frame at the evaluated pose; and
4. use reduced pose-only and Jacobian-only launch paths to avoid unnecessary
   recurrence work.

PlanarPCS uses compact SE(2) algebra. PCS uses specialized SE(3) helpers and a
cooperative CUDA schedule for Jacobian propagation and sample evaluation.

## GVS implementation

GVS reuses the joint/link recurrence and runtime topology operands from #177.
It evaluates every supported strain-basis family, stores full cell-boundary
states once, and evaluates each sample's remaining partial cell in parallel.

The implementation supports mixed basis families, padded local DOFs, fixed and
general joints, length-normalized rotational bases, unsorted/duplicate
samples, and runtime segment and sample counts. Pose-only preparation omits
velocity-dependent local terms; Jacobian and fused calls share the reduced
kinematics recurrence.

## Performance methodology

The full matrix contains 324 records per device: `PlanarPCS`, `PCS`, and `GVS`;
2 and 8 segments; environments `E = 1, 16, 256`; samples
`N = 1, 8, 64`; pose, Jacobian, and fused calls; and JAX/Warp backends.
Compilation/first-call time is recorded separately from synchronized warmed
execution. The complete environment-`vmap` callable is JIT-compiled once before
warm timing, so the measurements include compiled dispatch and kernel execution
but exclude repeated Python transformation construction. Samples are shared
across environments for this matrix.

GPU measurements use an NVIDIA GeForce RTX 5090 with FP64 enabled, 9 repeated
measurements, 2 warmups, and 10 asynchronously enqueued calls per synchronized
timing. CPU measurements use an Intel Core Ultra 9 285K, 7 repeated
measurements, 2 warmups, and 5 calls per synchronized timing. The tables below
report `JAX time / Warp time`, so values above 1 mean Warp is faster.

The benchmark harness and raw artifacts are intentionally not committed. This
keeps the production PR focused, consistent with #177; the complete raw matrix
is available from the author during review.

## GPU results

Geometric-mean warmed speedup over all 18 `(segments, E, N)` shapes for each
system/operation:

| System | Pose | Jacobian | Fused pose + Jacobian |
|---|---:|---:|---:|
| PlanarPCS | 1.07x | 1.80x | 1.91x |
| PCS | 1.11x | 1.95x | 1.85x |
| GVS | 7.90x | 9.31x | 9.31x |

The balanced geometric mean across all 162 paired GPU cases is **2.79x**.

For the largest measured shape—8 segments, `E=256`, `N=64`—the synchronized
warmed medians are:

| System | Operation | JAX (ms) | Warp (ms) | Speedup |
|---|---|---:|---:|---:|
| PlanarPCS | pose | 0.173 | 0.103 | 1.68x |
| PlanarPCS | Jacobian | 0.711 | 0.400 | 1.78x |
| PlanarPCS | fused | 0.927 | 0.403 | 2.30x |
| PCS | pose | 0.235 | 0.145 | 1.62x |
| PCS | Jacobian | 1.596 | 0.337 | 4.73x |
| PCS | fused | 1.575 | 0.321 | 4.90x |
| GVS | pose | 2.778 | 0.455 | 6.10x |
| GVS | Jacobian | 7.706 | 0.608 | 12.67x |
| GVS | fused | 7.715 | 0.606 | 12.74x |

Small PCS/PlanarPCS pose-only calls remain launch-bound and can favor JAX;
the fused kernels become useful as segment, environment, sample, or Jacobian
work increases. GVS benefits consistently because the fused path removes
substantially more recurrence preparation and intermediate work.

## CPU results

The same geometric-mean summary on CPU is:

| System | Pose | Jacobian | Fused pose + Jacobian |
|---|---:|---:|---:|
| PlanarPCS | 0.35x | 1.29x | 1.25x |
| PCS | 0.54x | 1.35x | 1.20x |
| GVS | 0.67x | 1.74x | 1.77x |

The balanced geometric mean across all CPU cases is **1.00x**. Warp does not
win uniformly: for the largest eight-segment, `E=256`, `N=64` PCS case, pose,
Jacobian, and fused speedups are 0.57x, 0.90x, and 0.95x. The production policy
therefore remains unchanged: `auto` prefers JAX on CPU, while Warp CPU execution
is available explicitly for low-level reuse and correctness testing. The
kernels do not branch into a separate CPU algorithm; GPU behavior remains the
optimization priority.

The largest-shape CPU medians are:

| System | Operation | JAX (ms) | Warp (ms) | Speedup |
|---|---|---:|---:|---:|
| PlanarPCS | pose | 0.124 | 0.337 | 0.37x |
| PlanarPCS | Jacobian | 3.163 | 3.318 | 0.95x |
| PlanarPCS | fused | 3.315 | 3.517 | 0.94x |
| PCS | pose | 2.722 | 4.778 | 0.57x |
| PCS | Jacobian | 41.974 | 46.562 | 0.90x |
| PCS | fused | 41.648 | 44.064 | 0.95x |
| GVS | pose | 6.251 | 9.627 | 0.65x |
| GVS | Jacobian | 92.266 | 49.326 | 1.87x |
| GVS | fused | 96.879 | 50.268 | 1.93x |

## Validation

Completed after rebasing:

- CUDA execution/backend suite: **84 passed**;
- CPU execution/backend suite: **81 passed, 3 skipped**;
- broad CPU continuum/articulated-system regression suite: **431 passed**;
- renderer regression suite: **70 passed**;
- benchmark-tool regression suite: **27 passed**;
- direct dispatch instrumentation covers spatial, environment, pairwise,
  non-pairwise, explicit/per-environment abscissa, and nested Cartesian `vmap`
  forms for poses and inertial Jacobians;
- Warp-request derivative checks cover scalar and abscissa-batched pose,
  Jacobian, and fused methods for all three families;
- the Warp annotation architecture check passes with PR #177's bracket syntax;
- Ruff lint and formatting pass for every changed Python file; and
- the documentation build and `git diff --check` pass.


src/soromox/systems/execution/warp/gvs/kinematics_executor.py
```diff
@@ -0,0 +1,158 @@
...
+    q: Array,
+    sample_s: Array,
+    operation: KinematicsOperation,
+) -> KinematicsResult:
+    """Execute batched GVS poses and inertial-frame Jacobians in Warp.
+
+    Args:
+        operands: Runtime GVS model data and static dimensions.
+        q: Batched active configurations with shape ``(E, D)``.
+        sample_s: Per-environment abscissae with shape ``(E, N)``.
```
mstoelzle: Should this parameter not be renamed to just `s?`

docs/user-guide/execution-backends.md
```diff
@@ -63,9 +63,9 @@ pip install "soromox[warp]"
 
 | System | Warp support | Requirements |
 |---|---|---|
-| `GVS` | Dynamics on GPU; explicit Warp execution is also available on CPU | FP64 state arrays |
-| `PCS` | Dynamics on GPU | Exactly five Gauss points and FP64 state arrays |
-| `PlanarPCS` | Dynamics on GPU | Exactly five Gauss points and FP64 state arrays |
+| `GVS` | Kinematics and dynamics on GPU; explicit Warp execution is also available on CPU | FP64 arrays |
```
mstoelzle: Please change the order: `PlanarPCS`, `PCS`, and `GVS`

src/soromox/systems/execution/warp/gvs/kinematics.py
```diff
@@ -0,0 +1,1574 @@
...
+        if slot == 0:
+            value = wp.float64(1.0)
+        else:
+            mode = (slot + 1) // 2
+            angle = wp.float64(2.0) * wp.float64(wp.pi) * wp.float64(mode) * x
+            if slot % 2 == 1:
+                value = wp.cos(angle)
+            else:
+                value = wp.sin(angle)
+    elif basis_type == 4 or basis_type == 5:
```
mstoelzle: This is problematic: What if the order of the strain basis is higher than 5?

src/soromox/systems/execution/warp/gvs/kinematics.py
```diff
@@ -0,0 +1,1574 @@
...
+                k += 1
+            jacobians[environment, sample, row, column] = output
+            row += 1
+        column += 1
+
+
+@wp.kernel(enable_backward=False)
+def gvs_jacobian_samples_kernel(
+    q: wp.array2d[wp.float64],
+    sample_s: wp.array2d[wp.float64],
```
mstoelzle: Should we rename to `s`?

src/soromox/systems/execution/warp/gvs/kinematics_executor.py
mstoelzle: Addressed in 16f83f4. I standardized the complete new kinematics execution path on `s`: the lazy loader, canonical and family executors, all PlanarPCS/PCS/GVS Warp kernels and launchers, and the dispatch instrumentation tests. The shape documentation still makes the canonical `(E, N)` batch explicit. I also audited argument order: canonical functions use `(model/operands, q, s, operation)`, while public Warp launchers now expose `(q, s, ...)` before prepared/model inputs, scratch buffers, and outputs.

docs/user-guide/execution-backends.md
mstoelzle: Addressed in 16f83f4. The support table now lists `PlanarPCS`, `PCS`, and `GVS` in that order.

src/soromox/systems/execution/warp/gvs/kinematics.py
mstoelzle: Good catch on the ambiguity. The values `4` and `5` were basis-family identifiers (Gaussian and inverse multiquadric), not a basis-order limit; `order` is supplied independently and the evaluator loops over it. In 16f83f4 I replaced every magic family identifier with named `BASIS_FAMILY_*` constants and renamed the local selector to `basis_family`. I also strengthened the Warp/JAX equivalence test to exercise all six families at basis order 6 (`max_dof=13`) on CPU and CUDA.

src/soromox/systems/execution/warp/gvs/kinematics.py
mstoelzle: Addressed in 16f83f4. This kernel parameter and every corresponding PCS/GVS kernel, launcher, executor, and loader parameter are now named `s`. Inside a sample kernel, the selected scalar is named `abscissa` to avoid shadowing the batched `s` array. Public launcher signature tests now enforce `(q, s, ...)` ordering.

vdperfetta01 (commented):
The code looks ok to me , but I noticed that some test are failed, is this something expected to happen?

PR #183, opened 2026-08-28
Add Warp threadlike actuation kernels

mstoelzle:
## Summary

This PR builds on #181 and adds Warp-native execution for built-in linear
threadlike actuation in `PlanarPCS`, `PCS`, and `GVS`.

It provides public allocation-free kernels, operand bundles, output-shape
contracts, and caller-owned launchers for batched actuation matrices and fused
direct-effort forces. It also fuses eligible threadlike actuation into the Warp
forward-dynamics pipeline so the system does not cross separately through the
JAX actuation path.

The runtime-shaped PCS dynamics pipeline is now production-enabled for every
positive Gauss quadrature count. The previous five-point dispatch restriction
has been removed for standalone and fused `PlanarPCS`/`PCS` dynamics.

The implementation keeps the JAX definitions as the semantic and
differentiation reference. Custom routing, mixed transmissions, non-direct
effort laws, passive threadlike impedance, and actuator-coordinate APIs remain
on JAX.

## Stack and merge order

This PR is based on the head of #181 (`codex/fused-kinematics-jacobians`).

1. Merge #181.
2. Retarget this PR to `main`.
3. Merge this PR after the base-only diff disappears.

## Public API and execution policy

The public methods accept the same optional backend selection as the other
accelerated system operations:

```python
robot.actuation_matrix(q, backend="auto")
robot.actuation_force(q, u, qd, backend="auto")
```

The current system-level policy is:

- spatial `PCS.actuation_matrix` may use the Warp implementation on a Warp GPU;
- `PlanarPCS` and `GVS` actuation matrices remain on JAX;
- all direct system-level `actuation_force` calls remain on JAX;
- eligible `PlanarPCS`, `PCS`, and `GVS` forward dynamics fuse threadlike
  direct-effort actuation into the Warp dynamics invocation;
- `PlanarPCS` and `PCS` Warp dynamics support every positive
  `num_gauss_points` value for standalone and fused execution;
- unsupported layouts fall back to JAX under `auto`, while explicit unsupported
  Warp requests explain that only the low-level Warp-native API is available;
- JAX transformations and all derivatives use the protected JAX
  implementations.

Fusion eligibility requires that every active transmission is a built-in
`ThreadlikeActuator` using `ThreadlikeRouting.linear` and `DirectEffort`.

The PCS fused forward path is intentionally enabled for consistency with GVS
and PlanarPCS. The conservative repeat below shows two four-path cases where it
is slightly slower than the previous split Warp-dynamics/JAX-actuation path,
[...]

PR #184, opened 2026-08-29
Add optional floating-base systems

mstoelzle:
## Summary

Adds optional floating-base state, kinematics, dynamics, execution, control,
rollout, and rendering support on top of #183. Fixed-base behavior remains the
default.

- Uses scalar-first Hamilton quaternion plus world position for spatial base
  configuration, and world-frame angular-first velocity at the base origin.
- Supports every system family through JAX.
- Adds public floating Warp kinematics and dynamics for GVS, PCS, and
  PlanarPCS, including batching, fused threadlike actuation, reusable operands,
  and JAX custom-JVP substitution.
- Keeps fixed mounting outside physical parameter PyTrees as
  `fixed_base_pose`; floating poses are supplied through runtime state.
- Updates controllers, rollouts, renderers, execution protocols, API docs, and
  parameter migration examples for unequal `nq`/`nv` layouts.

## Implementation

Continuum models reuse their existing fused local recurrences and contract
inertial contributions immediately; there is no generic mass-sample assembly
or second kinematic traversal. Shared code remains below the entry-kernel level
for strain interpolation, constant-strain operators, Lie/spatial algebra,
quadrature, and actuation.

Fixed and floating Warp entry points, operands, scratch storage, and
accumulators are statically specialized. Floating Warp modules remain separate
so importing a fixed executor cannot register floating kernels or change its
module hash. The fixed GVS, PCS, and PlanarPCS JAX lowerings are byte-identical
to the #183 base. Fixed Warp entry algorithms and launch configurations are
unchanged; shared SO(2), SO(3), and quaternion helpers are organized in common
modules used by both fixed and floating kernels.

Floating Warp kinematics caches each normalized runtime base rotation once per
sample. Spatial Jacobian composition uses one cooperative block per
environment/sample and parallelizes output entries using the block dimension
from the model's backend parameters. Floating Warp dynamics similarly computes
each root rotation once and reuses it across root velocity and gravity products.
The JAX path directly constructs the base Jacobian columns and rotates the
internal angular and linear blocks, avoiding intermediate inverse adjoints and
dense spatial rotation matrices.

Controllers keep `robot` as the complete runtime system and use an explicitly
named `fixed_base_robot` for the documented internal-coordinate approximation.
That robot is mounted at the current runtime pose, but deliberately excludes
base motion and floating/internal dynamic coupling.

## Validation

- Post-review state, kinematics, execution, controller, and
  coordinate-transformation suite: `412 passed, 2 skipped`
- Post-review fixed/floating Warp regression suite: `71 passed, 25 skipped`
  (skips require unavailable CUDA capabilities)
- Broader fixed continuum, articulated, actuation, and floating dynamics suite:
  `478 passed`
- Post-optimization focused floating-base and Warp backend suite: `49 passed`
- Post-rebase coherence and benchmark suite: `72 passed`
- Refreshed post-review StableHLO comparison: all `32/32` fixed artifacts match
  #183 exactly after location normalization, with identical operation counts
- Fixed Warp source audit: no floating inputs, scratch arrays, branches, or
  launches enter fixed kernels; floating operands and entry kernels remain in
  separate modules
- Ruff, `git diff --check`, and the Zensical documentation build pass
- Generated and visually inspected moving-base planar and spatial renderings

## Performance

Benchmarks ran on an NVIDIA GeForce RTX 5090 with JAX 0.11.0 and Warp 1.16.0 in
FP64. The original fixed-path comparison used a fresh process, a sustained 3 s
warm-up, and 101 timed samples. The post-optimization floating/fixed comparison
used four segments, nine kinematic points, batches 256 and 1024, three warm-up
runs, and 21 synchronized samples, alternating fixed and floating measurement
order.

### Floating base disabled

| Backend | Compiled-artifact comparison | Warm runtime vs. #183 | Compile/first-call vs. #183 | GPU memory delta |
| --- | --- | ---: | ---: | ---: |
| JAX | 32/32 fixed StableHLO artifacts exactly match | +1.99% | -0.11% | 0 MiB |
| Warp | Fixed entry algorithms and launch/custom-call graphs unchanged; shared rotation helpers reorganized | -6.07% | -12.85% | 0 MiB |

The JAX runtime result is within the 2.5% regression gate, while the identical
StableHLO demonstrates that the fixed compiled computation is unchanged. The
JAX audit also covers PlanarHSA, ArticulatedSoftRobot, and Pendulum. Fixed Warp
kernels contain no runtime base inputs, augmented scratch matrices, floating
branches, or additional launches. The shared-helper reorganization does not
change their arithmetic. Individual sub-millisecond timings varied in both
directions, and the outliers were not repeatable.

### Floating base activated

The following table reports floating/fixed warm-runtime ratios. Kinematics is
the fused batched pose-and-inertial-Jacobian method; dynamics is the fused
`dynamics_terms` method.

| Backend / system | Kinematics B256 | Kinematics B1024 | Dynamics B256 | Dynamics B1024 |
| JAX PCS | 1.062× | 1.131× | 1.059× | 1.044× |
| JAX GVS | 1.009× | 0.998× | 1.010× | 1.024× |
| Warp PCS | 0.786× | 0.976× | 0.938× | 0.970× |
| Warp GVS | 0.972× | 1.084× | 1.101× | 1.102× |

The previous sharp batched Warp kinematics increase is therefore eliminated:
at batch 256, floating PCS and GVS kinematics measure 21.4% and 2.8% faster than
their fixed counterparts; at batch 1024, PCS remains 2.4% faster and GVS is
8.4% slower. JAX activation remains modest, ranging from effectively unchanged
for GVS to +13.1% for PCS kinematics at batch 1024.

CUDA tracing isolates the primary improvement. For four-segment PCS at batch
256, spatial Jacobian composition fell from 326.6 microseconds to 9.06
microseconds (36.1× faster, -97.2%). Spatial pose composition fell from 28.6 to
9.15 microseconds (-68.0%). Reusing the normalized root rotation in floating
dynamics reduced the persistent PCS kernel from 125.535 to 121.344 microseconds
(-3.3%) and the GVS kernel from 185.695 to 181.119 microseconds (-2.5%).

Alternative GVS block-size selection, shared-memory root rotations, shared
wrench construction, and precomputed JAX base transforms were evaluated and
reverted because their effects were batch-dependent, neutral, or regressive.

Activation increased GPU allocation by at most 2 MiB for the core JAX
workloads. Warp usually added 0-2 MiB; the worst case was batched PCS
kinematics, which increased from 32 MiB to 162 MiB because it materializes total
Jacobians for 256 environments and 64 points.

A supplemental 14-workload JAX pass over PlanarHSA, ArticulatedSoftRobot, and
Pendulum measured an overall +7.48% runtime and +5.09% compile cost. All 14
corresponding fixed StableHLO artifacts exactly match #183. The smallest
articulated and pendulum calls took only 34-829 microseconds, so their ratios
were noisy; absolute activation changes ranged from -312 to +78 microseconds.

## Stack

This PR targets `codex/warp-threadlike-actuation` and should be retargeted to
`main` after #183 merges.


docs/user-guide/floating-base-systems.md
```diff
@@ -0,0 +1,105 @@
...
+separate statically specialized root and accumulator kernels, so fixed kernels
+do not receive floating state or allocate base-sized scratch storage.
+
+Controllers remain internal-coordinate controllers. For a floating robot they
+evaluate a temporary fixed model at the current runtime base pose, so kinematics
+and gravity use the current mounting. They intentionally ignore base velocity,
+acceleration, reaction motion, and floating/internal dynamic coupling, and they
+never actuate the base.
+
+## Newton conversion boundary
```
mstoelzle: Please remove this section

src/soromox/control/operational_space/base_controller.py
```diff
@@ -43,3 +46,23 @@ class OperationalSpaceBaseController(BaseController, ABC):
     """
 
     operational_space_dynamics: OperationalSpaceDynamics
+
```
mstoelzle: Instead of adding `_controller_dynamics_and_state` to the `BaseController` of actuation-space, configuration-space and operational-space, should we not add it to `soromox.control.base_controller.BaseController`?

src/soromox/control/operational_space/impedance_control_tracker.py
```diff
@@ -352,7 +352,7 @@ def __call__(self, system_state: SystemState) -> tuple[Array, Any | None]:
         )
 
         # Get the actuation matrix and compute actuator input
-        A = self.robot.actuation_matrix(q)
+        A = robot.actuation_matrix(q)
```
mstoelzle, line 358: Please double check if this change is correct.

src/soromox/coordinate_transformations/actuation_space.py
```diff
@@ -100,7 +101,8 @@ def __init__(
             ValueError: If H_unactuated has incorrect shape.
             ValueError: If the robot does not have the required attributes.
         """
-        self.robot = robot
+        self.system_robot = robot if robot.floating_base else None
```
mstoelzle: This attribute naming scheme is confusing

src/soromox/rendering/base.py
```diff
@@ -312,6 +319,13 @@ def compute_backbone_curve(self, q: Array) -> Array:
 
     def compute_backbone_poses(self, q: Array) -> Array:
         """Compute full FK poses at the configured backbone sample points."""
+        if bool(getattr(self.robot, "floating_base", False)):
```
mstoelzle: We should add the `floating_base` attribute to the `SoftRobot` base class and then assume that it is present.

src/soromox/rendering/matplotlib_renderer.py
```diff
@@ -1080,11 +1080,25 @@ def _setup_axes(
             elev = np.degrees(np.arctan2(view_vec[2], r_xy))
             ax.view_init(elev=elev, azim=azim)
         else:
-            ax.set_xlim(-width_m / 2, width_m / 2)
-            if center_origin:
-                ax.set_ylim(-width_m / 2, width_m / 2)
+            follows_runtime_base = bool(
+                getattr(self.robot, "floating_base", False)
```
mstoelzle: Same comment as above. Assume that the `floating_base` boolean is present in `self.robot`

src/soromox/systems/articulated/articulated_soft_robot.py
```diff
@@ -612,6 +632,113 @@ def _jacobian_and_time_derivative(
             screws, screw_dots, link_idx, g_s[:3, 3], qd
         )
 
+    @eqx.filter_jit
```
mstoelzle, line 655: Same comment as for the continuum soft robots: Can we not integrate the floating base functionality into the existing dynamics terms method with if/else clauses? This would avoid duplicate code without runtime regression.

src/soromox/systems/articulated/articulated_soft_robot.py
```diff
@@ -728,6 +888,13 @@ def damping_matrix(self, q: Array) -> Array:
         Returns:
             Damping matrix, shape `(num_links, num_links)`.
         """
+        if self.floating_base:
+            _, q_internal = self.split_configuration(q)
+            internal = self.D + self.passive_damping_matrix(q_internal)
```
mstoelzle: Try to avoid such duplicate code structures everywhere as much as possible. In this case, we do the same sum `self.D + passive_damping_matrix` in both branches

docs/development/changelog.md
```diff
@@ -13,6 +13,12 @@ and include benchmark baseline and measurement context for performance claims.
 
 ### Added
```
mstoelzle, line 14: Are there no breaking changes that we need to report?

src/soromox/systems/soft_robot.py
```diff
@@ -143,6 +159,458 @@ def __init__(
...
+        assert v_base is not None
+        if self.is_planar:
+            return jnp.concatenate([v_base, qd_internal], axis=-1)
+        omega_quaternion = jnp.concatenate(
+            [jnp.zeros_like(v_base[..., :1]), v_base[..., :3]], axis=-1
+        )
+        quaternion_dot = 0.5 * quaternion_multiply(omega_quaternion, q[..., :4])
+        return jnp.concatenate([quaternion_dot, v_base[..., 3:6], qd_internal], axis=-1)
+
+    def retract_configuration(self, q: Array, delta_v: Array) -> Array:
```
mstoelzle, line 483: Do we have ful test coverage for all these added `SoftRobot` in `test_soft_robot.py`?

src/soromox/systems/dynamical_system.py
```diff
@@ -118,6 +117,21 @@ def _combine_external_torques(
             return tau_environment
         return base_tau_ext + tau_environment
 
+    def _project_rollout_state(self, y: Array) -> Array:
```
mstoelzle: Maybe this should be a public method as it can also be useful for the user?

src/soromox/systems/soft_robot.py
```diff
@@ -50,20 +55,12 @@ class SoftRobot(DynamicalSystem):
     :meth:`precompute` hook.
 
     Attributes:
-        num_dofs (int): Number of degrees of freedom (configuration variables).
+        num_dofs: Compatibility alias for the generalized-velocity dimension.
```
mstoelzle: This is inconsistent for several reasons:
1. We don't want to preserve compatibility aliases
2. You removed `num_dofs` from `DynamicalSyste`

src/soromox/systems/soft_robot.py
```diff
@@ -143,6 +159,458 @@ def __init__(
...
+        """
+        array = jnp.asarray(value)
+        if array.ndim == 0 or array.shape[-1] != size:
+            raise ValueError(
+                f"{name} must have trailing dimension {size}, got {array.shape}."
+            )
+        return array
+
+    @staticmethod
+    def _broadcast_leading(*arrays: Array) -> tuple[Array, ...]:
```
mstoelzle: Do we have a suitable utility where we could put this helper method?

src/soromox/systems/soft_robot.py
```diff
@@ -155,14 +623,72 @@ def precompute(self) -> None:
         """
         return None
 
+    def _fixed_evaluation_view(self, base_pose: Array | None = None) -> Self:
+        """Return a fixed evaluation view without rebuilding model parameters.
```
mstoelzle: Its not really clear from neither method name nor docstring what this method does.

src/soromox/systems/soft_robot.py
```diff
@@ -155,14 +623,72 @@ def precompute(self) -> None:
...
+        if hasattr(view, "gravity_base") and view.gravity_base is not None:
+            adjoint_inverse = (
+                se2.adjoint_inverse if self.is_planar else se3.adjoint_inverse
+            )
+            object.__setattr__(
+                view, "gravity_base", adjoint_inverse(transform) @ view.g
+            )
+        return view
+
+    def controller_model_and_state(self, y: Array) -> tuple[Self, Array, Array]:
```
mstoelzle: Is this the most suitable location for such a controller-focused function?

src/soromox/systems/soft_robot.py
```diff
@@ -299,10 +826,131 @@ def forward_kinematics(self, q: Array, s: Array) -> Array:
...
+
+        Returns:
+            The absolute model pose at ``s``.
+        """
+        if self.floating_base:
+            return self._floating_forward_kinematics(q, s)
+        return self._forward_kinematics(q, s)
+
+    def _reference_inertial_jacobian(self, q: Array, s: Array) -> Array:
+        """Evaluate one inertial Jacobian through differentiable JAX code.
```
mstoelzle: Very low-quality docstring

src/soromox/systems/soft_robot.py
```diff
@@ -299,10 +826,131 @@ def forward_kinematics(self, q: Array, s: Array) -> Array:
...
+        relative_transform = self._homogeneous_pose(relative_pose)
+        jacobian_internal = self._jacobian_inertialframe(q_internal, s)
+        return self._floating_jacobians_from_relative_inertial(
+            q, relative_transform, jacobian_internal
+        )
+
+    def _reference_forward_kinematics_abscissa_batched(
+        self, q: Array, s: Array
+    ) -> Array:
+        """Evaluate a pose batch through the differentiable JAX traversal.
```
mstoelzle: It might not be very clear what you mean with "reference forward kinematics". Also, remove the code comment for "differentiable JAX traversal".

src/soromox/systems/soft_robot.py
```diff
@@ -299,10 +826,131 @@ def forward_kinematics(self, q: Array, s: Array) -> Array:
...
+            return vmap(
+                lambda transform: poses.planar_pose_from_transform(
+                    transform, self.global_eps
+                )
+            )(absolute)
+        return jnp.einsum("ab,nbc->nac", base_transform, relative_poses)
+
     def _forward_kinematics(self, q: Array, s: Array) -> Array:
         """
         Protected primal forward-kinematics hook for custom autodiff wrappers.
```
mstoelzle: Please improve the docstring

docs/api/systems/index.md
```diff
@@ -49,6 +49,9 @@ specifications are described in
 [Continuum Robot Components](continuum-components.md). Practical immutable
 replacement and gradient-based identification examples are in
 [Parameters, Updates, and Optimization](../../user-guide/parameters-and-optimization.md).
+State layouts, pose and velocity conventions, and the Newton conversion boundary
```
mstoelzle: Please remove the comment about Newton.

src/soromox/systems/execution/warp/pcs/floating_fused.py
```diff
@@ -0,0 +1,267 @@
...
+    derivative_second: wp.array2d[wp.float64],
+    velocity_second: wp.array2d[wp.float64],
+    gravity_second: wp.array2d[wp.float64],
+    inertia: wp.array3d[wp.float64],
+    coriolis_qd: wp.array2d[wp.float64],
+    gravity_force: wp.array2d[wp.float64],
+    actuation_internal: wp.array2d[wp.float64],
+    actuation_force: wp.array2d[wp.float64],
+):
+    """Enqueue augmented spatial PCS dynamics and internal direct effort."""
```
mstoelzle: This docstring style is not aligned with the rest of the package.

src/soromox/systems/execution/warp/pcs/floating_fused.py
mstoelzle, line 1: Would it make more sense to put this into `fused.py` as well?

src/soromox/systems/execution/transforms.py
```diff
@@ -136,7 +136,7 @@ def _evaluate_forward_kinematics_primal(
 ) -> Array:
     """Evaluate the scalar protected pose implementation."""
 
-    return model._forward_kinematics(q, s)
+    return model._reference_forward_kinematics(q, s)
```
mstoelzle: Again, its not clear what "reference forward kinematics" refers to.

src/soromox/systems/execution/warp/gvs/floating_fused.py
mstoelzle, line 1: Would it make sense to move this into `fused.py`?

src/soromox/systems/execution/warp/gvs/floating_chain.py
mstoelzle, line 1: Would it make sense to combin this into `chain.py`?

docs/user-guide/floating-base-systems.md
mstoelzle: Removed the downstream Newton conversion section from the floating-base guide. The SoRoMoX state contract now stands on its own without documenting an external adapter in this PR.

src/soromox/control/operational_space/base_controller.py
mstoelzle: Moved the shared state parser/remounting logic into `soromox.control.base_controller.BaseController`. The actuation- and operational-space base controllers no longer duplicate `_controller_dynamics_and_state`.

src/soromox/control/operational_space/impedance_control_tracker.py
mstoelzle: Double-checked and corrected this path. The controller now remounts `fixed_base_robot` at the runtime pose before all kinematic and dynamics evaluations, and the task-space torque mapping uses `jnp.linalg.solve` instead of forming an inverse. The controller/coordinate-transformation suite passes (`340 passed` in the broad run; included in the refreshed `412 passed, 2 skipped` suite).

src/soromox/coordinate_transformations/actuation_space.py
mstoelzle: Replaced the confusing `system_robot`/`robot` split. `robot` now always means the complete runtime system, while `fixed_base_robot` is the explicitly documented fixed-base, internal-coordinate approximation used by transformed controller dynamics.

src/soromox/rendering/base.py
mstoelzle: `floating_base` is now a required static field on `SoftRobot`, and the renderer accesses it directly without `getattr` compatibility fallbacks.

src/soromox/rendering/matplotlib_renderer.py
mstoelzle: Updated the Matplotlib renderer in the same way: it assumes the `SoftRobot.floating_base` contract and reads the runtime/fixed pose explicitly.

src/soromox/systems/articulated/articulated_soft_robot.py
mstoelzle: Refactored articulated dynamics around one shared relative COM-chain recurrence. Fixed and floating paths now share transforms, screws, internal Jacobians, velocities, and momentum terms; only the final fixed-size versus augmented accumulator remains statically separate. The fixed artifacts remain unchanged in the refreshed StableHLO audit.

src/soromox/systems/articulated/articulated_soft_robot.py
mstoelzle: Removed the duplicated damping expression. The internal damping matrix is assembled once, then returned directly for fixed systems or zero-padded once for floating systems.

docs/development/changelog.md
mstoelzle: Added a Breaking changes subsection covering removal of `SoftRobot.num_dofs` and removal of fixed mounting poses from physical parameter PyTrees, with the replacement APIs documented.

src/soromox/systems/soft_robot.py
mstoelzle: Added `tests/systems/test_soft_robot.py` with concentrated coverage for dimensions, batch-safe pack/split helpers, invalid inputs, normalization and antipodes, derivative/retraction behavior, base transforms, fixed-pose updates, and public state projection. The focused SoftRobot/floating suite passes.

src/soromox/systems/dynamical_system.py
mstoelzle: Promoted this to the public `DynamicalSystem.project_state` hook. Rollouts call it directly, and `SoftRobot` overrides it to normalize spatial floating quaternions.

src/soromox/systems/soft_robot.py
mstoelzle: Removed the `num_dofs` compatibility alias completely. Public code now uses `num_internal_dofs`, `num_coordinates`, or `num_velocities` according to the actual contract; tests assert that the alias is absent.

src/soromox/systems/soft_robot.py
mstoelzle: Moved the helper to the common array utility module as `broadcast_leading_axes` and reused it from the state packing and tangent helpers.

src/soromox/systems/soft_robot.py
mstoelzle: Renamed this to `_fixed_base_robot_at_pose` and rewrote the contract. It returns the original fixed robot when possible, or a shallow fixed-base copy mounted at the requested pose; that copy intentionally excludes base motion and floating/internal dynamic coupling.

src/soromox/systems/soft_robot.py
mstoelzle: Removed the controller-focused state parser from `SoftRobot`. It now lives in the common controller base class as `_controller_robot_and_state`, where the approximation is actually consumed.

src/soromox/systems/soft_robot.py
mstoelzle: Reworked the docstring to match the package format and spell out fixed versus floating behavior, input shapes, absolute output semantics, and use by accelerated-backend derivative substitution.

src/soromox/systems/soft_robot.py
mstoelzle: Removed the ambiguous `reference`/backend naming and the traversal comment. The hooks are now named by semantics: `_absolute_forward_kinematics` and `_absolute_inertial_jacobian`, with batched counterparts. Standard system methods remain the JAX implementations by package convention.

src/soromox/systems/soft_robot.py
mstoelzle: Expanded the protected relative-kinematics hook docstring with the internal-coordinate input contract, relative output representation, subclass responsibility, and raised exception.

docs/api/systems/index.md
mstoelzle: Removed the Newton mention from the systems API index as well as the longer conversion section in the user guide.

src/soromox/systems/execution/warp/pcs/floating_fused.py
mstoelzle: Updated the floating fused launchers to the established Warp launcher docstring structure, including explicit buffer-update semantics and CUDA failure conditions.

src/soromox/systems/execution/warp/pcs/floating_fused.py
mstoelzle: Kept this in `floating_fused.py` intentionally. Merging it into the fixed module causes Warp to register/compile floating kernels when only the fixed executor is imported, changing fixed module hashes and compile behavior. The two modules still share local operators, Lie/spatial primitives, quadrature, and actuation below the entry-launch level; the module doc now explains this boundary.

src/soromox/systems/execution/transforms.py
mstoelzle: Removed `reference` terminology here too. Execution transforms now call the semantically named absolute kinematics/Jacobian hooks; there are no `_jax_*` or `reference forward kinematics` method names left.

src/soromox/systems/execution/warp/gvs/floating_fused.py
mstoelzle: Kept the floating GVS fused launcher separate for the same specialization reason: combining modules changes fixed-only Warp registration and generated artifacts. Shared local work remains in common device functions; only augmented accumulation and entry operands are separate.

src/soromox/systems/execution/warp/gvs/floating_chain.py
mstoelzle: Kept `floating_chain.py` separate because its base columns and augmented scratch/accumulator widths must never be instantiated by fixed kernels. The module documentation now records that constraint, while the local kinematic and spatial primitives remain shared.

mkdocs.yml
```diff
@@ -62,6 +62,7 @@ nav:
   - User Guide:
     - Quick Start: user-guide/quick-start.md
     - Execution Devices and Backends: user-guide/execution-devices-and-backends.md
+    - Floating-base Systems: user-guide/floating-base-systems.md
```
mstoelzle: Please reflect about the best order of the user guides.

vdperfetta01 (approved):
I read through the code, but I didn't spot any incongruences or bugs due to variables renaming :)

PR #185, opened 2026-08-30
Promote execution package to top level

mstoelzle:
## Summary

This PR builds on #184 and promotes the execution-backend package from
`soromox.systems.execution` to the top-level `soromox.execution` namespace.

- Makes execution policy, transformation rules, and reusable Warp-native
  integration APIs a peer of systems, actuation, control, and rendering.
- Keeps `ExecutionBackend`, `GVSBackendParams`, and `PCSBackendParams`
  available from the package root while removing their `soromox.systems`
  re-exports.
- Moves the execution tests from `tests/systems/execution` to
  `tests/execution` and organizes backend, architecture, GVS, PCS, PlanarPCS,
  and cross-family actuation coverage explicitly.
- Updates the API reference, user guide, changelog, benchmark imports, and all
  internal references to the canonical namespace.

## Implementation

The complete `src/soromox/systems/execution` tree moves to
`src/soromox/execution`, including backend-neutral types, configuration,
dispatch, transformation rules, lazy loading, shared Warp mechanics, and the
GVS and PCS family implementations. System classes depend on the neutral
execution layer; the execution layer continues not to import concrete system
implementations.

The Warp hierarchy remains organized by implementation ownership:
family-neutral mechanics under `warp/common`, family-neutral threadlike
actuation under `warp/actuation`, and tightly coupled dynamics, kinematics,
operands, and actuation under `warp/gvs` and `warp/pcs`. The move does not add a
parallel `warp/systems` namespace or split fused family pipelines across
directories.

All 54 moved production Python files match the #184 base after applying only
the namespace substitution. Numerical algorithms, buffer contracts, kernel
launches, and backend-selection behavior are unchanged. Importing `soromox`
or `soromox.execution` still leaves the optional Warp dependency unloaded.

Tests mirror the promoted package at `tests/execution`. Backend-neutral
selection remains in `test_backend.py`; public and lazy-loading contracts live
in the architecture suites; and Warp family behavior is named consistently as
`test_gvs.py`, `test_pcs.py`, and `test_planar_pcs.py`. GVS and PCS architecture
coverage is paired for public names, documented launch contracts, and lazy
operand imports.

## Validation

- Complete execution suite with the declared test dependencies:
  `160 passed, 25 skipped`
- Focused package and Warp architecture suites: `14 passed`
- Mechanical source audit: all `54/54` moved production files match #184 after
  namespace substitution
- Repository-wide audit: no references remain to
  `soromox.systems.execution`, `src/soromox/systems/execution`, or
[...]

PR #186, opened 2026-08-30
Adopt scalar-last quaternion convention

mstoelzle:
## Summary

This PR builds on #185 and changes all SoRoMoX quaternion coordinates from the
scalar-first Hamilton convention `[qw, qx, qy, qz]` to the scalar-last
convention `[qx, qy, qz, qw]`.

- Aligns SoRoMoX with the quaternion representation used by Warp, Newton, and
  standard ROS geometry interfaces.
- Updates quaternion utilities, spatial base and operational-space poses,
  floating-base state integration, and both JAX and Warp execution.
- Migrates examples, tests, API documentation, and the floating-base guide
  while preserving the represented physical orientations.
- Records the coordinate-order migration as a breaking change.

The main motivation is interoperability with the robotics and execution
ecosystem around SoRoMoX. Warp quaternions and Newton rigid-body state use
scalar-last coordinates; standard ROS `geometry_msgs/Quaternion` messages and
`tf2::Quaternion` interfaces likewise use the `x, y, z, w` order. Adopting the
same layout avoids repeated component shuffles at these integration boundaries
and reduces the risk of silently interpreting a valid four-element array in
the wrong order. Scalar-last also matches SciPy's default quaternion
representation, making that common conversion path direct.

This stack is the appropriate point for the breaking change: #184 introduces
runtime quaternion floating-base state, and #185 exposes the execution package
at the top level for integrations such as Newton. Changing the convention here
keeps those new public state and execution interfaces aligned from the outset
instead of establishing a scalar-first layout that downstream integrations
would immediately need to adapt.

## Implementation

Quaternion construction, conversion, multiplication, conjugation, error
calculation, normalization, differentiation, and retraction now consistently
use scalar-last coordinates. Spatial poses use
`[qx, qy, qz, qw, x, y, z]`, and the spatial floating-base configuration uses
the same quaternion prefix followed by world position.

The shared geometry implementation and the floating-base JAX path use the new
layout directly. Warp floating-base kernels read the vector components from
indices 0 through 2 and the scalar from index 3. Existing quaternion literals
in system parameters, examples, renderer tests, and dynamics tests are
reordered so they continue to describe the same rotations.

External interfaces retain their documented conventions. In particular,
Viser's `wxyz` properties remain scalar-first at that API boundary. A SciPy
interop regression test verifies that SoRoMoX quaternion output now matches
SciPy's default scalar-last order without reordering.

## Validation

- Core geometry, pose, floating-base, typed-parameter, and soft-robot suite:
  `203 passed, 2 skipped`
- Focused Warp floating-base suite: `4 passed, 2 skipped` (CUDA-only PCS Warp
  dynamics unavailable on the test host)
- Convention-sensitive system and operational-space suite: `26 passed`
- Renderer suite: `92 passed`; the 8 remaining failures reproduce on #185 and
  come from Open3D test doubles that lack the existing `floating_base` field
- Repository-wide convention audit, `git diff --check`, and the Zensical
  documentation build pass

## Stack

This PR targets `codex/promote-execution-package` and should be retargeted to
`main` after #185 merges.

### DIFF
--- a/docs/api/utilities/parameters.md
+++ b/docs/api/utilities/parameters.md
@@ -115,10 +115,10 @@ base mounting does not rotate the gravity vector.
 The exact default values are:
 
 - Planar upright pose: `[pi / 2, 0, 0]`; gravity: `[0, -9.81]` m/s².
-- Spatial upright pose: `[sqrt(0.5), 0, -sqrt(0.5), 0, 0, 0, 0]`;
+- Spatial upright pose: `[0, -sqrt(0.5), 0, sqrt(0.5), 0, 0, 0]`;
   gravity: `[0, 0, -9.81]` m/s².
-- Planar poses use `[theta, x, y]`. Spatial poses use scalar-first Hamilton
-  quaternions in `[qw, qx, qy, qz, x, y, z]` order.
+- Planar poses use `[theta, x, y]`. Spatial poses use scalar-last Hamilton
+  quaternions in `[qx, qy, qz, qw, x, y, z]` order.
 
 Omitting both the robot constructor's `base_pose` and the parameter tree's
 `gravity` selects the defaults. For example, the factory below constructs an
--- a/docs/development/changelog.md
+++ b/docs/development/changelog.md
@@ -37,10 +37,6 @@ and include benchmark baseline and measurement context for performance claims.
 
 ### Changed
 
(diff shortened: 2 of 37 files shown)


vdperfetta01 (commented):
all good to me

vdperfetta01 (approved):

PR #187, opened 2026-08-30
Add rigid-soft underactuated pendulum benchmarks

mstoelzle:
## Summary

This PR adds a compact set of affine-curvature rigid-soft underactuated
benchmarks based on the Soft Inverted Pendulum (SIP), R-Soft Inverted Pendulum
(R-SIP), Soft Cart-Pole, Soft Pendubot, and Soft Furuta Pendulum.

- Adds GVS factories, physically named upright/hanging configurations,
  constant-input open-loop simulations, four-panel trajectory summaries, and
  Viser playback for all systems.
- Replaces the previous passive, directly clamped cart example with a
  force-actuated cart and passive revolute pole hinge.
- Adds a public `AffineJointActuator` for signed direct force/torque routing,
  with descriptive input bounds that do not implicitly clip controls.
- Adds `explicit_affine` and `material_derived` soft-link stiffness modes.
- Documents model order, coordinate conventions, paper alignment, assumptions,
  and known parameter gaps. Controllers remain outside this PR.

These examples implement the plant side of the papers' control toy problems:
they preserve the underactuated topology, direct input channel, physical
upright/hanging poses, and relevant input limits while deliberately using
low-dimensional affine bending coordinates for fast simulation and controller
development.

## Example and reference mapping

| Example | Control toy system implemented here | Direct input | Primary reference |
| --- | --- | --- | --- |
| Fixed SIP | One fixed affine-curvature soft link, including the paper's pure-tip-torque generalized-force map | Optional external tip torque | [1] |
| R-SIP | The SIP mounted on an actuated revolute base joint; soft coordinates remain passive | Base-joint torque | [2] |
| Soft Cart-Pole | A translating rigid cart, passive revolute hinge, and affine soft pole | Cart force | [3] |
| Soft Pendubot | Two serial `L/2` affine soft links with an actuated first joint and passive second joint | First-joint torque | [3] |
| Soft Furuta Pendulum | A horizontal rigid arm, perpendicular passive soft-link joint, and spatial affine soft link | First vertical-axis joint torque | [3] |

The models provide open-loop rollouts and rendering rather than reproducing the
feedback-linearization, direct-collocation, Box-IDDP, or NMPC controllers from
the papers. They are intended as lightweight SoRoMoX plants on which those and
other control methods can be built and compared separately.

### Signed direct joint actuation

`AffineJointActuator` combines `AffineJointTransmission` with `DirectEffort`.
For routing matrix `R`, it defines actuator coordinates
`y = R @ (q - q_ref) + y0` and maps a signed control to generalized force as
`tau = R.T @ u`, preserving power. Optional bounds, labels, and units are
exposed through actuator metadata; saturation remains the responsibility of a
controller or optimizer.

The benchmark actuation maps are:

| Model | Coordinate order | Actuation column | Metadata bounds |
| --- | --- | --- | ---: |
| R-SIP | `[theta_r, a0, a1]` | `[1, 0, 0]` | unbounded |
| Soft Cart-Pole | `[d, theta, a0, a1]` | `[1, 0, 0, 0]` | `+/-200 N` |
| Soft Pendubot | `[theta1, a10, a11, theta2, a20, a21]` | `[1, 0, 0, 0, 0, 0]` | `+/-10 N m` |
| Soft Furuta | `[theta1, theta2, a0, a1]` | `[1, 0, 0, 0]` | `+/-80 N m` |

### Affine soft links and stiffness modes

Every affine soft link uses length-scaled rotational basis rows,

`kappa(X) = (a0 + a1 X/L) / L`,

so `a0` and `a1` are dimensionless bending-angle coefficients. This is a
SoRoMoX coordinate convention rather than a requirement of the cited papers.

The SIP-family configuration supports two stiffness constructions:

- `explicit_affine`: `K = k H`, retaining the Della Santina/R-SIP generalized
  stiffness coefficient (`k = 1 N m` by default).
- `material_derived`: `K = E I/L H`, with `I = pi r^4/4`; defaults
  `E = 1 MPa`, `r = 0.03 m`, and `L = 1 m` give
  `E I/L = 0.63617255 N m`.

The explicit mode remains the default. The material-derived mode intentionally
changes only stiffness, keeping the SIP distributed mass, square inertial
geometry, and damping fixed so the stiffness choice can be compared in
isolation.

## Paper alignment, deviations, and missing information

| Area | Paper model | This PR | Reason / limitation |
| --- | --- | --- | --- |
| Soft-model order in [3] | Second-order Legendre expansion with 3 DoFs per active strain mode; the Cart-Pole has 11 DoFs and the Pendubot/Furuta each have 20 DoFs | Affine bending only: 4, 6, and 4 DoFs, respectively | This PR introduces low-dimensional control plants. It does not claim to reproduce the high-order full-strain dynamics, out-of-plane deformation, shear, extension, or torsion used in [3]. |
| R-SIP topology and input [2] | Revolute base actuation with passive affine soft coordinates, `q = [theta_r, theta0, theta1]` | Same topology and direct base-joint torque | The plant matches the paper's curvature-parameterized control toy system; the feedback-linearization controller and pick-and-place planner are not included. |
| R-SIP inertia [2] | The reported curvature-parameterization simulation concentrates `1 kg` at the tip | The `1 kg` mass is distributed uniformly along the GVS link | The existing SoRoMoX SIP represents inertia as a distributed continuum link and does not add a separate tip mass. |
| Cart-Pole angle [3] | `theta = 0` hangs downward and `theta = pi` is upright | `theta_model = theta_paper - pi`; all-zero coordinates are upright and the hanging helper uses `theta = -pi` | The offset makes SoRoMoX's zero configuration physically straight and upright. Forward-kinematics tests verify both semantic configurations. |
| Pendubot coordinates [3] | `(theta1, theta2) = (0, 0)` hangs and `(pi, 0)` is upright | Same physical convention | The helper configurations are paper-aligned and verified through forward kinematics. |
| Furuta coordinates [3] | The target is reported as `theta1 = 0`, `theta2 = pi`, plus the soft-link static equilibrium | SoRoMoX uses `theta2 = -pi/2` for upright and `+pi/2` for hanging | The numerical offset follows the chosen orthogonal joint-axis mounting; the physical poses, not raw angle values, define the helpers. |
| Soft-link material in [3] | Cylindrical links with `L = 1 m`, `R = 0.03 m`, `rho = 1000 kg/m^3`, `E = 1 MPa`, `nu = 0.5`, and `beta = 0.01 MPa s` | Pendubot and Furuta use these values with `G = E/3`; SIP/R-SIP/Cart-Pole retain the SIP mass/geometry/damping and expose the optional material-derived bending stiffness | `G = E/3` is equivalent to `nu = 0.5`. The stiffness-only mode avoids silently changing inertia and damping when comparing stiffness models, but is not a full reproduction of [3]'s Cart-Pole link. |
| Rigid cart/pivot properties [3] | The preprint specifies the soft-link material and joint damping but does not report sufficient cart mass/inertia/geometry to construct the rigid body | Exposed assumptions: `0.15 m` length, `0.06 m` radius, and `1 kg` mass on a finite zero-strain GVS link | The finite link supplies rigid inertia and places the passive pole pivot; the values are documented assumptions pending author clarification. |
| Furuta rigid-arm properties [3] | The preprint describes the rigid arm and perpendicular axes but does not report sufficient arm mass/inertia/geometry | Exposed assumptions: `0.5 m` length, `0.02 m` radius, and `0.5 kg` mass on a finite zero-strain GVS link | These values make the benchmark executable but should be replaced if the reference parameters become available. |
| Rotational input units in [3] | The prose reports Pendubot/Furuta limits as `10 N` and `80 N`, while the plotted torque axes use `N m` | Interpreted as `+/-10 N m` and `+/-80 N m` | The inputs act at revolute joints, so torque units are dimensionally consistent; the PR records the ambiguity rather than silently treating them as forces. |
| Input limits | Box constraints are enforced by the optimal controllers | Limits are actuator metadata and are not clipped by the plant | This keeps the actuator linear and makes saturation policy explicit in downstream control/optimization code. |
| Controllers and integration | [2] uses feedback linearization; [3] evaluates DC, Box-IDDP, and NMPC with implicit integration and warm starts | Constant-input open-loop Diffrax rollouts only | Controllers, costs, implicit optimal-control transcription, analytical controller derivatives, and paper performance reproduction are intentionally outside scope. |

## Validation

- Final affected actuation/GVS/example/renderer regression suite:
  `179 passed`
- Focused post-stiffness-mode model and renderer suite: `19 passed`
- Ruff check and formatting: passed
- `git diff --check`: passed
- Headless zero/bounded-input rollouts: finite for all four new underactuated systems
- Material-derived Soft Cart-Pole CLI rollout: completed and rendered successfully
- Summary trajectory figures for R-SIP, Cart-Pole, Pendubot, and Furuta: generated and visually inspected

## References

1. C. Della Santina, “The Soft Inverted Pendulum with Affine Curvature,”
   *2020 59th IEEE Conference on Decision and Control (CDC)*, pp. 4135–4142,
   2020. https://doi.org/10.1109/CDC42340.2020.9303976
2. D. Caradonna, M. Pierallini, C. Della Santina, F. Angelini, and A. Bicchi,
   “Model and Control of R-Soft Inverted Pendulum,” *IEEE Robotics and
   Automation Letters*, vol. 9, no. 6, pp. 5102–5109, 2024.
   https://doi.org/10.1109/LRA.2024.3389348
3. D. Caradonna, N. Nair, A. T. Mathew, D. Feliu-Talegon, I. Afgan,
   E. Falotico, C. Della Santina, and F. Renda, “Soft Swing-up: Benchmarking
   Model-Based Optimal Control for Rigid-Soft Underactuated Systems,”
   arXiv:2602.03435v2, 2026. https://arxiv.org/abs/2602.03435


mstoelzle:
Hi @Elektron97, thank you in advance for taking a look. I would especially
appreciate your guidance on two points about the 2026 preprint:

1. **Model order.** What led you to select the 11-DoF full-strain Soft
   Cart-Pole (and the corresponding 20-DoF Pendubot/Furuta models) instead of a
   reduced affine-curvature / SIP-like model?
2. **Missing rigid-system parameters.** Could you share the rigid cart
   mass/inertia/geometry and pivot placement, as well as the Furuta rigid-arm
   length, cross-section, mass/inertia, and exact joint-frame transforms used
   for the paper? If there is a parameter file or model source we can cite,
   that would be ideal. Could you also confirm that the reported Pendubot and
   Furuta limits of 10 and 80 are intended to be `N m` (as indicated by the
   torque plots), rather than `N` as written in the prose?

The PR documents our current assumptions explicitly so they can be replaced
without ambiguity once the reference values are available.

docs/development/changelog.md
```diff
@@ -22,6 +22,10 @@ and include benchmark baseline and measurement context for performance claims.
   tracking plots, and pressure-colored Viser recording.
 - Added renderer-neutral cross-section contours and loft construction, with a
   registration hook for extending both swept 3D renderers with new geometries.
+- Added signed direct `AffineGeneralizedCoordinateActuator` routing and
```
mstoelzle: There should be separate changelog entries for (i) the `AffineGeneralizedCoordinateActuator` and (ii) the added examples

docs/development/changelog.md
```diff
@@ -35,6 +39,8 @@ and include benchmark baseline and measurement context for performance claims.
   world-down gravity, and aligned Section Vf rendering with zero gravity.
 - Open3D and Viser swept backbones now preserve circular, elliptical, and
   rectangular cross-sections, including dimensions that vary with abscissa.
+- Added selectable `explicit_affine` and `material_derived` stiffness modes to
```
mstoelzle: This is not a delta change. Remove. Instead, you should mention here the renaming of the `AffineJointTransmission`

tests/systems/test_soft_pendulum_examples.py
```diff
@@ -0,0 +1,304 @@
+"""Tests for the low-dimensional rigid-soft pendulum examples."""
+
+import jax.numpy as jnp
+import numpy as np
+import pytest
+from numpy.testing import assert_allclose
```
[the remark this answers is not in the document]
mstoelzle: Addressed in ec4f387: this systems test module now enables JAX x64 before importing jax.numpy, so its tight numerical tolerances no longer depend on test import order.

tests/rendering/test_soft_pendulums_viser_renderer.py
```diff
@@ -0,0 +1,131 @@
+"""Tests for Viser motion rendering of the Soft Cart-Pole."""
+
+from contextlib import contextmanager
+
+import jax.numpy as jnp
+import pytest
+from numpy.testing import assert_allclose
```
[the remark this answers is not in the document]
mstoelzle: Addressed in ec4f387: this renderer test module now enables JAX x64 before importing jax.numpy, matching the repository test convention.

docs/development/changelog.md
mstoelzle: Addressed in ec4f387: the changelog now has separate Added entries for AffineGeneralizedCoordinateActuator and for the new affine-curvature GVS examples.

docs/development/changelog.md
mstoelzle: Addressed in ec4f387: the non-delta stiffness-mode entry was removed and replaced with a Changed entry documenting the AffineJointTransmission and parameter-type rename.

PR #190, opened 2026-08-31
Tendon friction model for PCS systems

Michele-Martini:
## Summary

Threadlike transmissions currently deliver effort undiminished along a routed path. This adds an opt-in transmission loss, expressed as a first-class model object chosen at construction, so a new loss law needs no change to any host and a third-party law needs no change to soromox.

Three laws ship: `Frictionless` (the default), `CapstanFriction`, and `ExponentialLengthFriction`. Everything is inert at the default, which is skipped entirely and reproduces the frictionless result bit for bit.

The PR also corrects a pre-existing sign error in the planar routed-path offset. That fix lands whether or not guide friction is adopted: it changes results for any existing planar threadlike routing, and it is what allows the planar and spatial hosts to be checked against each other for the first time.

---

# Part I — Theory

## Reference model

Guide friction is the dominant unmodelled loss in a tendon-driven continuum robot, and the term that separates commanded tension from delivered tension during identification. The frictionless limit soromox implements today is the special case of the force model in

> D. Feliu-Talegon, A. Y. Alkayas, Y. A. Adamu, A. T. Mathew and F. Renda, "Actuation Reading Insights: Estimating Shape and Forces in Tendon-Driven Slender Soft Robots," *IEEE/ASME Transactions on Mechatronics*, 30(6), 7878-7888, 2025.

There a tendon runs from a motorized capstan through holes in a chain of spacer disks. Each disk contact involves Coulomb friction against the bushing, so the tension is not uniform along the tendon, and the loss at each contact is quantified with the classical Capstan equation. This PR adds the friction term that model carries and soromox currently drops.

Guide friction is not one law, though. It depends on how the path is held: a tendon through spacer disks loses tension where it turns, while a tendon in a close-fitting sheath loses it with distance travelled. Hard-coding either would have to be undone to add the other, so the law is a replaceable object from the outset.

### Step 1 — the classical Capstan (Euler) law

A flexible band drawn over a fixed cylinder with total wrap angle $\varphi$ and Coulomb coefficient $\mu$ transmits

$$
T_{out} = T_{in} e^{-\mu\varphi}
$$

when the input end is the driving end. The result is independent of the drum radius: only the accumulated turn of the band matters. That single fact is what the whole model rests on.

### Step 2 — the rigid guide chain

Thread a tendon through $N$ discrete guides at points $p_k$. Between guides the tendon is straight, with chord length $\nu_k$ and unit vector $u_k$; at guide $k$ it turns by $\varphi_k$. Applying step 1 at each guide and accumulating,

$$
T_{k+1} = T_k e^{-\mu\varphi_k}
\qquad\Longrightarrow\qquad
T_k = T_0 e^{-\mu\sum_j^{k-1}\varphi_j}
$$

The force the tendon applies to the body at guide $k$ is the resultant of the two adjacent spans, $f_k = -T_k u_k + T_{k+1}u_{k+1}$. Taking virtual work and summing by parts, with a fixed base $\delta p_0 = 0$,

$$
\delta W = \sum_k f_k\cdot\delta p_k
= -\sum_k T_k u_k\cdot(\delta p_k - \delta p_{k-1})
= -\sum_k T_k \delta\nu_k
$$

This step needs no assumption whatsoever on the $T_k$, which is the crux of the design: the friction reaction is already contained in $T_k \neq T_{k+1}$, so nothing has to be subtracted afterwards. The friction load belongs inside the actuation matrix, not beside it.

### Step 3 — the continuum limit

Let the guide spacing go to zero. The sums become integrals over the backbone coordinate $s$, and the accumulated turn becomes

$$
\Theta(q, s) = \int_0^s \lVert\omega\times\hat u\rVert \mathrm{d}s'
$$

Define the transmission ratio $\eta(q, s) = e^{-\mu\Theta(q, s)}$, the fraction of the effort applied at the base that survives to arc length $s$, so that $T(s) = e \eta(q,s)$ for a base effort $e$. It equals one at the base and decays monotonically toward the tip.

For one path, $r(s)$ is the material-frame offset and the local tangent is $t = v + \omega\times r + r'$. With $\nu = \lVert t\rVert$ and $\hat u = t/\nu$, the length-gradient density is $\Phi = \partial\nu/\partial\xi = [ r\times\hat u\ ;\ \hat u ]$, and substituting $\delta\xi = B_\xi \delta q$ gives

$$
A_\eta(q) = B_\xi^\top\!\int_0^L \eta(q, s) \Phi(q, s) \mathrm{d}s
$$

At $\mu = 0$ this is $A_0 = (\partial\ell/\partial q)^\top$, today's implementation, so the model is a strict generalization. The scalar $\eta$ multiplies the whole basis, moment row included: the moment arm needs no separate treatment.

### Step 4 — one ODE, two laws

Steps 1 to 3 are the curvature-driven case. Both shipped laws in fact solve the same tension ODE along the path,

$$
\frac{\mathrm{d}T}{\mathrm{d}s} = -\lambda(q, s) T
\qquad\Longrightarrow\qquad
\eta = e^{-\int_0^s \lambda \mathrm{d}s'}
$$

and differ only in where the normal load pressing the path against its guide comes from. `CapstanFriction` takes it from curvature, $N = Tw$ with $w = \lVert\omega\times\hat u\rVert$, giving $\lambda = \mu w$ and $\eta = e^{-\mu\Theta}$. `ExponentialLengthFriction` takes it from a sheath gripping in proportion to tension, $N = cT$, giving a constant $\lambda = \mu c = k$ and $\eta = e^{-ks}$, so loss accrues with distance rather than with turning.

That makes $k$ phenomenological in a way $\mu$ is not: it lumps the friction coefficient with the radial grip, so it must be fitted rather than looked up. It suits a tendon in a close-fitting sheath or lumen, and because it never reads the wrap angle it is the one law `GVS` can host.

Every law must satisfy two rules. $\eta$ depends on the configuration only, never on the effort, so the actuation force stays linear in the effort and every existing controller keeps working; and $\eta \in (0, 1]$, so a law can only remove effort, never add it.

## Wrap-angle density

$\hat T = R\hat u$ gives $\mathrm{d}\hat T/\mathrm{d}s = R(\omega\times\hat u + \hat u')$, so the exact density is $\lVert\omega\times\hat u + \hat u'\rVert$. The implementation drops $\hat u'$, which is first order in shear-strain times routing-slope and identically zero at constant offset, leaving $w = \lVert\omega\times\hat u\rVert$.

In the plane, $\omega = (0,0,\kappa)$ is perpendicular to the in-plane unit tangent, so $w = \lvert\kappa\rvert$ identically, for any offset and any strain. Under pure torsion $\omega = (k_x,0,0)$ at offset radius $\rho$ gives $w = k_x^2\rho/\nu$ with $\nu = \sqrt{1 + k_x^2\rho^2}$, exactly the helix curvature of the resulting path converted from per-tendon-length to per-$s$. An off-axis path under pure torsion does wrap its guides, and this term captures it.

Since $\omega$ is constant per PCS segment, $w$ is piecewise constant and $\Theta$ is piecewise linear, therefore exact at every quadrature node with no nested quadrature and no added error. Planar $w$ is path-independent; spatial $w$ depends on $\hat u$ and so is per path.

### Wrap-angle smoothing

A static `wrap_angle_smoothing` $\epsilon$ on the host structure replaces $\lvert\kappa\rvert$ by $\sqrt{\kappa^2 + \epsilon^2}$. It never under-estimates the wrap angle and over-estimates it by at most $\epsilon L$, so $\eta$ stays in $(0, 1]$, and $\epsilon = 0$ reproduces the exact $\lvert\kappa\rvert$ bit for bit.

It exists for identification. With the exact $\lvert\kappa\rvert$ a straight configuration has $\Theta = 0$, which makes $\mathrm{d}A/\mathrm{d}\mu$ a clean zero, indistinguishable from a converged fit. A nonzero $\epsilon$ gives $\Theta > 0$ even when straight, and also removes the $C^0$ kink on every $\kappa_i = 0$ hyperplane.

## Effect on the energy of the system

With a loss installed the equations of motion read

$$
M\ddot q + C\dot q + \frac{\partial U_g}{\partial q} + \frac{\partial U_{el}}{\partial q} + D(q)\dot q = \tau_u,
\qquad
\tau_u = A_\eta e - (A_\eta - A_0)K \Delta\ell
$$

where $e$ is the work-conjugate effort the effort model produces from the control, $\Delta\ell$ is the passive-path extension, and $K$ the diagonal passive stiffness.

### The robot-side budget stays exact

`elastic_force` keeps only the conservative half of the passive force, so it remains exactly $\partial U_{el}/\partial q$ and

$$
\frac{\mathrm{d}}{\mathrm{d}t}(T + U) = \dot q\cdot\tau_u - \dot q^\top D\dot q
$$

holds verbatim under any law. A transmission loss does not add a dissipation term to the robot; it changes what $\tau_u$ is. This is the headline invariant, and it holds to $10^{-16}$ for all four combinations of actuator and passive coefficients.

### Why the passive force is split

`elastic_force` is wired by a custom JVP to be the gradient of `elastic_energy`. The attenuated spring force is the gradient of nothing: differentiating $A_\eta e$ picks up a per-node outer product $(\partial\eta/\partial\xi)(\partial\nu/\partial\xi)^\top$ whose antisymmetric part is generically nonzero, because $\partial\eta/\partial\xi$ lies in the bending channel alone while $\partial\nu/\partial\xi$ has every component, so the two never cancel. Carrying the attenuated force in `elastic_force` would make `jax.grad(elastic_energy)` disagree with a finite difference of that same function: a silent wrong gradient, unrepairable by redefining the energy. Hence

$$
\tau_{pt} = \underbrace{A_0K \Delta\ell}_{\text{conservative}} + \underbrace{(A_\eta - A_0)K \Delta\ell}_{\text{non-conservative}}
$$

with the first term in `elastic_force` and the second applied through `actuation_force`. The split is bookkeeping: the total force, and the trajectory, are unchanged.

### Where the dissipation goes

Power supplied at the path input is $P_{in} = e s_c (A_0^\top\dot q)$, using the true kinematics; power delivered to the backbone is $P_{out} = \dot q\cdot A_\eta e$. The difference

$$
P_{guides} = e s_c (A_0 - A_\eta)^\top\dot q,
\qquad
A_0 - A_\eta = B_\xi^\top\!\int_0^L (1 - \eta) \Phi \mathrm{d}s
$$

is absorbed by the guides, with $1 - \eta \ge 0$ pointwise. It is not a $D$ term, because it is configuration-rate dependent rather than velocity-quadratic.

### Why the passive damper uses the weighted map on both sides

The damper force conjugate to the path rate admits two forms. Reading the rate through the exact Jacobian, $A_\eta D A_0^\top$, is indefinite: for a single path it is $d (a_\eta\cdot\dot q)(a_0\cdot\dot q)$, negative wherever the two projections disagree in sign, so the damper would inject energy and break passivity. On a two-segment robot the symmetric part of that form has a measured eigenvalue of $-2.2\times10^{-4}$ where the chosen form gives $+1.2\times10^{-4}$.

$A_\eta D A_\eta^\top$ is a congruence transform of a PSD diagonal, hence symmetric positive semi-definite under any law. It is also the physical reading: the damper acts in the weighted work-conjugate coordinate, precisely the one in which $e\cdot\dot y_a = \dot q\cdot\tau_u$. Unlike `elastic_force`, damping carries no energy-gradient contract, so it needs no split.

### Pulling and pushing

$\eta$ is a function of the configuration alone, so both driving directions share the same ratio and $\tau_u$ stays linear in $e$.

Pulling a tendon, $s_c u < 0$, drives the configuration down the routed-length gradient: the backbone bends toward the tendon and compresses, and because $\eta$ decays with arc length the distal nodes are down-weighted, the load shifts proximally, and the tip deflects less than in the frictionless case. Pushing, whether a tendon at $u < 0$ or a push rod at $u > 0$, has $s_c u > 0$: the path lengthens and the backbone bends away from it, while the same $\eta$ still down-weights the distal nodes, because $\eta$ cannot depend on the effort. Both signs flip together in $P_{guides}$, so pushing through curved guides dissipates exactly as pulling does.

### What the model does not cover

Both cases above assume the effort drives the path. When the robot's own motion pays the path out against the effort, $e s_c \dot\ell < 0$ and $P_{guides}$ turns negative: the model appears to return energy. Euler's law gives $e^{-\mu\Theta}$ only when the input end drives and $e^{+\mu\Theta}$ when the load back-drives, whereas $\eta$ always takes the driving branch. Two consequences follow, both pinned by tests.

Work around a closed configuration loop at constant effort is nonzero and orientation dependent, so no actuation potential exists: a loss is intrinsically non-conservative. And the passive remainder is not sign-definite, so at a nonzero passive coefficient an unactuated robot no longer loses energy monotonically, because a back-driven passive path is still modelled as driven. At a zero passive coefficient monotone decay is restored.

The effect is bounded, since $\lvert\eta\Phi\rvert \le \lvert\Phi\rvert$ pointwise: the model never transmits more than the frictionless integrand and cannot manufacture energy without limit. A slip-aware variant would need $\eta$ to depend on $\mathrm{sign}(A_0^\top\dot q)$, which makes $\tau_u$ non-smooth in $\dot q$ and destroys linearity in $e$; that is a different contract and belongs in a subclass. In practice: for quasi-static and identification work the effort drives throughout and the model is exact; for long dynamic rollouts with sign-changing path rates, keep the passive coefficient conservative or zero.

## Pre-existing defect: the planar routed-path offset sign

Independent of friction, and verified against the repo's own kinematics rather than assumed.

`PlanarPCS` defines the backbone by $p' = \mathrm{Rot}(\theta)[\sigma_x, \sigma_y]$ with $\theta' = \kappa$, and places the routed path at $p + d \mathrm{Rot}(\theta)e_y$. Differentiating, $\mathrm{d}(\mathrm{Rot}(\theta)e_y)/\mathrm{d}s = -\kappa \mathrm{Rot}(\theta)e_x$, so

$$
p_d'(s) = (\sigma_x - d\kappa) T + (\sigma_y + d') N
$$

The implementation used $\sigma_x + d\kappa$, with a moment row of matching sign. The spatial host derives the correct expression from $v + \omega\times r + r'$, so planar disagreed with spatial, and planar contradicted its own `_threadlike_path_positions`, which the renderer draws from.

It was not even self-consistent: the axial term matched an offset of $-d$ while the shear term matched $+d$. The two coincide only when $d' = 0$, which is every current test and example, which is why it survived. At $\kappa = 10$, $L = 0.1$, $d = 0.02$ the model returned $0.120$ for a path whose rendered arc length is $0.080$. Both hosts now return $0.080$ and agree to $10^{-17}$.

---

# Part II — Implementation

## The law contract

A law is an object, so the hosts depend on a contract rather than on a formula.

```python
class ThreadlikeFriction(BaseSystemParams):
    is_lossless: ClassVar[bool] = False
    requires_wrap_angle: ClassVar[bool] = True

    @abstractmethod
    def transmission_ratio(self, context: ThreadlikeQuadratureContext) -> Array:
        """Return the surviving effort fraction per path, shape (num_paths,)."""
```

Subclassing `BaseSystemParams` rather than a bare `eqx.Module` supplies `replace`, `validate_structure`, `validate_values`, and the tracer-safe `validate_for_update` a law needs anyway, and makes it a normal parameter subtree that `eqx.tree_at` reaches for identification.

The class variables are how a host serves only what a law reads. `is_lossless` lets a host skip the friction path entirely, which is what keeps the default free. `requires_wrap_angle` lets a host refuse a law it cannot serve.

| Law | Ratio | Parameter | Hosts |
|---|---|---|---|
| `Frictionless` | $1$ | none | all; the default |
| `CapstanFriction` | $e^{-\mu\Theta}$ | $\mu$, Coulomb coefficient | `PCS`, `PlanarPCS` |
| `ExponentialLengthFriction` | $e^{-ks}$ | $k$, per metre | all, including `GVS` |

## The quadrature context

`ThreadlikeQuadratureContext` carries the routed-path geometry at one quadrature node, batched over paths. Some fields mean the same thing on every host and some do not, which decides whether a law is portable.

| Field | Meaning | Portable |
|---|---|---|
| `arc_length` | backbone coordinate $s$ | yes |
| `segment_index` | segment containing the node | yes |
| `offset`, `offset_derivative` | material-frame offset and its $s$-derivative | yes |
| `tangent_norm` | routed-path stretch relative to the backbone | yes |
| `active` | whether the node lies in each path's span | yes |
| `unit_tangent` | normalized tangent, 2-vector planar and 3-vector spatial | no |
| `strain` | 3-vector planar and 6-vector spatial | no |
| `wrap_angle` | accumulated turn $\Theta$, or `None` | only where supplied |

`wrap_angle` is populated only when the installed law sets `requires_wrap_angle`, so the default path costs nothing.

## Files

| File | Change |
|---|---|
| `src/soromox/actuation/friction.py` | new: the law contract, the context, and the three laws |
| `src/soromox/actuation/threadlike.py` | `friction` on the transmission params, presets, impedance, host validation |
| `src/soromox/systems/pcs/planar_pcs.py`, `pcs.py` | geometry and basis split, weighted quadrature, wrap-angle density |
| `src/soromox/systems/gvs/core.py` | same split; accepts wrap-angle-free laws |
| `src/soromox/systems/pcs/structures.py` | `wrap_angle_smoothing` |
| `src/soromox/actuation/core.py`, `src/soromox/systems/soft_robot.py` | the non-conservative passive force |

Each host previously computed the routed geometry inside `_threadlike_local_basis` and discarded everything but the basis. That method is split into a geometry step that builds the context and a basis step that consumes it, so a law reads geometry the host already had rather than recomputing it. The per-path `vmap` already present in the moment matrix batches the context for free, which is exactly the shape `transmission_ratio` must return.

`ThreadlikeImpedance` now holds a `ThreadlikeTransmission`, matching `ArticulatedTendonImpedance`, so the law is declared in one place instead of duplicated onto the impedance parameters.

Host support follows the law, not its value. `PCS` and `PlanarPCS` supply the wrap angle. `GVS` strain varies along arc length, so its wrap angle is not piecewise linear and it supplies none, but it hosts wrap-angle-free laws.

## Opting in

```python
robot = PCS.from_links(
    links,
    structure=PCSStructure(num_gauss_points=5),
    actuators=ThreadlikeActuator.tendons(
        routing, friction=CapstanFriction(coefficient=0.2)
    ),
)

tau = robot.actuation_force(q, jnp.array([5.0, 0.0]))
```

`friction_coefficient=0.2` is kept as sugar for the same thing. Passing both it and `friction=` raises, because the two would silently disagree.

A custom law needs no change to soromox:

```python
class HyperbolicFriction(ThreadlikeFriction):
    a: Array

    def transmission_ratio(self, context):
        return 1.0 / (1.0 + self.a * context.wrap_angle)
```

## Identifying a loss parameter

Loss parameters are ordinary dynamic leaves, so they are traceable and differentiable under `jit`.

```python
@jax.jit
def loss(mu):
    transmission = robot.actuators[0].params.transmission
    identified = robot.update_actuator_params(
        0,
        transmission=transmission.replace(
            friction=transmission.friction.replace(coefficient=mu)
        ),
    )
    residual = (
        identified.elastic_force(q_meas)
        + identified.gravitational_force(q_meas)
        - identified.actuation_force(q_meas, u_meas)
    )
    return jnp.sum(residual**2)


value, grad = jax.value_and_grad(loss)(jnp.array(0.2))
```

The nesting matches the routing idiom already in the repo: one level through a sibling field of `ThreadlikeTransmissionParams`. Swapping the law changes the parameter PyTree structure and so retraces, while replacing a value inside a law does not, so identification should start from `CapstanFriction(coefficient=0.0)` rather than from `Frictionless()`. Note also that $\mathrm{d}A/\mathrm{d}\mu$ vanishes wherever $\Theta = 0$, so set `wrap_angle_smoothing` if the fit can approach a straight pose.

## Migration

- Planar routings must negate their `y` offsets to reproduce previous behaviour. The two are exactly equivalent when the routing slope is zero, which covers every existing test and example.
- `ThreadlikeTransmissionParams.friction_coefficient` becomes `friction`, holding the law.
- `ThreadlikeImpedanceParams.routing` becomes `transmission.routing`. The `ThreadlikeImpedance` constructor keeps its `routing=` keyword.
- A wrap-angle law is refused on `GVS` at any coefficient, including zero, where `friction_coefficient=0.0` was previously accepted. `GVS` supplies no wrap angle for the law to read, and substituting $\Theta = 0$ would silently ignore the law rather than report that it cannot be served. Drop the argument or install `ExponentialLengthFriction`.

---

# Part III — Verification

## Invariants

All hold exactly at the lossless default.

| Invariant | Under a lossy law |
|---|---|
| `moment_matrix == jacrev(coordinates).T` | broken by design; assumed by `coordinate_transformations/actuation_space.py` |
| `ThreadlikeImpedance._length_jacobian == jacrev(path_lengths)` | preserved; it passes no law |
| an actuation potential exists | broken, provably |
| `elastic_force == grad(elastic_energy)` and the energy custom JVPs | preserved; only the conservative half stays |
| $\mathrm{d}(T+U)/\mathrm{d}t = \dot q\cdot\tau_u - \dot q^\top D\dot q$ | preserved under any law |
| `actuation_force(q, 0) == 0` | broken for a loaded passive path; affine in $u$, not linear |
| total dynamics unchanged by the passive split | preserved |
| unactuated energy decreases monotonically | preserved at zero passive coefficient; broken above it |
| added passive damping symmetric PSD | preserved by the $A_\eta D A_\eta^\top$ choice |
| Gauss exactness for constant routing | broken; $A$ becomes mildly `num_gauss_points` dependent |
| $A(q)$ is $C^1$ | $C^0$ by default; `wrap_angle_smoothing` restores smoothness |
| backward pass finite at a collapsed tangent | preserved |
| $\tau = A(q)e$ linear in $e$ | preserved; every controller keeps working |
| `path_lengths`, `path_positions` | unchanged; pure geometry |

## Test coverage

Model, in `tests/actuation/test_threadlike_friction.py`: the Capstan ratio matches the closed form $(1 - e^{-\mu\kappa L})/(\mu\kappa L)$ to $10^{-8}$ and the per-segment ratio decreases monotonically toward the tip; the torsion wrap density matches the helix curvature $k_x^2\rho/\nu$ exactly; the length law matches its own closed form on both PCS hosts and runs on `GVS`, where a Capstan law is still refused.

Energy: the budget identity holds to $10^{-16}$ for all four combinations of actuator and passive coefficients; guide dissipation is positive whenever the effort drives the path, for a pulled tendon and a push rod alike; closed-loop work at constant effort is nonzero and orientation dependent, and vanishes at the default; passive damping is PSD including along the direction that makes the rejected form indefinite; `elastic_force` equals `grad(elastic_energy)` in both `custom_jvp` modes and the split leaves `forward_dynamics` unchanged.

Extension point: a law defined inside the test file installs through the public contract alone, and is asserted to differ from `CapstanFriction`, so a law that accidentally reduced to Capstan cannot pass while appearing to prove extensibility. The host-agnostic context fields are compared between `PlanarPCS` and `PCS`, which is what makes a portable law portable.

Sign fix: a settled two-tendon `PlanarPCS` bends toward the pulled tendon and shortens it, mirrors under the opposite pull, and lengthens it under a push; the analytic path length matches the densely sampled rendered polyline; and planar and spatial moment matrices agree at both zero and nonzero coefficients.

## Validation

866 tests across every file referencing a symbol this PR touches, with no failures.

| Scope | Result |
|---|---|
| `tests/actuation` | 87 passed |
| `test_pcs`, `test_planar_pcs`, 2D/3D coherence, lengths, pressure, shared components | 359 passed |
| `test_actuation_space_dynamics`, `tests/control`, `test_gvs`, articulated, McKibben, defaults, `paper_results` actuation | 420 passed |

```bash
uv run ruff format --check src tests examples   # clean
uv run ruff check src tests examples            # clean
uv run --extra docs zensical build --clean      # no issues found
```

Two source-level probes were applied and reverted, to confirm the tests fail for the right reasons. Forcing the transmission ratio to one fails every friction-dependent test while the lossless and smoothing tests keep passing. Collapsing the conservative split is caught by the `custom_jvp_mode(False)` branch, where real autodiff of the energy disagrees with `elastic_force` by 67 percent.

## Documentation

The `Guide friction` section of the threadlike actuation page is written around the law object, with a `Custom friction laws` subsection mirroring the existing `Custom routing laws` example and a table marking each context field portable or host-shaped. The changelog records the additions, the migration, and the documentation change.

## Scope and follow-up

- Laws whose ratio depends on the effort or on the sign of the path rate, such as a slip-aware Capstan or a tension-dependent loss, are out of scope. Both break the linearity of $\tau = A(q)e$ that every controller in the repo assumes. The quadrature context carries neither effort nor velocity, so the interface cannot express them by accident.
- The Capstan law is not extended to `GVS`. Its wrap angle needs a cumulative quadrature because strain varies along arc length.
- Tendon slack is not modelled. A tendon at negative effort pushes rather than going slack; `push_rods` remains the modality for pushing.
- Cost: the friction path adds one `abs`, `clip` and `sum` over segments plus one `exp` per quadrature node, and is skipped entirely at the default.

### DIFF
--- a/docs/api/actuation/threadlike.md
+++ b/docs/api/actuation/threadlike.md
@@ -18,12 +18,13 @@ The transmission model assumes that:
 - the routing is fixed in the material frame and depends only on arc length;
 - the actuator coordinate is determined by path length, scaled by a constant
   effective area for the pressure-chamber preset; and
-- each path carries one scalar work-conjugate effort, uniform along its route.
+- each path carries one scalar work-conjugate effort, attenuated along its
+  route only by the optional Capstan friction model described below.
 
 Consequently, configuration- or time-dependent rerouting, changes in chamber
-cross-section, along-path friction or slack, and internal actuator dynamics are
-outside the transmission model. Representing these effects requires an extended
-routing, transmission, or effort model.
+cross-section, path slack, and internal actuator dynamics are outside the
+transmission model. Representing these effects requires an extended routing,
+transmission, or effort model.
 
 Threadlike components require continuum segment topology and the host's native
 path-integration hooks. `PCS`, `PlanarPCS`, and `GVS` provide this contract.
@@ -165,13 +166,120 @@ coordinates = robot.actuator_coordinates(q)
 ```
 
 For tendons, `coordinates == -lengths`; for a pressure preset, coordinates are
-equivalent volumes. In every case,
+equivalent volumes. Without guide friction,
 `jax.jacrev(robot.actuator_coordinates)(q) == robot.actuation_matrix(q).T` up
-to quadrature tolerance.
+to quadrature tolerance. A nonzero friction coefficient breaks that identity by
+design, because the moment matrix then carries a transmission loss that path
+length does not; see [Guide friction](#guide-friction).
 
 PCS integrates the local path basis and performs the constant-strain projection
 once. GVS applies its strain basis inside the existing quadrature. The public
 behavior is common, while each host keeps its native evaluation path.
+
+## Guide friction
+
+By default a path transmits its effort undiminished along its whole route. Set
+`friction_coefficient` to model Coulomb friction against the guides the path
+runs through, following the force model of
+[Feliu-Talegon et al. (2025)](https://doi.org/10.1109/TMECH.2025.3550846). A
+tendon threaded through spacer disks loses tension at every contact, and the
+loss at each contact obeys the classical Capstan (Euler) law
+\(T_{out} = T_{in}e^{-\mu\varphi}\), which depends on the accumulated turn
+\(\varphi\) alone and not on the guide radius.
+
+Taking the guide spacing to zero turns the product of per-contact losses into
+
+\[
+\eta(q, s) = e^{-\mu\Theta(q, s)}, \qquad
+\Theta(q, s) = \int_0^s \lVert\omega \times \hat{u}\rVert\,\mathrm{d}s'
+\]
+
+the fraction of the base effort that survives to arc length \(s\), with
+\(\hat{u}\) the unit routed-path tangent. Because \(\eta\) weights the integrand
+of the moment matrix rather than scaling its result, the distal part of a path
+contributes less than the proximal part:
+
+\[
+A_\mu(q) = B_\xi^\top \int_0^L \eta(q, s)\,\Phi(q, s)\,\mathrm{d}s
+\]
+
+\(\eta\) depends on the configuration only, so the actuation force stays linear
+in the effort and every controller keeps working. Since \(\eta \in (0, 1]\), the
+model can only ever remove effort, never add it, and a coefficient of zero
+reproduces the frictionless matrix exactly.
+
+The wrap density reduces to \(\lvert\kappa\rvert\) identically in the plane, for
+any offset and any strain. Under torsion it recovers the helix curvature of an
+off-axis path, so a twisted robot correctly reports a nonzero wrap angle.
+
+```python
+robot = PCS.from_links(
+    links,
+    structure=PCSStructure(num_gauss_points=5),
+    # a scalar shares one coefficient; pass shape (num_paths,) for per-path values
+    actuators=ThreadlikeActuator.tendons(routing, friction_coefficient=0.2),
+)
+
+tau = robot.actuation_force(q, jnp.array([5.0, 0.0]))
+```
+
+`PCS` and `PlanarPCS` implement the wrap-angle model. `GVS` strain varies along
+arc length, so its wrap angle is not piecewise linear; it rejects a nonzero
+coefficient during construction with a descriptive `TypeError`.
+
+### Identifying the coefficient
+
+The coefficient is an ordinary dynamic leaf, so it is traceable and
+differentiable under `jit`:
+
+```python
+def with_friction(mu):
+    transmission = robot.actuators[0].params.transmission.replace(
+        friction_coefficient=mu
+    )
+    return robot.update_actuator_params(0, transmission=transmission)
+
+
+@jax.jit
+def loss(mu):
+    identified = with_friction(mu)
+    residual = (
+        identified.elastic_force(q_meas)
+        + identified.gravitational_force(q_meas)
+        - identified.actuation_force(q_meas, u_meas)
+    )
+    return jnp.sum(residual**2)
+
+
+value, grad = jax.value_and_grad(loss)(jnp.array(0.2))
+```
+
+Zero is a valid optimizer start, but \(\mathrm{d}A/\mathrm{d}\mu\) vanishes
+wherever \(\Theta = 0\), so a perfectly straight pose carries no information
+about \(\mu\). Set `wrap_angle_smoothing` on the host structure to replace
+\(\lvert\kappa\rvert\) with \(\sqrt{\kappa^2 + \epsilon^2}\) if the fit can
+approach a straight configuration. It never under-estimates the wrap angle and
+over-estimates it by at most \(\epsilon L\), and it also removes the \(C^0\)
+kink at zero curvature.
+
+### What changes at a nonzero coefficient
+
+- `moment_matrix` is no longer the transpose of the length gradient, so
+  `actuation_space` conversions that assume the identity do not apply.
+- The generalized actuation force has no potential: work around a closed
+  configuration loop at constant effort is nonzero and orientation dependent.
+- For a loaded passive path, `actuation_force(q, u)` is affine in `u` rather
+  than linear, and `actuation_force(q, 0)` is nonzero. Prefer
+  `actuation_force(q, u)` over `actuation_matrix(q) @ e` wherever passive
+  elements exist.
+- The moment matrix acquires a mild dependence on `num_gauss_points`, because
+  the attenuated integrand is no longer polynomial.
+- `path_lengths`, `path_velocities`, and `path_poses` are unchanged. They are
+  pure geometry at any coefficient.
+- The model assumes the effort drives the path. A back-driven path is still
+  modelled with the driving branch of Euler's law, so keep the passive
+  coefficient conservative or zero for long dynamic rollouts with
+  sign-changing path rates.
 
 ## Passive paths
 
@@ -196,7 +304,10 @@ robot = PCS(
 ```
 
 The passive element contributes to elastic force, damping, and elastic energy;
-it never consumes an active control channel.
+it never consumes an active control channel. A passive path may also declare a
+`friction_coefficient`, in which case part of its spring force becomes
+non-conservative and is applied through `actuation_force`; see
+[Guide friction](#guide-friction).
 
 ## Updating routing and actuator parameters
 
--- a/docs/api/actuation/threadlike.md
+++ b/docs/api/actuation/threadlike.md
@@ -168,7 +168,7 @@ coordinates = robot.actuator_coordinates(q)
 For tendons, `coordinates == -lengths`; for a pressure preset, coordinates are
 equivalent volumes. Without guide friction,
 `jax.jacrev(robot.actuator_coordinates)(q) == robot.actuation_matrix(q).T` up
-to quadrature tolerance. A nonzero friction coefficient breaks that identity by
+to quadrature tolerance. A lossy friction law breaks that identity by
 design, because the moment matrix then carries a transmission loss that path
 length does not; see [Guide friction](#guide-friction).
 
@@ -178,9 +178,21 @@ behavior is common, while each host keeps its native evaluation path.
 
 ## Guide friction
 
-By default a path transmits its effort undiminished along its whole route. Set
-`friction_coefficient` to model Coulomb friction against the guides the path
-runs through, following the force model of
+By default a path transmits its effort undiminished along its whole route. Pass
+a `friction=` law to model a transmission loss. The law is an object, so a new
+loss model needs no change to any host, and a third-party model needs no change
+to soromox.
+
+soromox ships three:
+
+| Law | Ratio | Hosts |
+|---|---|---|
+| `Frictionless` | \(1\) | all; the default |
+| `CapstanFriction` | \(e^{-\mu\Theta}\) | `PCS`, `PlanarPCS` |
+| `ExponentialLengthFriction` | \(e^{-ks}\) | all, including `GVS` |
+
+`CapstanFriction` models Coulomb friction against the guides a path runs
+through, following the force model of
 [Feliu-Talegon et al. (2025)](https://doi.org/10.1109/TMECH.2025.3550846). A
 tendon threaded through spacer disks loses tension at every contact, and the
 loss at each contact obeys the classical Capstan (Euler) law
@@ -203,10 +215,11 @@ contributes less than the proximal part:
 A_\mu(q) = B_\xi^\top \int_0^L \eta(q, s)\,\Phi(q, s)\,\mathrm{d}s
 \]
 
-\(\eta\) depends on the configuration only, so the actuation force stays linear
-in the effort and every controller keeps working. Since \(\eta \in (0, 1]\), the
-model can only ever remove effort, never add it, and a coefficient of zero
-reproduces the frictionless matrix exactly.
+Every law obeys the same two rules. \(\eta\) depends on the configuration only,
+never on the effort, so the actuation force stays linear in the effort and every
+controller keeps working; and \(\eta \in (0, 1]\), so a law can only ever remove
+effort, never add it. The lossless default reproduces the frictionless matrix
+exactly and is skipped entirely, at no cost.
 
 The wrap density reduces to \(\lvert\kappa\rvert\) identically in the plane, for
 any offset and any strain. Under torsion it recovers the helix curvature of an
@@ -217,27 +230,39 @@ robot = PCS.from_links(
     links,
     structure=PCSStructure(num_gauss_points=5),
     # a scalar shares one coefficient; pass shape (num_paths,) for per-path values
-    actuators=ThreadlikeActuator.tendons(routing, friction_coefficient=0.2),
+    actuators=ThreadlikeActuator.tendons(
+        routing, friction=CapstanFriction(coefficient=0.2)
+    ),
 )
 
 tau = robot.actuation_force(q, jnp.array([5.0, 0.0]))
 ```
 
-`PCS` and `PlanarPCS` implement the wrap-angle model. `GVS` strain varies along
-arc length, so its wrap angle is not piecewise linear; it rejects a nonzero
-coefficient during construction with a descriptive `TypeError`.
+`friction_coefficient=0.2` is kept as sugar for the same thing. Passing both it
+and `friction=` raises, because the two would silently disagree.
 
-### Identifying the coefficient
+A law declares what it needs, and a host refuses a law it cannot serve.
+`CapstanFriction` sets `requires_wrap_angle`, which `PCS` and `PlanarPCS`
+supply. `GVS` strain varies along arc length, so its wrap angle is not
+piecewise linear and it supplies none; it rejects a wrap-angle law during
+construction with a descriptive `TypeError`, at any coefficient including zero,
+because capability follows the law you install rather than the value you gave
+it. `ExponentialLengthFriction` reads only arc length, so it runs on `GVS` too.
 
-The coefficient is an ordinary dynamic leaf, so it is traceable and
+### Identifying a loss parameter
+
+Loss parameters are ordinary dynamic leaves, so they are traceable and
 differentiable under `jit`:
 
 ```python
 def with_friction(mu):
-    transmission = robot.actuators[0].params.transmission.replace(
-        friction_coefficient=mu
+    transmission = robot.actuators[0].params.transmission
+    return robot.update_actuator_params(
+        0,
+        transmission=transmission.replace(
+            friction=transmission.friction.replace(coefficient=mu)
+        ),
     )
-    return robot.update_actuator_params(0, transmission=transmission)
 
 
 @jax.jit
@@ -254,15 +279,64 @@ def loss(mu):
 value, grad = jax.value_and_grad(loss)(jnp.array(0.2))
 ```
 
-Zero is a valid optimizer start, but \(\mathrm{d}A/\mathrm{d}\mu\) vanishes
-wherever \(\Theta = 0\), so a perfectly straight pose carries no information
-about \(\mu\). Set `wrap_angle_smoothing` on the host structure to replace
+Swapping the law changes the parameter PyTree structure and so retraces, while
+replacing a value inside a law does not. Start identification from a lossy law
+at a zero parameter, `CapstanFriction(coefficient=0.0)`, rather than from
+`Frictionless()`. Zero is a valid optimizer start, but
+\(\mathrm{d}A/\mathrm{d}\mu\) vanishes wherever \(\Theta = 0\), so a perfectly
+straight pose carries no information about \(\mu\). Set `wrap_angle_smoothing` on the host structure to replace
 \(\lvert\kappa\rvert\) with \(\sqrt{\kappa^2 + \epsilon^2}\) if the fit can
 approach a straight configuration. It never under-estimates the wrap angle and
 over-estimates it by at most \(\epsilon L\), and it also removes the \(C^0\)
 kink at zero curvature.
 
-### What changes at a nonzero coefficient
+### Custom friction laws
+
+The hosts depend only on the law contract, not on the shipped models. A custom
+law supplies the parameters it needs, declares its requirements, and returns the
+surviving effort fraction per path:
+
+```python
+from soromox.actuation import ThreadlikeFriction
+
+
+class HyperbolicFriction(ThreadlikeFriction):
+    a: Array
+
+    def transmission_ratio(self, context):
+        return 1.0 / (1.0 + self.a * context.wrap_angle)
+
+
+actuator = ThreadlikeActuator.tendons(
+    routing, friction=HyperbolicFriction(a=jnp.asarray(0.5))
+)
+```
+
+The `context` is a `ThreadlikeQuadratureContext` describing the routed path at
+one quadrature node, batched over paths. Some of its fields mean the same thing
+on every host and some do not:
+
+| Field | Meaning | Portable |
+|---|---|---|
+| `arc_length` | backbone coordinate \(s\) | yes |
+| `segment_index` | segment containing the node | yes |
+| `offset`, `offset_derivative` | material-frame offset `(n, 3)` and its \(s\)-derivative | yes |
+| `tangent_norm` | routed-path stretch relative to the backbone | yes |
+| `active` | whether the node lies in each path's span | yes |
+| `unit_tangent` | normalized tangent, `(n, 2)` planar and `(n, 3)` spatial | no |
+| `strain` | `(n, 3)` planar and `(n, 6)` spatial | no |
+| `wrap_angle` | accumulated turn \(\Theta\), or `None` | only where supplied |
+
+A law reading only the portable fields runs unchanged on every host. A law
+reading `strain` or `unit_tangent` must document which host it targets.
+
+Set `requires_wrap_angle = False` if the law does not read `wrap_angle`, and the
+hosts will neither compute it nor refuse the law. Set `is_lossless = True` for a
+law that always returns one, and the hosts will skip it entirely. A law must
+keep its ratio in \((0, 1]\) and must not read the effort, which would make the
+actuation force nonlinear in the control.
+
+### What changes under a lossy law
 
 - `moment_matrix` is no longer the transpose of the length gradient, so
   `actuation_space` conversions that assume the identity do not apply.
@@ -305,7 +379,7 @@ robot = PCS(
 
 The passive element contributes to elastic force, damping, and elastic energy;
 it never consumes an active control channel. A passive path may also declare a
-`friction_coefficient`, in which case part of its spring force becomes
+`friction=` law, in which case part of its spring force becomes
 non-conservative and is applied through `actuation_force`; see
 [Guide friction](#guide-friction).
 
--- a/docs/api/actuation/threadlike.md
+++ b/docs/api/actuation/threadlike.md
@@ -185,11 +185,25 @@ to soromox.
 
 soromox ships three:
 
-| Law | Ratio | Hosts |
-|---|---|---|
-| `Frictionless` | \(1\) | all; the default |
-| `CapstanFriction` | \(e^{-\mu\Theta}\) | `PCS`, `PlanarPCS` |
-| `ExponentialLengthFriction` | \(e^{-ks}\) | all, including `GVS` |
+| Law | Ratio | Parameter | Hosts |
+|---|---|---|---|
+| `Frictionless` | \(1\) | none | all; the default |
+| `CapstanFriction` | \(e^{-\mu\Theta}\) | \(\mu\), Coulomb coefficient | `PCS`, `PlanarPCS` |
+| `ExponentialLengthFriction` | \(e^{-ks}\) | \(k\), per metre | all, including `GVS` |
+
+Both lossy laws solve the same tension ODE
+\(\mathrm{d}T/\mathrm{d}s = -\lambda(s)T\), whose solution is
+\(\eta = e^{-\int\lambda\,\mathrm{d}s}\). They differ only in where the normal
+load pressing the path against its guide comes from. Under `CapstanFriction`
+it is curvature-induced, \(N = Tw\), giving \(\lambda = \mu w\). Under
+`ExponentialLengthFriction` it comes from a snug sheath gripping in proportion
+to tension, \(N = cT\), giving a constant \(\lambda = \mu c = k\), so loss
+accrues with distance travelled rather than with turning.
+
+That makes \(k\) phenomenological in a way \(\mu\) is not: it lumps the
+friction coefficient together with the radial grip, so it has to be fitted
+rather than looked up. It suits a tendon in a close-fitting sheath or lumen,
+and because it never reads the wrap angle it is the one law `GVS` can host.
 
 `CapstanFriction` models Coulomb friction against the guides a path runs
 through, following the force model of
--- a/docs/api/actuation/threadlike.md
+++ b/docs/api/actuation/threadlike.md
@@ -19,7 +19,7 @@ The transmission model assumes that:
 - the actuator coordinate is determined by path length, scaled by a constant
   effective area for the pressure-chamber preset; and
 - each path carries one scalar work-conjugate effort, attenuated along its
-  route only by the optional Capstan friction model described below.
+  route by an optional friction model described below.
 
 Consequently, configuration- or time-dependent rerouting, changes in chamber
 cross-section, path slack, and internal actuator dynamics are outside the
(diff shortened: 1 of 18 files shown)


docs/api/actuation/threadlike.md
```diff
@@ -165,14 +166,209 @@ coordinates = robot.actuator_coordinates(q)
...
 
 PCS integrates the local path basis and performs the constant-strain projection
 once. GVS applies its strain basis inside the existing quadrature. The public
 behavior is common, while each host keeps its native evaluation path.
 
+## Guide friction
+
+By default a path transmits its effort undiminished along its whole route. Pass
+a `friction=` law to model a transmission loss. The law is an object, so a new
+loss model needs no change to any host, and a third-party model needs no change
```
mstoelzle: >  The law is an object, so a new loss model needs no change to any host, and a third-party model needs no change to soromox.

Please remove such comments.

docs/api/actuation/threadlike.md
```diff
@@ -165,14 +166,209 @@ coordinates = robot.actuator_coordinates(q)
...
+Both lossy laws solve the same tension ODE
+\(\mathrm{d}T/\mathrm{d}s = -\lambda(s)T\), whose solution is
+\(\eta = e^{-\int\lambda\,\mathrm{d}s}\). They differ only in where the normal
+load pressing the path against its guide comes from. Under `CapstanFriction`
+it is curvature-induced, \(N = Tw\), giving \(\lambda = \mu w\). Under
+`ExponentialLengthFriction` it comes from a snug sheath gripping in proportion
+to tension, \(N = cT\), giving a constant \(\lambda = \mu c = k\), so loss
+accrues with distance travelled rather than with turning.
+
+That makes \(k\) phenomenological in a way \(\mu\) is not: it lumps the
```
mstoelzle: Very strange paragraph

docs/api/actuation/threadlike.md
```diff
@@ -165,14 +166,209 @@ coordinates = robot.actuator_coordinates(q)
...
+\(\hat{u}\) the unit routed-path tangent. Because \(\eta\) weights the integrand
+of the moment matrix rather than scaling its result, the distal part of a path
+contributes less than the proximal part:
+
+\[
+A_\mu(q) = B_\xi^\top \int_0^L \eta(q, s)\,\Phi(q, s)\,\mathrm{d}s
+\]
+
+Every law obeys the same two rules. \(\eta\) depends on the configuration only,
+never on the effort, so the actuation force stays linear in the effort and every
```
mstoelzle: Please remove verbose clauses such as "never on the effort".

docs/api/actuation/threadlike.md
```diff
@@ -165,14 +166,209 @@ coordinates = robot.actuator_coordinates(q)
...
+    # a scalar shares one coefficient; pass shape (num_paths,) for per-path values
+    actuators=ThreadlikeActuator.tendons(
+        routing, friction=CapstanFriction(coefficient=0.2)
+    ),
+)
+
+tau = robot.actuation_force(q, jnp.array([5.0, 0.0]))
+```
+
+`friction_coefficient=0.2` is kept as sugar for the same thing. Passing both it
```
mstoelzle: > `friction_coefficient=0.2` is kept as sugar for the same thing.

What??? lol

docs/api/actuation/threadlike.md
```diff
@@ -165,14 +166,209 @@ coordinates = robot.actuator_coordinates(q)
...
+
+A law declares what it needs, and a host refuses a law it cannot serve.
+`CapstanFriction` sets `requires_wrap_angle`, which `PCS` and `PlanarPCS`
+supply. `GVS` strain varies along arc length, so its wrap angle is not
+piecewise linear and it supplies none; it rejects a wrap-angle law during
+construction with a descriptive `TypeError`, at any coefficient including zero,
+because capability follows the law you install rather than the value you gave
+it. `ExponentialLengthFriction` reads only arc length, so it runs on `GVS` too.
+
+### Identifying a loss parameter
```
mstoelzle: This doc should focus on the actuation/friction model and not how such a model would be used during optimization.

docs/api/actuation/threadlike.md
```diff
@@ -165,14 +166,209 @@ coordinates = robot.actuator_coordinates(q)
...
+)
+```
+
+The `context` is a `ThreadlikeQuadratureContext` describing the routed path at
+one quadrature node, batched over paths. Some of its fields mean the same thing
+on every host and some do not:
+
+| Field | Meaning | Portable |
+|---|---|---|
+| `arc_length` | backbone coordinate \(s\) | yes |
```
mstoelzle: `arc_length` is inconsistent with the naming scheme we use in the package. I would go with `s` or `abscissa`.

docs/api/actuation/threadlike.md
```diff
@@ -165,14 +166,209 @@ coordinates = robot.actuator_coordinates(q)
...
+| `segment_index` | segment containing the node | yes |
+| `offset`, `offset_derivative` | material-frame offset `(n, 3)` and its \(s\)-derivative | yes |
+| `tangent_norm` | routed-path stretch relative to the backbone | yes |
+| `active` | whether the node lies in each path's span | yes |
+| `unit_tangent` | normalized tangent, `(n, 2)` planar and `(n, 3)` spatial | no |
+| `strain` | `(n, 3)` planar and `(n, 6)` spatial | no |
+| `wrap_angle` | accumulated turn \(\Theta\), or `None` | only where supplied |
+
+A law reading only the portable fields runs unchanged on every host. A law
```
mstoelzle: What do you mean with "that it remains unchanged on every host"?

src/soromox/actuation/core.py
```diff
@@ -168,6 +168,26 @@ def elastic_energy(self, robot: SoftRobot, q: Array) -> Array:
         del robot
         return jnp.zeros((), dtype=q.dtype)
 
+    def nonconservative_force(self, robot: SoftRobot, q: Array) -> Array:
+        """Return the part of this element's force that has no potential.
+
+        ``elastic_force`` is contracted to be the gradient of
+        ``elastic_energy``, so a force an element transmits but cannot derive
+        from a potential belongs here instead, keeping both contracts
```
mstoelzle: > keeping both contracts truthful. ``SoftRobot.actuation_force`` applies the summed result, so the total generalized force is unchanged by the split.

Please remove verbose comments such as this

src/soromox/actuation/friction.py
```diff
@@ -0,0 +1,259 @@
+"""Transmission-loss laws along threadlike routed paths.
+
+A threadlike transmission carries effort from the base of a routed path toward
+its tip. Guides, seals, and the surrounding material remove some of that effort
+before it reaches the backbone. A :class:`ThreadlikeFriction` model expresses
+how much survives, as a fraction ``eta(q, s)`` of the base effort, and the host
+folds that fraction into the arc-length quadrature that assembles the moment
+matrix.
+
+Because the fraction weights the integrand rather than scaling the result, the
```
mstoelzle: Verbose clause

src/soromox/actuation/friction.py
```diff
@@ -0,0 +1,259 @@
...
+    Instances are produced by the host inside its per-path ``vmap``, so every
+    array field carries a leading ``(num_paths,)`` axis.
+
+    Some fields mean the same thing on every host and some do not. A law that
+    reads only the host-agnostic fields runs unchanged on ``PlanarPCS``,
+    ``PCS``, and ``GVS``; a law that reads a host-shaped field must document
+    which host it targets.
+
+    Attributes:
+        arc_length: Backbone coordinate ``s`` of the node, in metres.
```
mstoelzle: Rename to `s` or `abscissa`

src/soromox/actuation/friction.py
```diff
@@ -0,0 +1,259 @@
...
+        ValueError: If the value is neither a scalar nor shaped ``(count,)``.
+    """
+    shape = jnp.asarray(value).shape
+    if shape not in ((), (count,)):
+        raise ValueError(
+            f"{name} must be a scalar or have shape ({count},), got {shape}."
+        )
+
+
+class ThreadlikeQuadratureContext(eqx.Module):
```
mstoelzle: Please double check if adding this context structure is necessary or we could do without

src/soromox/actuation/friction.py
```diff
@@ -0,0 +1,259 @@
...
+            self,
+            wrap_angle,
+            is_leaf=lambda leaf: leaf is None,
+        )
+
+
+class ThreadlikeFriction(BaseSystemParams):
+    """Fraction of routed-path effort that survives to a quadrature node.
+
+    Subclasses declare what they need from the host through the class
```
mstoelzle: verbose

src/soromox/actuation/friction.py
```diff
@@ -0,0 +1,259 @@
...
+        """
+        return eqx.tree_at(
+            lambda context: context.wrap_angle,
+            self,
+            wrap_angle,
+            is_leaf=lambda leaf: leaf is None,
+        )
+
+
+class ThreadlikeFriction(BaseSystemParams):
```
mstoelzle: Are these supposed to be parameter containers? 
- If not, then its incorrect to inherit from `BaseSystemParams`.
- If so, then we might want to consider to append a "Params". E.g., `ThreadlikeFrictionParams`

src/soromox/actuation/friction.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def validate_for_paths(self, num_paths: int) -> None:
+        """Validate the coefficient shape against the routed-path count."""
+        validate_loss_parameter_shape(self.coefficient, num_paths, "coefficient")
+
+
+class ExponentialLengthFriction(ThreadlikeFriction):
+    """Loss accumulating with routed-path length rather than with turning.
+
+    Both shipped laws solve the same tension ODE ``dT/ds = -lambda(s) T``,
```
mstoelzle: Why are you referring here to "both laws"?

src/soromox/actuation/threadlike.py
```diff
@@ -101,6 +111,67 @@ def _channel_array(value: Any, count: int, name: str) -> Array:
     return array
 
 
+def _resolve_friction(
+    friction: ThreadlikeFriction | None,
+    friction_coefficient: Any,
+) -> ThreadlikeFriction:
+    """Resolve the two spellings of a transmission-loss law into one model.
+
+    ``friction_coefficient`` is retained sugar for the common Capstan case and
```
mstoelzle: Please rephrase

src/soromox/actuation/threadlike.py
```diff
@@ -101,6 +111,67 @@ def _channel_array(value: Any, count: int, name: str) -> Array:
     return array
 
 
+def _resolve_friction(
```
mstoelzle: This should probably go into `friction.py`

src/soromox/actuation/threadlike.py
```diff
@@ -392,15 +485,32 @@ def coordinates(self, robot, q: Array) -> Array:
...
         )
 
+    @property
+    def friction(self) -> ThreadlikeFriction:
+        """Transmission-loss law installed on this path family."""
+        return self.params.friction
+
     def moment_matrix(self, robot, q: Array) -> Array:
-        raw_matrix = robot._threadlike_moment_matrix(q, self.routing)
+        raw_matrix = robot._threadlike_moment_matrix(
```
mstoelzle, line 436: Is it correct that the friction will be contained as part of the actuation matrix?

src/soromox/systems/pcs/pcs.py
```diff
@@ -103,6 +107,7 @@ class PCS(SoftRobot):
     num_strains: int = eqx.field(static=True)  # Number of strains (6 * num_segments)
     _segment_dof_ends: tuple[int, ...] = eqx.field(static=True)
     scale_rotational_basis_by_length: bool = eqx.field(static=True)
+    wrap_angle_smoothing: float = eqx.field(static=True, default=0.0)
```
mstoelzle: I don't think this attribute name is sufficiently semantically meaningful.

src/soromox/systems/pcs/pcs.py
```diff
@@ -2671,26 +2686,155 @@ def _threadlike_local_basis(
...
+        self, strains: Array, routing: ThreadlikeRouting
+    ) -> Array:
+        """Return the Capstan wrap-angle density of each segment and path.
+
+        The density is ``|omega x u_hat|``, the turning rate of the routed path
+        tangent with its material-frame rotation dropped. Torsion contributes
+        whenever the path is off-axis: for a constant offset at radius ``rho``
+        under pure twist ``kx`` it evaluates to ``kx**2 rho / nu``, the
+        curvature of the resulting helix. In a planar configuration the angular
+        strain is perpendicular to the unit tangent and it reduces to
```
mstoelzle: I don't think we should discuss the planar case in the `PCS ` class

src/soromox/systems/pcs/pcs.py
```diff
@@ -316,6 +321,12 @@ def __init__(
         self.scale_rotational_basis_by_length = bool(
             structure.scale_rotational_basis_by_length
         )
+        self.wrap_angle_smoothing = float(structure.wrap_angle_smoothing)
```
mstoelzle: To me, it is very questionable that this should be a model attribute. Instead, I would intuitively expect the wrap angle smoothing to be an entry in the friction params or the threadlike actuation params. Would this be feasible?

src/soromox/systems/pcs/pcs.py
```diff
@@ -2671,26 +2686,155 @@ def _threadlike_local_basis(
...
+        )
+
+    def _threadlike_local_basis(self, context: ThreadlikeQuadratureContext) -> Array:
+        """Return one spatial routed-path length gradient density.
+
+        Args:
+            context: Node geometry from :meth:`_threadlike_local_geometry`.
+
+        Returns:
+            phi: Length-gradient density of shape ``(6,)``.
```
mstoelzle: I would say let's avoid a mathematical symbol name `phi` here because no user is going to know what we mean with `phi` (its not established and agreed upon in soft robot modeling and control literature). Instead, refer to it by its physical meaning. For example, `length_gradient_density` or something similar

mstoelzle: This comment doesn't just apply to this method, but also to other methods and functions where `phi` or other mathematical symbols are used in argument, attribute, or variable names.

src/soromox/systems/pcs/structures.py
```diff
@@ -57,11 +57,23 @@ class PCSStructure(eqx.Module):
             ``None`` activates every strain coordinate.
         scale_rotational_basis_by_length: Whether rotational strain coordinates
             are normalized by their link length.
+        wrap_angle_smoothing: Regularizer for the threadlike Capstan wrap
```
mstoelzle: This should not be in the `PCSStructure`. It should either go into the friction params, the threadlike actuation params, or the threadlike actuation structure.

mstoelzle: Same comment not only applies to the `PCS` system, but also to the `PlanarPCS` and `PCS` systems.

mstoelzle (changes requested):
Thanks @Michele-Martini for contributing the tendon friction - I think it is very interesting to have this feature in the SoRoMoX package.

Overally, I think mainly the implementation needs to be cleaned-up. Particularly, the docstrings and docs entries are still of quite bad quality ("AI slop").

Also, can you please explain to me the wrap-angle / wrap-density more intuitively to me?

docs/api/actuation/threadlike.md
Michele-Martini: Removed friction_coefficient altogether. Yeah the sugar made me roll over lmao

src/soromox/actuation/friction.py
Michele-Martini: I put it for generality. This should allow someone to be able to write its own custom FrictionModel. To do so, one may need information about the robot state.

src/soromox/actuation/friction.py
Michele-Martini: Split into pure ThreadlikeFriction with *Params field and transmission_rate method.

src/soromox/actuation/threadlike.py
Michele-Martini: According to the reference, yes, as it may be in general configuration dependent, and are physically generated by the routing.

src/soromox/systems/pcs/pcs.py
Michele-Martini: renamed

src/soromox/systems/pcs/pcs.py
Michele-Martini: moved to CapstanFrictionParams.eps

docs/api/actuation/threadlike.md
Michele-Martini: section removed

docs/api/actuation/threadlike.md
Michele-Martini: rephrased

src/soromox/systems/pcs/planar_pcs.py
```diff
@@ -2571,7 +2706,22 @@ def point_matrix(point_index: Array) -> Array:
...
+                    # The planar wrap angle is path-independent, so broadcast it
+                    # to one entry per path before the law consumes it.
+                    context = context.with_wrap_angle(
+                        jnp.broadcast_to(
+                            self._threadlike_wrap_angle(
+                                wrap_density, points[point_index]
+                            ),
+                            (count,),
+                        )
+                    )
```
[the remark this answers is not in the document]
Michele-Martini: wrap angle and path abscissa now consider the starting segment

docs/development/changelog.md
```diff
@@ -13,6 +13,30 @@ and include benchmark baseline and measurement context for performance claims.
 
 ### Added
 
+- Added opt-in Capstan guide friction to threadlike transmissions. A
+  `friction_coefficient` on `ThreadlikeActuator` and `ThreadlikeImpedance`
+  attenuates effort along a routed path as `exp(-mu * Theta(s))`, where `Theta`
+  is the accumulated turn of the path, instead of applying it uniformly. `PCS`
+  and `PlanarPCS` implement the wrap-angle model; `GVS` rejects a nonzero
+  coefficient during construction. The coefficient is a differentiable leaf and
```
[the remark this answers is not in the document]
Michele-Martini: fixed in changelog

src/soromox/actuation/friction.py
```diff
@@ -0,0 +1,259 @@
...
+
+    requires_wrap_angle: ClassVar[bool] = True
+
+    def transmission_ratio(self, context: ThreadlikeQuadratureContext) -> Array:
+        """Return the Capstan ratio ``exp(-mu * Theta)`` at the node."""
+        return jnp.exp(-jnp.asarray(self.coefficient) * context.wrap_angle)
+
+    def validate_for_paths(self, num_paths: int) -> None:
+        """Validate the coefficient shape against the routed-path count."""
+        validate_loss_parameter_shape(self.coefficient, num_paths, "coefficient")
```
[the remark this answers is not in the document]
Michele-Martini: added validation methods for unphysical values

Michele-Martini:
The the old "wrap density" -> "path curvature" is basically the norm of the curvatures at s [rad/m], so that "wrap angle" is its line integral until s, being thus the "cumulative angle" until s [rad]

src/soromox/systems/pcs/pcs.py
Michele-Martini: length_gradient_density

docs/api/actuation/threadlike.md
```diff
@@ -165,14 +166,197 @@ coordinates = robot.actuator_coordinates(q)
...
+`ThreadlikeFriction` model, which is installed on a `ThreadlikeTransmission`
+through the `friction=` keyword of `ThreadlikeActuator` and
+`ThreadlikeImpedance`.
+The current implemented models are the following:
+
+| Model | Ratio | Parameter | Hosts |
+|---|---|---|---|
+| `ThreadlikeFriction.frictionless()` | \(1\) | none | all; the default |
+| `ThreadlikeFriction.exponential_length(...)` | \(e^{-k\ell}\) | `rate` \(k\), per metre | all |
+| `ThreadlikeFriction.capstan(...)` | \(e^{-\mu\Theta}\) | `coefficient` \(\mu\), Coulomb; `eps` \(\epsilon\), per metre | `PCS`, `PlanarPCS` |
```
mstoelzle, line 194: Why is the capstan not supported for `GVS` systems? If it were, we could simplify this table and remove the last column.

docs/api/actuation/threadlike.md
```diff
@@ -165,14 +166,197 @@ coordinates = robot.actuator_coordinates(q)
...
+
+Ideally, effort is transmitted undiminished along its route, from the motor
+until its attachment point. However, friction between the routing and the
+robot body is often non-negligible in real applications, and attenuates the
+effort along the path. It is possible to simulate this behavior via the
+`ThreadlikeFriction` model, which is installed on a `ThreadlikeTransmission`
+through the `friction=` keyword of `ThreadlikeActuator` and
+`ThreadlikeImpedance`.
+The current implemented models are the following:
+
```
mstoelzle, line 189: A reference here would be crucial

docs/api/actuation/threadlike.md
```diff
@@ -165,14 +166,197 @@ coordinates = robot.actuator_coordinates(q)
...
+\[
+k_p(q, s) = \lVert\omega \times \hat{u}\rVert,
+\qquad
+\eta(q, s) = e^{-\mu\Theta(q, s)},
+\quad
+\Theta(q, s) = \int_{s_0}^{s} k_p(q, s')\,\mathrm{d}s'
+\]
+
+Here \(\omega\) is the angular strain, \(\hat{u}\) the unit routed-path tangent,
+and \(\Theta\) the wrap angle accumulated from the anchor. Since
```
mstoelzle, line 231: There should be first a paragraph that explains what the wrap angle is.

docs/development/changelog.md
```diff
@@ -52,15 +65,41 @@ and include benchmark baseline and measurement context for performance claims.
   `soromox.rendering.actuator_visual_layers(...)`. Renderers continue to adapt
   standard `SoftRobot` systems automatically, while robot-defined hooks remain
   available for specialized visualization behavior.
+- Replaced `ThreadlikeImpedanceParams.routing` with a nested
+  `transmission` field, so a passive routed path now carries a
+  `ThreadlikeTransmission` like its articulated counterpart. The
+  `ThreadlikeImpedance` constructor keeps its `routing=` keyword; direct
+  construction of `ThreadlikeImpedanceParams` and `update_params(routing=...)`
```
mstoelzle, line 39: Check this change for consistency.

docs/development/changelog.md
```diff
@@ -52,15 +65,41 @@ and include benchmark baseline and measurement context for performance claims.
...
+  become `transmission=...`.
+- Corrected the planar routed-path offset sign, which used
+  `sigma_x + d * kappa` where `PlanarPCS` kinematics require
+  `sigma_x - d * kappa`. Planar routings must negate their `y` offsets to
```
mstoelzle, line 43: Double check if we want to change this convention. Reproduction of previous behavior is not a valid criterion for the convention design.

docs/development/changelog.md
```diff
@@ -52,15 +65,41 @@ and include benchmark baseline and measurement context for performance claims.
...
+  reproduce previous behavior, which is exactly equivalent when the routing
+  slope is zero.
+- Made host support for friction depend on the installed model rather than on
```
mstoelzle, line 46: > Made host support for friction depend on the installed model rather than on
  its value.

Too verbose

docs/development/changelog.md
```diff
@@ -52,15 +65,41 @@ and include benchmark baseline and measurement context for performance claims.
...
+  along arc length and does not, so it refuses a wrap-angle model at any
+  coefficient, zero included. Use `ThreadlikeFriction.exponential_length(...)`
+  there instead.
 
 ### Fixed
 
 - Fixed GVS abscissa-batched partial-cell poses and inertial Jacobians so
   rotational strain-basis values consistently honor
   `scale_rotational_basis_by_length` for every supported basis family.
+- Fixed planar threadlike path lengths and moment arms disagreeing with `PCS`
```
mstoelzle, line 70: Double check

src/soromox/actuation/friction.py
```diff
@@ -0,0 +1,325 @@
...
+from typing import Any, cast
+
+import equinox as eqx
+from jax import Array
+from jax import numpy as jnp
+
+from soromox.systems.params import BaseSystemParams
+
+
+def validate_friction_parameter_shape(value: Any, count: int, name: str) -> None:
```
mstoelzle, line 15: Check the naming of the function and if it is consistent with the rest of the package.

src/soromox/actuation/friction.py
```diff
@@ -0,0 +1,325 @@
...
+        ValueError: If the value is non-finite or negative.
+    """
+    array = jnp.asarray(value)
+    if not bool(jnp.all(jnp.isfinite(array))):
+        raise ValueError(f"{name} must be finite, got {array}.")
+    if bool(jnp.any(array < 0.0)):
+        raise ValueError(f"{name} must be non-negative, got {array}.")
+
+
+class ThreadlikeQuadratureContext(eqx.Module):
```
mstoelzle, line 53: Double check if this context structure is necessary.

src/soromox/actuation/friction.py
```diff
@@ -0,0 +1,325 @@
...
+
+    coefficient: Array = eqx.field(
+        default_factory=lambda: jnp.zeros((), dtype=jnp.float64)
+    )
+    eps: Array = eqx.field(default_factory=lambda: jnp.zeros((), dtype=jnp.float64))
+
+    def __check_init__(self) -> None:
+        self.validate_for_update()
+
+    def validate_structure_compatibility(self, num_paths: int) -> None:
```
mstoelzle, line 139: Double check the method name against the established convention in the paper.

src/soromox/actuation/friction.py
```diff
@@ -0,0 +1,325 @@
...
+            friction: Length-decay law.
+        """
+        return cls(
+            params=ExponentialLengthFrictionParams(rate=jnp.asarray(rate)),
+            ratio_fn=exponential_length_transmission_ratio,
+            requires_wrap_angle=False,
+            is_frictionless=False,
+        )
+
+    def transmission_ratio(self, context: ThreadlikeQuadratureContext) -> Array:
```
mstoelzle, line 267: Double check if `transmission_ratio` is the most suitable name.

src/soromox/actuation/threadlike.py
```diff
@@ -101,6 +107,15 @@ def _channel_array(value: Any, count: int, name: str) -> Array:
     return array
 
 
+def _normalize_friction(friction: Any) -> ThreadlikeFriction | None:
```
mstoelzle, line 110: Should helper function go into `friction.py`?

src/soromox/actuation/threadlike.py
```diff
@@ -393,14 +433,20 @@ def coordinates(self, robot, q: Array) -> Array:
...
 
     def moment_matrix(self, robot, q: Array) -> Array:
-        raw_matrix = robot._threadlike_moment_matrix(q, self.routing)
+        raw_matrix = robot._threadlike_moment_matrix(
+            q, self.routing, friction=self.friction
+        )
         return raw_matrix * self.params.coordinate_scale[None, :]
 
+    def kinematic_matrix(self, robot, q: Array) -> Array:
+        """Return the frictionless length gradient ``(dl/dq).T`` of the paths."""
```
mstoelzle, line 442: The docstring doesn't stick to the established convention in terms of style and format.

src/soromox/actuation/threadlike.py
```diff
@@ -393,14 +433,20 @@ def coordinates(self, robot, q: Array) -> Array:
         )
 
     def moment_matrix(self, robot, q: Array) -> Array:
```
mstoelzle, line 435: The docstring doesn't stick to the established convention in terms of style and format.

src/soromox/actuation/threadlike.py
```diff
@@ -724,60 +807,95 @@ def _from_params(
...
         """Return passive routed-path positions at backbone coordinate ``s``."""
-        return robot._threadlike_path_positions(q, s, self.routing)
+        return self._transmission.path_poses(robot, q, s)
 
     def _length_jacobian(self, robot, q: Array) -> Array:
-        return robot._threadlike_moment_matrix(q, self.routing).T
+        return self._transmission.kinematic_matrix(robot, q).T
+
+    def _transmission_matrix(self, robot, q: Array) -> Array:
+        return self._transmission.moment_matrix(robot, q)
```
mstoelzle, line 877: Check naming convention

src/soromox/actuation/threadlike.py
```diff
@@ -724,60 +807,95 @@ def _from_params(
...
 
     def elastic_force(self, robot, q: Array) -> Array:
+        """Return the conservative part of the passive path force."""
```
mstoelzle, line 880: The docstring doesn't stick to the established convention in terms of style and format.

src/soromox/actuation/threadlike.py
```diff
@@ -724,60 +807,95 @@ def _from_params(
...
         lengths = self.path_lengths(robot, q)
         jacobian = self._length_jacobian(robot, q)
         return jacobian.T @ (
             self.params.stiffness * (lengths - self.params.rest_length)
         )
 
-    def damping_matrix(self, robot, q: Array) -> Array:
+    def nonconservative_force(self, robot, q: Array) -> Array:
```
mstoelzle, line 887: The docstring doesn't stick to the established convention in terms of style and format.

src/soromox/actuation/threadlike.py
```diff
@@ -724,60 +807,95 @@ def _from_params(
...
+        """Return the part of the spring force that friction makes non-potential."""
+        lengths = self.path_lengths(robot, q)
+        transmission = self._transmission_matrix(robot, q)
         jacobian = self._length_jacobian(robot, q)
-        return jacobian.T @ (self.params.damping[:, None] * jacobian)
+        return (transmission - jacobian.T) @ (
+            self.params.stiffness * (lengths - self.params.rest_length)
+        )
+
+    def damping_matrix(self, robot, q: Array) -> Array:
```
mstoelzle, line 896: The docstring doesn't stick to the established convention in terms of style and format.

src/soromox/systems/gvs/core.py
```diff
@@ -5283,29 +5291,69 @@ def _threadlike_local_basis(
...
+            context: Node geometry.
+
+        Returns:
+            length_gradient_density: Shape ``(6,)``.
+        """
+        tangent = context.unit_tangent
+        local = jnp.hstack([so3.skew(context.offset) @ tangent, tangent])
+        return context.active * local
+
+    def _threadlike_moment_matrix(
```
mstoelzle, line 5909: Check naming convention.

src/soromox/systems/pcs/pcs.py
```diff
@@ -2671,26 +2679,167 @@ def _threadlike_local_basis(
...
+                derivative = jnp.append(routing.derivative(path_params, s), 1.0)
+                tangent = (derivative + se3.hat(strain) @ offset)[:-1]
+                unit = safe_normalize(tangent, eps=self.global_eps)
+                return safe_norm(jnp.cross(angular, unit))
+
+            return vmap(curvature_path)(params)
+
+        return vmap(curvature_segment)(jnp.arange(self.num_segments), midpoints)
+
+    def _threadlike_safe_wrap_angle(
```
mstoelzle, line 3206: Probably better to rename to `_threadlike_wrap_angle`.

src/soromox/systems/pcs/planar_pcs.py
```diff
@@ -2528,27 +2538,157 @@ def _threadlike_local_basis(
...
+
+        Args:
+            strains: Segment strains of shape ``(num_segments, 3)``.
+
+        Returns:
+            path_curvature: Shape ``(num_segments,)``, in radians per metre.
+        """
+        return jnp.abs(strains[:, 0])
+
+    def _threadlike_safe_wrap_angle(
```
mstoelzle, line 3070: Should probably also be renamed to `_threadlike_wrap_angle`.

src/soromox/systems/pcs/planar_pcs.py
```diff
@@ -2611,7 +2766,7 @@ def path_density(
                     offset = routing.offset(path_params, s)[1]
                     derivative = routing.derivative(path_params, s)[1]
                     axial = (
-                        strains[segment_index, 1] + offset * strains[segment_index, 0]
+                        strains[segment_index, 1] - offset * strains[segment_index, 0]
```
mstoelzle, line 3211: Please double check if the sign convention should be changed.

mstoelzle (changes requested):

mstoelzle:
Closing this PR because it is superseded by #203.

Compared with this proposal, #203:

- moves the planar threadlike sign correction into the separately merged #202;
- focuses the friction API on a continuum Capstan effort-loss model inspired by the cited discrete spacer-disk work, without claiming to implement discrete guides or a separate length-loss law;
- supports PlanarPCS, PCS, and GVS using the full spatial turn rate, including the analytical material-tangent derivative;
- specializes the PCS and PlanarPCS calculation to constant segment strain, where the tangent derivative reduces to `a' = omega × r' + r''`, while GVS retains the full varying-strain formula;
- replaces the quadrature-context object with explicit `accumulated_turn` arguments;
- separates the kinematic `coordinate_jacobian` from the effort-to-force `actuation_matrix`, which allows friction to affect force transmission without changing geometric actuator coordinates; and
- keeps friction active-only, leaving passive `ThreadlikeImpedance.routing` and its conservative spring-damper behavior unchanged.

#203 also contains the revised derivation, corrected paper reference, convention-aligned documentation, and expanded regression and autodiff-singularity coverage.

PR #192, opened 2026-08-31
Added tapered radii to ViserRenderer and fix lag in rendering

Johannap1:
Adds per-point tapered radii to the tube-mesh backbone rendering in ViserRenderer, plus a fix for rendering lag cause by JAX jit/threading interaction.

### DIFF
--- a/src/soromox/rendering/viser_renderer.py
+++ b/src/soromox/rendering/viser_renderer.py
@@ -49,6 +49,7 @@ from soromox.rendering.base import BaseSoftRobotRenderer
 from soromox.rendering.camera_config import CameraConfig
 from soromox.rendering.color_config import RendererColorConfig, ensure_rgba
 from soromox.rendering.video_encoding import FFmpegVideoWriter, VideoEncodingConfig
+from soromox.systems.components import CrossSectionGeometry
 from soromox.systems.soft_robot import SoftRobot
 
 # =============================================================================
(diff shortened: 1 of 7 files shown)


src/soromox/rendering/viser_renderer.py
```diff
@@ -417,6 +427,17 @@ def _get_robot_radius(self) -> float:
             return float(r)
         return 0.02  # Default radius
 
+    def _get_robot_radii(self):
+        """Get per-point robot radii for tapered visualization."""
+        try:
```
mstoelzle: Please avoid such try / except clauses.

src/soromox/rendering/viser_renderer.py
```diff
@@ -417,6 +427,17 @@ def _get_robot_radius(self) -> float:
...
+            s_ps = jnp.linspace(0.0, self.L_max, self.num_points)
+            q_dummy = jnp.zeros(self.robot.num_dofs)
+            return jnp.array(
+                [self.robot.cross_section_geometry(q_dummy, s)[1][0] for s in s_ps]
```
mstoelzle: Use vmap for move efficient looping over `s_ps`.

src/soromox/rendering/viser_renderer.py
```diff
@@ -391,6 +400,7 @@ def __init__(
 
         # Get robot radius for sizing
         self._robot_radius = self._get_robot_radius()
```
mstoelzle: We should now have two functions (`_robot_radius` and `_robot_radii`) for the same purpose.

src/soromox/rendering/viser_renderer.py
```diff
@@ -391,6 +400,7 @@ def __init__(
...
+        self._robot_radii = self._get_robot_radii()
```
mstoelzle, line 403: I think we should generalize this now to any of the three supported cross sectional geometries (circular, ellipse, and rectangular).

src/soromox/rendering/viser_renderer.py
mstoelzle: `robot.cross_section_geometry(q,s)` is an officially supported interface of the `SoftRobot` base class. Therefore, the try/except should not be necessary as we can rely on this method to be present.

mstoelzle (changes requested):
Thanks, @Johannap1, for spotting this outdated code and the lack of support for properly scaling the cylinders/tube when the radii of the base and tip of the cylinder are not the same.

Beyond some smaller issues, I think we should take this opportunity and directly add support for the more general cross section geometries beyond circular (i.e., also ellipse and rectangular) to avoid similar issues for non-circular (GVS) links.

mstoelzle:
Hi @Johannap1, while addressing the Viser tapering issue, could you please update Open3DRenderer as well? Its swept path currently creates one set of cross-section ring offsets from sections[p] and reuses it for both endpoints of the edge from p to p+1, so sections[p+1] is ignored. This means tapered circular, elliptical, and rectangular sections remain piecewise constant and discontinuous in Open3D. The swept helper should accept the geometry and dimensions, or ring offsets, for both endpoints, with regression tests using different endpoint sizes. Open3D already supports the three cross-section types otherwise.

mstoelzle:
Superseded by #200, which retains the Viser tapering work and generalizes both Viser and Open3D to shared circular, elliptical, and rectangular cross-sections, abscissa-varying profiles, link-local interfaces, and consistent discrete markers. Closing this PR in favor of #200.

PR #197, opened 2026-09-01
Fix model-parameter gradients across custom JVPs

mstoelzle:
## Summary

- preserve derivatives with respect to dynamic model leaves across the public
  forward-kinematics, Jacobian, and energy custom JVPs while retaining their
  optimized state-derivative paths
- add compiled regressions for fixed-base-pose and gravity optimization,
  including numerical agreement with direct autodiff and repeated values
  without retracing
- document `with_fixed_base_pose(...)`, joint parameter/mounting optimization,
  and the fixed-base versus floating-base ownership of the pose; add an
  Unreleased changelog entry under Fixed

## Regression

#184 moved a fixed mounting pose out of the typed parameter PyTree introduced
around #156. The pose remained a dynamic, differentiable model leaf, so its
value could change inside a compiled function without recompilation. However,
the forward-kinematics custom JVP preserved only configuration and abscissa
tangents and discarded the incoming model tangent, producing a zero gradient
with respect to the mounting pose.

The audit found the same dropped-model-tangent pattern in the shared
fixed-base `jacobian`, `kinetic_energy`, `gravitational_energy`,
`elastic_energy`, `potential_energy`, and `total_energy` custom JVPs. Each rule
now adds the model-only derivative to its established optimized state tangent.
The accelerated kinematics, dynamics, forward-dynamics, and actuation adapters
already differentiated their protected JAX implementations with respect to the
complete model and required no analogous change.

The GVS regression evaluates `jax.value_and_grad` inside `eqx.filter_jit` for
two fixed mounting poses and verifies one trace, finite nonzero gradients, and
agreement with direct autodiff. A Pendulum regression applies the same contract
to gravity. A parameterized base-class regression independently checks the
Jacobian and all five energy boundaries.

## API design

`fixed_base_pose` remains outside the typed dynamic parameter struct because it
is a model leaf for fixed robots but runtime configuration state for floating
robots. Keeping it separate lets one parameter PyTree be reused across
mountings and avoids making that tree depend on a fixed-versus-floating choice.

Gravity remains in the parameter PyTree: it has the same stable world-frame
environment semantics for fixed and floating systems. Moving it to a separate
model leaf would not improve JIT or differentiation and would fragment the
parameter-update interface.

## Validation

- `127 passed` across the complete default-model and Pendulum suites, execution
  and compiled model-tangent checks, and cross-family PCS, PlanarPCS, GVS,
  PlanarHSA, and articulated energy/Jacobian autodiff checks
- an additional `45 passed` across broader selected GVS, Pendulum,
  ArticulatedSoftRobot, and PlanarHSA kinematics checks
- Ruff lint and format checks on every changed Python file
- Python byte-compilation checks on every changed Python file
- Zensical documentation build
- scoped `git diff --check`

### DIFF
--- a/docs/user-guide/parameters-and-optimization.md
+++ b/docs/user-guide/parameters-and-optimization.md
@@ -206,6 +206,7 @@ variables without reconstructing the model or recompiling for each value:
 ```python
 import equinox as eqx
 import jax
+import jax.numpy as jnp
 
 target_position = jnp.array([0.1, -0.05, 0.25])
 
--- a/docs/user-guide/parameters-and-optimization.md
+++ b/docs/user-guide/parameters-and-optimization.md
@@ -16,6 +16,7 @@ damping, and isotropic materials, see
 | Construct a robot from physical descriptions | `LinkSpec`, `JointSpec`, `GVSSegment` |
 | Replace runtime numeric values | `update_params`, `update_link_params`, `update_joint_params` |
 | Replace several nested components together | nested `.replace(...)` followed by `with_params(...)` |
+| Change or optimize a fixed mounting pose | `with_fixed_base_pose(...)` |
 | Identify isotropic material properties | caller-owned `IsotropicMaterialParams` and `with_isotropic_material(...)` |
 | Fit anisotropic or coupled mechanics | optimize `params.link.stiffness` and `params.link.damping` directly |
 | Change basis order, segment count, or padding | reconstruct with a new structure |
@@ -153,6 +154,22 @@ Top-level values use `update_params`:
 pcs = pcs.update_params(gravity=jnp.zeros(3))
 ```
 
+A fixed mounting pose is a dynamic model leaf rather than a field in the
+physical parameter PyTree. Update it independently with
+`with_fixed_base_pose(...)`:
+
+```python
+candidate_base_pose = pcs.fixed_base_pose.at[4].add(0.01)
+mounted_pcs = pcs.with_fixed_base_pose(candidate_base_pose)
+```
+
+This keeps one physical parameter PyTree reusable across different mountings
+and between fixed and floating models without making the fixed pose static or
+nondifferentiable.
+
+Floating systems instead carry their base pose in the runtime configuration
+and reject `with_fixed_base_pose(...)`.
+
 Updates may change numeric values but not static layout. Changing the number of
 links, GVS basis orders, active strain selectors, joint types, matrix padding,
 or quadrature padding requires reconstruction.
@@ -182,6 +199,32 @@ candidate_params = pcs.params.replace(
 )
 candidate = energy(candidate_params, q)
 ```
+
+Fixed mounting poses can likewise be supplied as dynamic optimization
+variables without reconstructing the model or recompiling for each value:
+
+```python
+import jax
+
+target_position = jnp.array([0.1, -0.05, 0.25])
+
+@eqx.filter_jit
+def mounted_loss_and_grad(base_pose, configuration):
+    def loss(pose):
+        candidate = pcs.with_fixed_base_pose(pose)
+        tip_pose = candidate.forward_kinematics(configuration, candidate.length)
+        return jnp.sum((tip_pose[:3, 3] - target_position) ** 2)
+
+    return jax.value_and_grad(loss)(base_pose)
+
+loss, base_pose_gradient = mounted_loss_and_grad(pcs.fixed_base_pose, q)
+```
+
+Repeated calls share one compiled executable while the pose retains the same
+shape and dtype. Use a constrained rotation parameterization or normalize the
+quaternion when optimizing a spatial mounting. To identify physical parameters
+and mounting jointly, pass both `params` and `base_pose` as dynamic arguments
+and chain `with_params(params).with_fixed_base_pose(base_pose)`.
 
 The two calls share one compiled executable when every parameter leaf keeps
 the same PyTree position, shape, and dtype. Numeric system, environment,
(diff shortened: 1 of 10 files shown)


src/soromox/execution/transforms.py
```diff
@@ -310,16 +310,20 @@ def scalar_jvp(
...
+                q, s, qd, sd, model_tangent=model_tangent
+            )
         if operation == "jacobian":
             return eqx.filter_jvp(
                 lambda model_, q_, s_: model_._absolute_inertial_jacobian(q_, s_),
                 primals,
                 tangents,
             )
-        pose, pose_tangent = model._forward_kinematics_jvp(q, s, qd, sd)
+        pose, pose_tangent = model._forward_kinematics_jvp(
```
mstoelzle, line 320: Why are there two separate dispatches for `model._forward_kinematics_jvp` even though the call is the same?

src/soromox/systems/soft_robot.py
```diff
@@ -1001,27 +1026,38 @@ def _forward_kinematics_jvp(
...
+                    lambda q_, s_: forward_kinematics(q_, s_),
+                    (q, s),
+                    (qd, sd),
+                )
+        elif sd is None:
             pose = forward_kinematics(q, s)
-            return pose, jnp.zeros_like(pose)
+            pose_tangent = jnp.zeros_like(pose)
+        elif self.floating_base:
+            pose, pose_tangent = jvp(lambda s_: forward_kinematics(q, s_), (s,), (sd,))
```
mstoelzle, line 1057: Is this not a circular call? And does this use autodiff unnecessarily? We always prefer analytical gradients if available.

src/soromox/systems/soft_robot.py
```diff
@@ -972,6 +973,24 @@ def _forward_kinematics(self, q: Array, s: Array) -> Array:
             f"{type(self).__name__} must implement _forward_kinematics."
         )
 
+    def _add_model_jvp_tangent(
```
mstoelzle, line 976: Is this the best name?

tests/execution/test_transforms.py
```diff
@@ -96,6 +99,50 @@ def _solve_inertia(self, inertia: Array, rhs: Array) -> Array:
...
+        """Differentiate the probe's model and state inputs."""
+
+        qd = jnp.zeros_like(q) if qd is None else qd
+        sd = jnp.zeros_like(s) if sd is None else sd
+        pose, pose_tangent = jax.jvp(
+            lambda q_, s_: self._absolute_forward_kinematics(q_, s_),
+            (q, s),
+            (qd, sd),
+        )
+        if model_tangent is not None and jax.tree.leaves(model_tangent):
```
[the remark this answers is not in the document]
mstoelzle: Addressed in 9ee05e1f. The execution-test probe now uses the same explicit inexact-array tangent check as the production helper.

src/soromox/systems/soft_robot.py
```diff
@@ -1001,27 +1007,41 @@ def _forward_kinematics_jvp(
...
-        if self.floating_base:
-            return jvp(lambda s_: forward_kinematics(q, s_), (s,), (sd,))
+            pose_tangent = jnp.zeros_like(pose)
+        elif self.floating_base:
+            pose, pose_tangent = jvp(lambda s_: forward_kinematics(q, s_), (s,), (sd,))
+        else:
+            pose, poses = self.forward_kinematics_and_arc_length_derivative(q, s)
+            pose_tangent = poses * sd
+
+        if model_tangent is not None and jax.tree.leaves(model_tangent):
```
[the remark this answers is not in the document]
mstoelzle: Addressed in 9ee05e1f. The shared helper now requires at least one inexact-array tangent leaf before invoking filter_jvp, and a regression verifies that an inactive tangent tree skips model evaluation.

src/soromox/execution/transforms.py
mstoelzle: Good point. Addressed in 9ee05e1f by hoisting the common forward-kinematics JVP call after the Jacobian-only branch. Pose returns it directly, while the combined operation reuses it before evaluating the Jacobian.

src/soromox/systems/soft_robot.py
mstoelzle: Renamed in 9ee05e1f to _add_model_jvp_contribution, which describes that the helper adds the model-only contribution to an already established output tangent.

docs/user-guide/parameters-and-optimization.md
```diff
@@ -183,6 +200,32 @@ candidate_params = pcs.params.replace(
 candidate = energy(candidate_params, q)
 ```
 
+Fixed mounting poses can likewise be supplied as dynamic optimization
+variables without reconstructing the model or recompiling for each value:
+
+```python
+import jax
+
```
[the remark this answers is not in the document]
mstoelzle: Addressed in 9ee05e1f by adding the missing Equinox import so the example is self-contained.

src/soromox/systems/soft_robot.py
mstoelzle: This is not circular: _absolute_forward_kinematics is a protected primal path that reaches _forward_kinematics, not the public custom-JVP wrapper. The configuration JVP deliberately differentiates the stored pose representation; the analytical Jacobian instead returns an inertial velocity, and the previous conversion path duplicated pose-specific logic without a measured benefit. The arc-length-only case still uses the analytical fused hook. I retained this established main-branch strategy and clarified the rationale in the docstring in 9ee05e1f.

src/soromox/systems/soft_robot.py
mstoelzle: Thanks for clarifying. I restored the name to `_add_model_jvp_tangent` in bb905649. I kept `model` because the tangent covers the complete robot PyTree (including `fixed_base_pose`), rather than only the typed parameters.

PR #200, opened 2026-09-01
Generalize Open3D and Viser cross-section rendering

mstoelzle:
## Summary

This PR generalizes Open3D and Viser rendering from circular radii to shared,
renderer-neutral cross-section geometry. It supersedes #192 and includes its
Viser tapering fix together with the corresponding Open3D fix.

- preserve circular, elliptical, and rectangular cross-sections in swept mode
- evaluate constant and abscissa-varying dimensions at every backbone station,
  including `LinearProfile` tapers
- build both backends from the same ordered material-frame contours and
  link-local loft topology
- render consistent discrete proxies in both backends: spheres for circles,
  boxes for rectangles, and ellipsoids for ellipses
- expose a contour-registration boundary for adding future cross-section
  families without backend-specific mesh construction
- remove renderer dependence on cross-section bounding radii

### Shared cross-section boundary

`soromox.rendering.cross_sections` owns the backend-neutral geometry contract:

- `CrossSection` stores one geometry family and its dimensions at one abscissa
- built-in contour builders generate ordered, implicitly closed material-frame
  contours for circles, ellipses, and rectangles
- `register_cross_section_contour(...)` extends swept rendering in both
  backends with one builder
- `loft_cross_sections(...)` and `loft_cross_section_contours(...)` produce the
  shared vertices and triangle topology consumed by Open3D and Viser
- `DiscreteCrossSectionMarker` defines the common sphere/box/ellipsoid proxy
  and material-frame alignment used by discrete mode

The contour is a geometric boundary, not a circular "ring" abstraction.
Rectangles keep explicit corners; higher resolution only adds points along
their straight edges.

### Varying profiles and link interfaces

Both swept renderers evaluate `robot.cross_section_geometry(q, s)` at each
backbone station and loft the two endpoint contours of every link-local edge.
Open3D therefore no longer reuses the first endpoint dimensions at the second
endpoint, which fixes its piecewise-constant tapered profiles.

Each link owns one-sided base and tip stations. The renderers cap both sides of
an internal interface and do not loft directly across links, so a rectangular
link followed by an elliptical or circular link remains a deliberate shape
transition rather than a short morphing funnel.

### Swept versus discrete

`num_points` controls longitudinal backbone stations. In swept mode, those
stations form continuous link-local surfaces; in discrete mode, they place
independent 3D markers that may overlap or leave gaps.

`cross_section_resolution` controls the transverse vertices of swept contours
and is ignored in discrete mode. Discrete boxes and ellipsoids preserve the
local cross-section dimensions and material-frame orientation, with finite
longitudinal depth used only as a marker proxy.

## Breaking change

The geometry-specific renderer options have been replaced by one shared name:

- Viser: `cylinder_sections` -> `cross_section_resolution`
- Open3D: `tube_resolution` -> `cross_section_resolution`

Production code, paper renderers, examples, tests, API documentation, and the
changelog are migrated together. No compatibility aliases are retained.

## Visual validation

All four frames use the same deformed and twisted three-link GVS robot with 63
backbone stations:

1. a rectangular base link with linearly tapered height and constant width;
2. an elliptical middle link with constant semi-major axis and linearly tapered
   semi-minor axis;
3. a circular tip link with linearly tapered radius.

The Open3D frames were captured through its native interactive `Visualizer`.
The Viser frames were captured from a connected Viser WebGL browser client.

| Backend | Swept | Discrete |
| --- | --- | --- |
| Open3D | Open3D swept cross-sections | Open3D discrete cross-section markers |
| Viser | Viser swept cross-sections | Viser discrete cross-section markers |

## Validation

- full rendering regression suite: `126 passed`
- affected paper renderer tests excluding two Git-LFS-data-dependent committed
  pose checks: `25 passed, 2 deselected`
- `git diff --check`: passed
- native Open3D swept/discrete frames: generated and visually inspected
- browser-connected Viser swept/discrete frames: generated and visually
  inspected

The two deselected Section Vd checks require
`paper_results/secVd_control_gain_optimization/data/collocated/optimization_results.npz`,
which is present as an unmaterialized Git LFS pointer in this checkout; the
failure is unrelated to rendering.


docs/api/rendering/renderers.md
```diff
@@ -28,6 +28,56 @@ BaseSoftRobotRenderer (abstract base)
 | Base plate | `base_plate_radius_scale` and `base_plate_thickness` configure the base geometry | Open3D and Viser; Matplotlib draws a lightweight base marker |
 | Ground plane | `show_ground_plane` and `ground_plane_size` configure a base-aligned reference plane; its colors come from `RendererColorConfig` | Matplotlib, Open3D, and Viser |
 
+### Swept Cross-Sections
```
mstoelzle: Please carefully evaluate if (a) this is still the most suitable heading, (b) if the content, extent and details in this section are appropiately calibrated against the rest of the docs page.

src/soromox/rendering/viser_renderer.py
```diff
@@ -389,8 +417,26 @@ def __init__(
         self._backbone_cast_shadow = backbone_cast_shadow
         self._sphere_cast_shadow = sphere_cast_shadow
 
-        # Get robot radius for sizing
-        self._robot_radius = self._get_robot_radius()
+        # Cross-sections are geometry metadata and are independent of robot state.
+        s_points = np.asarray(self._backbone_abscissae, dtype=np.float64)
+        q_reference = jnp.zeros(self.robot.num_coordinates)
+        self._cross_sections = evaluate_cross_sections(
```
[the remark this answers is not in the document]
mstoelzle: Thanks. I checked this against the renderer boundary before changing it. BaseSoftRobotRenderer accepts a SoftRobot, and num_coordinates is the public configuration-size property on that contract. A fallback would silently broaden the API to duck-typed adapters and could supply an invalid reference configuration when q is not ignored. This PR intentionally favors the explicit contract over compatibility with incomplete stubs, so I am keeping the direct num_coordinates access; the test dummy now implements the declared contract.

docs/api/rendering/renderers.md
mstoelzle: Agreed. The heading had become inaccurate once the section also described discrete markers, and the gallery plus implementation narrative was too detailed for this API-reference page. In c1eb3849 I renamed it to Cross-Section Geometry and reduced it to a compact swept-versus-discrete comparison, the two resolution controls, zero-length-link behavior, and the extension hook. The four native rendering frames remain in the PR description as validation evidence rather than in the API reference.

src/soromox/rendering/cross_sections.py
```diff
@@ -0,0 +1,452 @@
...
+    robot: Any,
+    q: Any,
+    abscissae: Iterable[float],
+) -> tuple[CrossSection, ...]:
+    """Evaluate a robot's cross-section descriptor at each backbone position."""
+    sections = []
+    for abscissa in np.asarray(tuple(abscissae), dtype=np.float64).reshape(-1):
+        geometry, dimensions = robot.cross_section_geometry(q, abscissa)
+        sections.append(CrossSection(geometry, dimensions))
+    return tuple(sections)
```
ivrolan: Now that robot.cross_section_geometry() uses lax.switch(), this evaluation for rendering should also use jit if possible.

Something like:
```
stations = np.asarray(tuple(abscissae), dtype=np.float64).reshape(-1)

geometries, dimensions = jax.vmap(robot.cross_section_geometry, in_axes=(None, 0))(q, jnp.asarray(stations))
```

Otherwise, I think the lax.switch() and related vmap-compatible functions in cross_section_geometry() would just introduce lag for free because of the compiling time. Did you feel any difference w.r.t. the previous version?

mstoelzle: Agreed—the scalar loop was materially expensive. On my local CPU/default-JAX setup with a mixed-profile three-link GVS and 63 stations, it measured about 1.8 s warm, versus about 8 ms for a persistent plain vmap and 0.14 ms for a persistent compiled vmap before constructing the Python CrossSection objects. I implemented a module-level eqx.filter_jit + jax.vmap path; the production-shaped evaluator measured about 63 ms cold and 0.30 ms warm including CrossSection construction. Since every renderer model is a SoftRobot, the follow-up removes the scalar compatibility path entirely and types the evaluator accordingly. The small array-only helper remains a separate, stable JIT boundary so JAX can reuse its compiled executable: the outer function converts traced arrays into NumPy-backed CrossSection objects and therefore cannot itself be JIT-compiled. Test doubles now honor the vmap-compatible model contract. The complete rendering suite passes (130 tests), including the mixed rectangular/elliptical/circular LinearProfile regression.

Johannap1 (approved):
Looking good :)

PR #203, opened 2026-09-03
Add continuum Capstan friction to threadlike actuation

mstoelzle:
## Summary

This PR supersedes #190 with a continuum formulation of Capstan effort loss
for fixed material-frame threadlike routing. It is based on
`feature/tendon_friction_model_pcs` and targets `main`.

It depends on #202 for the standalone `PlanarPCS` threadlike sign correction.
That fix was split out so it can merge independently.

- apply a Capstan effort ratio continuously at the host quadrature nodes for
  `PlanarPCS`, `PCS`, and `GVS`
- derive routed-path turn from the spatial unit tangent, including analytical
  strain and material-routing derivatives
- keep friction active-only; passive `ThreadlikeImpedance` continues to own
  routing geometry and conservative spring-damper mechanics
- remove `ThreadlikeQuadratureContext` in favor of explicit
  `accumulated_turn` arguments for friction laws
- distinguish the actuator-coordinate Jacobian from the effort-to-force
  actuation matrix throughout the package
- keep frictional threadlike actuation on JAX while retaining Warp execution
  for built-in frictionless linear routing

## Relation to the cited paper

The implementation is inspired by the spacer-disk tendon model in
[Feliu-Talegon et al. (2025)](https://doi.org/10.1109/TMECH.2025.3581774).
That paper applies the established Capstan relation at discrete guide disks,

$$
T_i=T_{i-1}e^{-\mu\varphi_i},
\qquad
T_n=T_0\exp\!\left(-\mu\sum_i\varphi_i\right).
$$

SoRoMoX does not model those disks. Instead, it adapts the same exponential
effort-loss relation to distributed contact along an embedded continuum path.
The current API has no discrete guide locations, radii, or reactions and does
not support a discrete-guide mode.

## Continuum derivation

Let:

- $s$ be the backbone abscissa in the undeformed configuration and $s_0$ the routed path's anchor;
- $R(s)\in SO(3)$ and $p(s)\in\mathbb{R}^3$ be the backbone material
  orientation and position;
- $\xi(s)=(\omega(s),v(s))$ be angular and linear Cosserat strain;
- $r(s)\in\mathbb{R}^3$ be the path offset expressed in the material frame;
- a prime denote differentiation with respect to $s$; and
- $[\omega]_\times$ denote the skew matrix satisfying
  $[\omega]_\times r=\omega\times r$.

The Cosserat kinematics and spatial routed path are

$$
R'=R[\omega]_\times,
\qquad
p'=Rv,
\qquad
x=p+Rr.
$$

Differentiating $x$ gives

$$
x'=Ra,
\qquad
a=v+\omega\times r+r'.
$$

Here $a$ is the path tangent expressed in the material frame. Define its norm
$\nu=\lVert a\rVert$, its material unit tangent $u=a/\nu$, and its spatial
unit tangent $\hat t=Ru$. Analytical differentiation gives

$$
a'=v'+\omega'\times r+\omega\times r'+r'',
$$

and therefore

$$
u'=\frac{(I-uu^T)a'}{\nu}.
$$

Using $R'=R[\omega]_\times$,

$$
\frac{d\hat t}{ds}=R(\omega\times u+u').
$$

Because rotation preserves the Euclidean norm, the unsigned spatial turn rate
is

$$
\rho(s)=\left\lVert\omega\times u+u'\right\rVert.
$$

### Constant-strain specialization

PCS and PlanarPCS assume that $\xi=(\omega,v)$ is constant within each
segment. Consequently, $\omega'=v'=0$ there and the general material-tangent
derivative specializes to

$$
a'=\omega\times r'+r''.
$$

Both constant-strain hosts evaluate this reduced expression directly. The
routing may still vary with the backbone abscissa, so $r'$ and $r''$ can make
$u'$ nonzero even though the segment strain is constant. GVS does not make this
simplification: it retains
$a'=v'+\omega'\times r+\omega\times r'+r''$ and obtains $\omega'$ and
$v'$ analytically from the derivative of the installed strain basis, including
the normalized-coordinate-to-backbone-abscissa scaling. Production code uses
these analytical formulas; test-only JVPs independently verify the general and
constant-strain implementations.

Accumulated turn is

$$
\Theta(s)=\int_{s_0}^{s}\rho(\sigma)\,d\sigma.
$$

The host's quadrature integrates this quantity over complete and partial
segments. If a piecewise host discretization introduces a genuine one-sided
tangent jump, its unsigned angle is included at that interface. These
interfaces are part of the continuum discretization; they are not configurable
friction disks.

The surviving anchor-effort fraction is

$$
\eta(s)=\frac{e(s)}{e(s_0)}=e^{-\mu\Theta(s)},
$$

where $\mu\ge 0$ is the Capstan friction coefficient. The public helper is
named `capstan_effort_ratio` because threadlike actuation also represents push
rods, simplified muscles, and pressure chambers, even though tendons are the
primary application.

For the local spatial strain, the unscaled routed-length gradient density is

$$
g_\xi=\begin{bmatrix}r\times u\\u\end{bmatrix}.
$$

If $B(s)$ maps generalized coordinates to strain and $c$ is the modality's
signed or scaled work-coordinate factor, the active effort-to-force map is

$$
A_\mu(q)=c\int_{s_0}^{s_1}
\eta(q,s)B(s)^Tg_\xi(q,s)\,ds.
$$

Applying $\eta$ to every local contribution is the continuum adaptation; a
single post-integration scalar would not represent distributed loss.

## Coordinate and force-map convention

`Transmission` now exposes two explicit matrices:

$$
\dot y_a=J_a(q)\dot q,
\qquad
\tau_u=A(q)e,
$$

where:

- `coordinate_jacobian(...)` returns
  $J_a=\partial y_a/\partial q$ with shape
  `(num_channels, num_velocities)`; and
- `actuation_matrix(...)` returns $A$ with shape
  `(num_velocities, num_channels)`.

The robot aggregate is `actuator_coordinate_jacobian(...)`.

For an ideal transmission, virtual work gives $A=J_a^T$. Friction leaves path
coordinates and velocities geometric while changing the force map, so in
general $A\ne J_a^T$. `ActuationSpaceDynamics` consequently uses $J_a$ for its
coordinate transformation and transforms the physical force map as
$A_y=J^{-T}A$. Existing collocated actuation-space controllers still assume
the ideal $A=J_a^T$ case; this limitation is now explicit in the documentation.

This is a breaking replacement of `Transmission.moment_matrix(...)`; no
compatibility alias is retained.

## Friction API and scope

`ThreadlikeFriction.capstan(coefficient=...)` installs the continuum model.
Custom laws receive explicit `(params, accumulated_turn)` arguments. Custom
routing laws provide analytical `offset_fn`, `derivative_fn`, and
`second_derivative_fn`; production code does not use autodiff to recover the
material-tangent derivative.

Friction is intentionally limited to active transmissions. Passive
`ThreadlikeImpedance` keeps its `.routing` attribute and conservative
spring-damper force, damping, and energy contracts. The implementation does not
add length-proportional exponential loss, sticking, backdriving direction,
hysteresis, slack, or discrete guides.

## Validation

- full non-system suite before the final boundary-turn regression was added:
  `1384 passed, 26 skipped`
- review-targeted threadlike and actuation-space regression suite:
  `228 passed`
- Warp threadlike actuation suite: `50 passed, 14 skipped`
- targeted PlanarPCS/PCS coherence, floating-base, and GVS public-actuation
  regressions: `12 passed`
- `git diff --check`: passed

### DIFF
--- a/docs/api/actuation/threadlike.md
+++ b/docs/api/actuation/threadlike.md
@@ -6,11 +6,18 @@ contractile muscles, and pressure chambers represented by an equivalent axial
 volume coordinate. A path may span any contiguous inclusive range of body
 segments.
 
-Following the general routed-path formulation of
-[Renda et al. (2022)](https://doi.org/10.1109/LRA.2022.3183248), a routing is a
-material-frame offset from the backbone. `PCS`, `PlanarPCS`, and `GVS` combine
-that routing with their native strain kinematics to evaluate path length,
-coordinate velocity, and generalized actuator force.
+!!! info "Threadlike-routing reference"
+    The routing geometry follows:
+
+    > Renda, F., Armanini, C., Mathew, A. T., & Boyer, F. (2022).
+    > *Geometrically-Exact Inverse Kinematic Control of Soft Manipulators With
+    > General Threadlike Actuators' Routing*. IEEE Robotics and Automation
+    > Letters, 7(3), 7311–7318.
+    > [https://doi.org/10.1109/LRA.2022.3183248](https://doi.org/10.1109/LRA.2022.3183248)
+
+A routing is a material-frame offset from the backbone. `PCS`, `PlanarPCS`, and
+`GVS` combine that routing with their native strain kinematics to evaluate path
+length, coordinate velocity, and generalized actuator force.
 
 The model assumes fixed material-frame routing, one scalar coordinate and
 effort per path, and an optional static effort loss along active paths. It does
@@ -18,6 +25,8 @@ not model configuration-dependent rerouting, slack, actuator dynamics, or
 changes in pressure-chamber cross-section.
 
 ## Routing
+
+### Linear routing
 
 `ThreadlikeRouting` stores a vectorized path family. Linear paths use
 
@@ -78,25 +87,53 @@ class SinusoidalRoutingParams(BaseThreadlikeRoutingParams):
     end_segment_index: tuple[int, ...] = eqx.field(static=True)
 
     @property
-    def num_paths(self):
-        """Return the number of paths in this parameter batch."""
+    def num_paths(self) -> int:
+        """Return the number of paths in this parameter batch.
+
+        Returns:
+            Number of routed paths.
+        """
         return self.amplitude.shape[0]
 
 
-def offset(params, s):
-    """Return material-frame sinusoidal routing offsets at ``s``."""
+def offset(params: SinusoidalRoutingParams, s: jax.Array) -> jax.Array:
+    """Return material-frame sinusoidal routing offsets.
+
+    Args:
+        params: Batched sinusoidal routing parameters.
+        s: Scalar backbone abscissa.
+
+    Returns:
+        Material-frame offsets with shape ``(num_paths, 3)``.
+    """
     y = params.amplitude * jnp.sin(params.frequency * s)
     return jnp.stack((jnp.zeros_like(y), y, jnp.zeros_like(y)), axis=-1)
 
 
-def derivative(params, s):
-    """Return the first arc-length derivative of ``offset``."""
+def derivative(params: SinusoidalRoutingParams, s: jax.Array) -> jax.Array:
+    """Return the first derivative of the routing offset.
+
+    Args:
+        params: Batched sinusoidal routing parameters.
+        s: Scalar backbone abscissa.
+
+    Returns:
+        First derivatives with shape ``(num_paths, 3)``.
+    """
     y_s = params.amplitude * params.frequency * jnp.cos(params.frequency * s)
     return jnp.stack((jnp.zeros_like(y_s), y_s, jnp.zeros_like(y_s)), axis=-1)
 
 
-def second_derivative(params, s):
-    """Return the second arc-length derivative of ``offset``."""
+def second_derivative(params: SinusoidalRoutingParams, s: jax.Array) -> jax.Array:
+    """Return the second derivative of the routing offset.
+
+    Args:
+        params: Batched sinusoidal routing parameters.
+        s: Scalar backbone abscissa.
+
+    Returns:
+        Second derivatives with shape ``(num_paths, 3)``.
+    """
     y_ss = -(params.frequency**2) * params.amplitude * jnp.sin(
         params.frequency * s
     )
@@ -159,9 +196,6 @@ coordinate_jacobian = robot.actuator_coordinate_jacobian(q)
 actuation_matrix = robot.actuation_matrix(q)
 ```
 
-The singular `coordinate` in `coordinate_jacobian` follows the usual compound
-noun: it is the Jacobian of the vector returned by `actuator_coordinates`.
-The robot-level prefix is plural because all installed actuators are stacked.
 The two matrix contracts are
 
 \[
@@ -179,14 +213,14 @@ coordinates or velocities.
 ## Continuum Capstan friction
 
 `ThreadlikeFriction.capstan(coefficient=mu)` attenuates active threadlike effort
-along increasing backbone coordinate, starting at each path's configured
+along increasing backbone abscissa, starting at each path's configured
 anchor. It is available for `PCS`, `PlanarPCS`, and `GVS`. The default
 `ThreadlikeFriction.frictionless()` returns unit effort ratio.
 
 ### Path-turn derivation
 
-Let \(s\) be undeformed backbone arc length, \(R(s)\in SO(3)\) the material
-orientation, \(p(s)\in\mathbb{R}^3\) the backbone position, and
+Let \(s\) be the backbone abscissa in the undeformed configuration,
+\(R(s)\in SO(3)\) the material orientation, \(p(s)\in\mathbb{R}^3\) the backbone position, and
 \(\xi=(\omega,v)\) the angular and linear Cosserat strain. Then
 
 \[
@@ -224,9 +258,10 @@ Consequently,
 `turn_rate` is \(\rho\), in radians per metre of undeformed backbone. It
 includes both material-frame rotation and change of the tangent inside that
 frame. Omitting \(u'\) is valid only when the normalized material tangent is
-constant. Inside a PCS segment, \(\omega'=v'=0\), but routing derivatives can
-still make \(u'\ne0\). GVS computes \(\omega'\) and \(v'\) analytically from
-the derivative of its strain basis.
+constant. Inside a PCS segment, \(\omega'=v'=0\), so SoRoMoX uses the
+constant-strain specialization \(a'=\omega\times r'+r''\). Routing derivatives
+can still make \(u'\ne0\). GVS retains the full expression and computes
+\(\omega'\) and \(v'\) analytically from the derivative of its strain basis.
 
 The `accumulated_turn` from anchor \(s_0\) is
 
@@ -241,10 +276,17 @@ unsigned angle at the interface.
 ### Relation to the cited discrete model
 
 The exponential Capstan relation is established for ropes, belts, and tendons
-sliding over contact. SoRoMoX takes inspiration from the discrete spacer-disk
-tendon model of
-[Feliu-Talegon et al. (2025)](https://doi.org/10.1109/TMECH.2025.3581774), in
-which each disk obeys
+sliding over contact. SoRoMoX takes inspiration from a discrete spacer-disk
+tendon model:
+
+!!! info "Capstan-model reference"
+    > Feliu-Talegon, D., Alkayas, A. Y., Adamu, Y. A., Mathew, A. T., &
+    > Renda, F. (2025). *Actuation Reading Insights: Estimating Shape and
+    > Forces in Tendon-Driven Slender Soft Robots*. IEEE/ASME Transactions on
+    > Mechatronics, 30(6), 7878–7888.
+    > [https://doi.org/10.1109/TMECH.2025.3581774](https://doi.org/10.1109/TMECH.2025.3581774)
+
+In that model, each disk obeys
 
 \[
 T_i=T_{i-1}e^{-\mu\varphi_i},
@@ -302,8 +344,19 @@ class HyperbolicFrictionParams(BaseThreadlikeFrictionParams):
     coefficient: Array
 
 
-def hyperbolic_effort_ratio(params, accumulated_turn):
-    """Return a custom effort ratio as a function of accumulated turn."""
+def hyperbolic_effort_ratio(
+    params: HyperbolicFrictionParams,
+    accumulated_turn: Array,
+) -> Array:
+    """Return a custom effort ratio as a function of accumulated turn.
+
+    Args:
+        params: Hyperbolic friction parameters.
+        accumulated_turn: Accumulated path turn with shape ``(num_paths,)``.
+
+    Returns:
+        Surviving effort ratios with shape ``(num_paths,)``.
+    """
     return 1.0 / (1.0 + params.coefficient * accumulated_turn)
 
 
@@ -358,8 +411,7 @@ static topology and require reconstruction.
 
 ## Rendering
 
-Renderer adapters turn path queries into semantic visual layers. Physics
-objects do not import renderer primitives or store visual style. Generic
+Renderer adapters turn path queries into semantic visual layers. Generic
 renderers draw swept geometry when a radius is configured and use polylines
 otherwise. I-SUPPORT retains its detailed bellows renderer.
 
--- a/docs/api/actuation/threadlike.md
+++ b/docs/api/actuation/threadlike.md
@@ -182,7 +182,7 @@ coordinates = robot.actuator_coordinates(q)
 For tendons, `coordinates == -lengths`; for a pressure preset, coordinates are
 equivalent volumes. Without guide friction,
 `jax.jacrev(robot.actuator_coordinates)(q) == robot.actuation_matrix(q).T` up
-to quadrature tolerance. A nonzero friction coefficient breaks that identity by
+to quadrature tolerance. A lossy friction law breaks that identity by
 design, because the moment matrix then carries a transmission loss that path
 length does not; see [Guide friction](#guide-friction).
 
@@ -192,9 +192,21 @@ behavior is common, while each host keeps its native evaluation path.
 
 ## Guide friction
 
-By default a path transmits its effort undiminished along its whole route. Set
-`friction_coefficient` to model Coulomb friction against the guides the path
-runs through, following the force model of
+By default a path transmits its effort undiminished along its whole route. Pass
+a `friction=` law to model a transmission loss. The law is an object, so a new
+loss model needs no change to any host, and a third-party model needs no change
+to soromox.
+
+soromox ships three:
+
+| Law | Ratio | Hosts |
+|---|---|---|
+| `Frictionless` | \(1\) | all; the default |
+| `CapstanFriction` | \(e^{-\mu\Theta}\) | `PCS`, `PlanarPCS` |
+| `ExponentialLengthFriction` | \(e^{-ks}\) | all, including `GVS` |
+
+`CapstanFriction` models Coulomb friction against the guides a path runs
+through, following the force model of
 [Feliu-Talegon et al. (2025)](https://doi.org/10.1109/TMECH.2025.3550846). A
 tendon threaded through spacer disks loses tension at every contact, and the
 loss at each contact obeys the classical Capstan (Euler) law
@@ -217,10 +229,11 @@ contributes less than the proximal part:
 A_\mu(q) = B_\xi^\top \int_0^L \eta(q, s)\,\Phi(q, s)\,\mathrm{d}s
 \]
 
-\(\eta\) depends on the configuration only, so the actuation force stays linear
-in the effort and every controller keeps working. Since \(\eta \in (0, 1]\), the
-model can only ever remove effort, never add it, and a coefficient of zero
-reproduces the frictionless matrix exactly.
+Every law obeys the same two rules. \(\eta\) depends on the configuration only,
+never on the effort, so the actuation force stays linear in the effort and every
+controller keeps working; and \(\eta \in (0, 1]\), so a law can only ever remove
+effort, never add it. The lossless default reproduces the frictionless matrix
+exactly and is skipped entirely, at no cost.
 
 The wrap density reduces to \(\lvert\kappa\rvert\) identically in the plane, for
 any offset and any strain. Under torsion it recovers the helix curvature of an
@@ -231,27 +244,39 @@ robot = PCS.from_links(
     links,
     structure=PCSStructure(num_gauss_points=5),
     # a scalar shares one coefficient; pass shape (num_paths,) for per-path values
-    actuators=ThreadlikeActuator.tendons(routing, friction_coefficient=0.2),
+    actuators=ThreadlikeActuator.tendons(
+        routing, friction=CapstanFriction(coefficient=0.2)
+    ),
 )
 
 tau = robot.actuation_force(q, jnp.array([5.0, 0.0]))
 ```
 
-`PCS` and `PlanarPCS` implement the wrap-angle model. `GVS` strain varies along
-arc length, so its wrap angle is not piecewise linear; it rejects a nonzero
-coefficient during construction with a descriptive `TypeError`.
+`friction_coefficient=0.2` is kept as sugar for the same thing. Passing both it
+and `friction=` raises, because the two would silently disagree.
 
-### Identifying the coefficient
+A law declares what it needs, and a host refuses a law it cannot serve.
+`CapstanFriction` sets `requires_wrap_angle`, which `PCS` and `PlanarPCS`
+supply. `GVS` strain varies along arc length, so its wrap angle is not
+piecewise linear and it supplies none; it rejects a wrap-angle law during
+construction with a descriptive `TypeError`, at any coefficient including zero,
+because capability follows the law you install rather than the value you gave
+it. `ExponentialLengthFriction` reads only arc length, so it runs on `GVS` too.
 
-The coefficient is an ordinary dynamic leaf, so it is traceable and
+### Identifying a loss parameter
+
+Loss parameters are ordinary dynamic leaves, so they are traceable and
 differentiable under `jit`:
 
 ```python
 def with_friction(mu):
-    transmission = robot.actuators[0].params.transmission.replace(
-        friction_coefficient=mu
+    transmission = robot.actuators[0].params.transmission
+    return robot.update_actuator_params(
+        0,
+        transmission=transmission.replace(
+            friction=transmission.friction.replace(coefficient=mu)
+        ),
     )
-    return robot.update_actuator_params(0, transmission=transmission)
 
 
 @jax.jit
@@ -268,15 +293,64 @@ def loss(mu):
 value, grad = jax.value_and_grad(loss)(jnp.array(0.2))
 ```
 
-Zero is a valid optimizer start, but \(\mathrm{d}A/\mathrm{d}\mu\) vanishes
-wherever \(\Theta = 0\), so a perfectly straight pose carries no information
-about \(\mu\). Set `wrap_angle_smoothing` on the host structure to replace
+Swapping the law changes the parameter PyTree structure and so retraces, while
+replacing a value inside a law does not. Start identification from a lossy law
+at a zero parameter, `CapstanFriction(coefficient=0.0)`, rather than from
+`Frictionless()`. Zero is a valid optimizer start, but
+\(\mathrm{d}A/\mathrm{d}\mu\) vanishes wherever \(\Theta = 0\), so a perfectly
+straight pose carries no information about \(\mu\). Set `wrap_angle_smoothing` on the host structure to replace
 \(\lvert\kappa\rvert\) with \(\sqrt{\kappa^2 + \epsilon^2}\) if the fit can
 approach a straight configuration. It never under-estimates the wrap angle and
 over-estimates it by at most \(\epsilon L\), and it also removes the \(C^0\)
 kink at zero curvature.
 
-### What changes at a nonzero coefficient
+### Custom friction laws
+
+The hosts depend only on the law contract, not on the shipped models. A custom
+law supplies the parameters it needs, declares its requirements, and returns the
+surviving effort fraction per path:
+
+```python
+from soromox.actuation import ThreadlikeFriction
+
+
+class HyperbolicFriction(ThreadlikeFriction):
+    a: Array
+
+    def transmission_ratio(self, context):
+        return 1.0 / (1.0 + self.a * context.wrap_angle)
+
+
+actuator = ThreadlikeActuator.tendons(
+    routing, friction=HyperbolicFriction(a=jnp.asarray(0.5))
+)
+```
+
+The `context` is a `ThreadlikeQuadratureContext` describing the routed path at
+one quadrature node, batched over paths. Some of its fields mean the same thing
+on every host and some do not:
+
+| Field | Meaning | Portable |
+|---|---|---|
+| `arc_length` | backbone coordinate \(s\) | yes |
+| `segment_index` | segment containing the node | yes |
+| `offset`, `offset_derivative` | material-frame offset `(n, 3)` and its \(s\)-derivative | yes |
+| `tangent_norm` | routed-path stretch relative to the backbone | yes |
+| `active` | whether the node lies in each path's span | yes |
+| `unit_tangent` | normalized tangent, `(n, 2)` planar and `(n, 3)` spatial | no |
+| `strain` | `(n, 3)` planar and `(n, 6)` spatial | no |
+| `wrap_angle` | accumulated turn \(\Theta\), or `None` | only where supplied |
+
+A law reading only the portable fields runs unchanged on every host. A law
+reading `strain` or `unit_tangent` must document which host it targets.
+
+Set `requires_wrap_angle = False` if the law does not read `wrap_angle`, and the
+hosts will neither compute it nor refuse the law. Set `is_lossless = True` for a
+law that always returns one, and the hosts will skip it entirely. A law must
+keep its ratio in \((0, 1]\) and must not read the effort, which would make the
+actuation force nonlinear in the control.
+
+### What changes under a lossy law
 
 - `moment_matrix` is no longer the transpose of the length gradient, so
   `actuation_space` conversions that assume the identity do not apply.
@@ -319,7 +393,7 @@ robot = PCS(
 
 The passive element contributes to elastic force, damping, and elastic energy;
 it never consumes an active control channel. A passive path may also declare a
-`friction_coefficient`, in which case part of its spring force becomes
+`friction=` law, in which case part of its spring force becomes
 non-conservative and is applied through `actuation_force`; see
 [Guide friction](#guide-friction).
 
--- a/docs/api/actuation/threadlike.md
+++ b/docs/api/actuation/threadlike.md
@@ -19,7 +19,7 @@ The transmission model assumes that:
 - the actuator coordinate is determined by path length, scaled by a constant
   effective area for the pressure-chamber preset; and
 - each path carries one scalar work-conjugate effort, attenuated along its
-  route only by the optional Capstan friction model described below.
+  route by an optional friction model described below.
 
 Consequently, configuration- or time-dependent rerouting, changes in chamber
 cross-section, path slack, and internal actuator dynamics are outside the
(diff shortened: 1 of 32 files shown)


docs/api/actuation/threadlike.md
```diff
@@ -1,49 +1,33 @@
...
-Articulated hosts do not, so installing a threadlike component on them raises a
-descriptive `TypeError` during construction.
+Threadlike actuation models paths embedded in a continuum robot's material
+frame. The same geometry supports pulling tendons, pushing rods, simplified
+contractile muscles, and pressure chambers represented by an equivalent axial
+volume coordinate. A path may span any contiguous inclusive range of body
+segments.
+
+Following the general routed-path formulation of
+[Renda et al. (2022)](https://doi.org/10.1109/LRA.2022.3183248), a routing is a
```
mstoelzle: Can we highlight this reference in a nice citation box?

docs/api/actuation/threadlike.md
```diff
@@ -1,49 +1,33 @@
...
+
+The model assumes fixed material-frame routing, one scalar coordinate and
+effort per path, and an optional static effort loss along active paths. It does
+not model configuration-dependent rerouting, slack, actuator dynamics, or
+changes in pressure-chamber cross-section.
 
 ## Routing
 
-`ThreadlikeRouting` stores a vectorized family of paths. Linear paths use
+`ThreadlikeRouting` stores a vectorized path family. Linear paths use
```
mstoelzle, line 31: There should probably be a subsection dedicated to LinearRouting as we might generalize to other routing geometries later.

docs/api/actuation/threadlike.md
```diff
@@ -81,23 +79,30 @@ class SinusoidalRoutingParams(BaseThreadlikeRoutingParams):
...
-            raise ValueError("frequency must match amplitude")
-
 
 def offset(params, s):
+    """Return material-frame sinusoidal routing offsets at ``s``."""
     y = params.amplitude * jnp.sin(params.frequency * s)
     return jnp.stack((jnp.zeros_like(y), y, jnp.zeros_like(y)), axis=-1)
 
 
 def derivative(params, s):
```
mstoelzle: Would it make sense to generalize the `derivative` API for n-order derivatives instead of having dedicated methods for the 1st-derivate, 2nd-derivative, etc.?

docs/api/actuation/threadlike.md
```diff
@@ -81,23 +79,30 @@ class SinusoidalRoutingParams(BaseThreadlikeRoutingParams):
...
+    """Return the first arc-length derivative of ``offset``."""
     y_s = params.amplitude * params.frequency * jnp.cos(params.frequency * s)
     return jnp.stack((jnp.zeros_like(y_s), y_s, jnp.zeros_like(y_s)), axis=-1)
 
 
+def second_derivative(params, s):
+    """Return the second arc-length derivative of ``offset``."""
```
mstoelzle: Please stick to the docstring convention in terms of format and style established in this package.

docs/api/actuation/threadlike.md
```diff
@@ -81,23 +79,30 @@ class SinusoidalRoutingParams(BaseThreadlikeRoutingParams):
...
-
 
 def offset(params, s):
+    """Return material-frame sinusoidal routing offsets at ``s``."""
     y = params.amplitude * jnp.sin(params.frequency * s)
     return jnp.stack((jnp.zeros_like(y), y, jnp.zeros_like(y)), axis=-1)
 
 
 def derivative(params, s):
+    """Return the first arc-length derivative of ``offset``."""
```
mstoelzle: Please stick to the docstring convention in terms of format and style established in this package.

docs/api/actuation/threadlike.md
```diff
@@ -126,56 +131,200 @@ positive input always means the named physical action.
...
 coordinates = robot.actuator_coordinates(q)
+coordinate_jacobian = robot.actuator_coordinate_jacobian(q)
+actuation_matrix = robot.actuation_matrix(q)
 ```
 
-For tendons, `coordinates == -lengths`; for a pressure preset, coordinates are
-equivalent volumes. In every case,
-`jax.jacrev(robot.actuator_coordinates)(q) == robot.actuation_matrix(q).T` up
-to quadrature tolerance.
+The singular `coordinate` in `coordinate_jacobian` follows the usual compound
```
mstoelzle: Remove this paragraph where you discuss singular vs. plural.

docs/api/actuation/threadlike.md
```diff
@@ -126,56 +131,200 @@ positive input always means the named physical action.
...
+SoRoMoX evaluates this integral using the host quadrature. Genuine one-sided
+tangent jumps introduced by a piecewise host discretization contribute their
+unsigned angle at the interface.
+
+### Relation to the cited discrete model
+
+The exponential Capstan relation is established for ropes, belts, and tendons
+sliding over contact. SoRoMoX takes inspiration from the discrete spacer-disk
+tendon model of
+[Feliu-Talegon et al. (2025)](https://doi.org/10.1109/TMECH.2025.3581774), in
```
mstoelzle: Please highlight this reference in a nice citation box.

docs/api/actuation/threadlike.md
```diff
@@ -126,56 +131,200 @@ positive input always means the named physical action.
...
+## Continuum Capstan friction
+
+`ThreadlikeFriction.capstan(coefficient=mu)` attenuates active threadlike effort
+along increasing backbone coordinate, starting at each path's configured
+anchor. It is available for `PCS`, `PlanarPCS`, and `GVS`. The default
+`ThreadlikeFriction.frictionless()` returns unit effort ratio.
+
+### Path-turn derivation
+
+Let \(s\) be undeformed backbone arc length, \(R(s)\in SO(3)\) the material
```
mstoelzle: By convention, we refer to `s` as the backbone absicssa instead of arc length.

docs/api/actuation/threadlike.md
```diff
@@ -186,67 +335,37 @@ passive = ThreadlikeImpedance(
...
-        kind_radii={"tendon": 5e-4},
-    )
-)
-renderer.show(q, actuator_inputs=u, color_config=visuals)
-```
-
-I-SUPPORT retains its detailed bellows renderer. Its renderer accepts the same
-`actuator_inputs=` argument and also supports `pressures=` as a mutually
-exclusive convenience alias.
+Renderer adapters turn path queries into semantic visual layers. Physics
```
mstoelzle: >  Physics objects do not import renderer primitives or store visual style.
Remove this comment

src/soromox/actuation/__init__.py
```diff
@@ -10,6 +10,18 @@
...
 )
+from .friction import (
+    BaseThreadlikeFrictionParams,
+    CapstanFrictionParams,
+    FrictionlessParams,
+    ThreadlikeFriction,
+    capstan_effort_ratio,
+    frictionless_effort_ratio,
+    threadlike_tangent,
+    threadlike_tangent_derivative,
```
mstoelzle: Can the tangent derivative and the unit tangent derivative be integrated/collapsed into a single method? Or do they represent vastly different quantities that need to be computed differently?

src/soromox/actuation/core.py
```diff
@@ -306,7 +315,13 @@ def coordinates(self, robot: SoftRobot, q: Array) -> Array:
...
         return q
 
-    def moment_matrix(self, robot: SoftRobot, q: Array) -> Array:
+    def coordinate_jacobian(self, robot: SoftRobot, q: Array) -> Array:
+        """Return the identity coordinate Jacobian."""
+        del robot
+        return jnp.eye(self.num_channels, dtype=q.dtype)
+
+    def actuation_matrix(self, robot: SoftRobot, q: Array) -> Array:
+        """Return the identity effort-to-generalized-force map."""
```
mstoelzle: Please stick to the docstring style and format established in the package.

src/soromox/actuation/core.py
```diff
@@ -306,7 +315,13 @@ def coordinates(self, robot: SoftRobot, q: Array) -> Array:
         del robot
         return q
 
-    def moment_matrix(self, robot: SoftRobot, q: Array) -> Array:
+    def coordinate_jacobian(self, robot: SoftRobot, q: Array) -> Array:
+        """Return the identity coordinate Jacobian."""
```
mstoelzle: Please stick to the docstring style and format established in the package.

src/soromox/actuation/core.py
```diff
@@ -221,24 +230,24 @@ def efforts(
             qd: Optional generalized velocities of shape
                 ``(robot.num_velocities,)``. If omitted for a velocity-dependent
                 effort model, zero generalized velocity is used.
-            transmission_matrix: Optional precomputed transmission matrix with
-                shape ``(robot.num_velocities, self.num_channels)``.
+            coordinate_jacobian: Optional precomputed coordinate Jacobian with
```
mstoelzle, line 291: How should the `coordinate_jacobian` be defined? As the transpose or as it is currently? Also, the docstring should specify which velocities the jacobian maps to which other velocities.

src/soromox/actuation/friction.py
```diff
@@ -0,0 +1,509 @@
...
+    jumps caused by the host discretization are added at crossed segment
+    boundaries. These are continuum-model interfaces, not discrete friction
+    guides.
+
+    Args:
+        turn_rate_fn: Function accepting ``(segment_index, physical_points)``
+            and returning turn rates shaped ``(num_points, num_paths)``.
+        boundary_turn_fn: Function accepting a left segment index and returning
+            one-sided tangent angles shaped ``(num_paths,)``.
+        s: Physical backbone coordinate through which to integrate.
```
mstoelzle: This should be called "backbone abscissa"

src/soromox/actuation/threadlike.py
```diff
@@ -123,17 +145,21 @@ class BaseThreadlikeRoutingParams(BaseSystemParams):
 
     @property
     def num_paths(self) -> int:
+        """Return the number of routed paths represented by the parameters."""
         raise NotImplementedError
 
     @property
     def start_segment_index_array(self) -> Array:
+        """Return the first active segment of every path as an integer array."""
```
mstoelzle: Please stick to the docstring style and format established in the package.

src/soromox/actuation/threadlike.py
```diff
@@ -123,17 +145,21 @@ class BaseThreadlikeRoutingParams(BaseSystemParams):
...
         return jnp.asarray(self.start_segment_index, dtype=jnp.int32)
 
     @property
     def end_segment_index_array(self) -> Array:
+        """Return the last active segment of every path as an integer array."""
         return jnp.asarray(self.end_segment_index, dtype=jnp.int32)
 
     def validate_structure_for_robot(self, num_segments: int) -> None:
+        """Validate path spans against a host robot's segment count."""
```
mstoelzle: Please stick to the docstring style and format established in the package.

src/soromox/actuation/threadlike.py
```diff
@@ -143,6 +169,7 @@ def validate_structure_for_robot(self, num_segments: int) -> None:
                 )
 
     def assert_same_topology(self, other: BaseThreadlikeRoutingParams) -> None:
+        """Reject a replacement that changes routing type, count, or spans."""
```
mstoelzle: Please stick to the docstring style and format established in the package.

src/soromox/actuation/threadlike.py
```diff
@@ -382,39 +493,97 @@ def __init__(
...
+            One work-conjugate coordinate per path.
+        """
         return self.params.coordinate_scale * robot._threadlike_path_lengths(
             q, self.routing
         )
 
-    def moment_matrix(self, robot, q: Array) -> Array:
-        raw_matrix = robot._threadlike_moment_matrix(q, self.routing)
+    def coordinate_jacobian(self, robot: SoftRobot, q: Array) -> Array:
+        """Return the Jacobian of scaled path coordinates.
```
mstoelzle: Again, please document which velocities the Jacobian maps to which velocities.

src/soromox/actuation/threadlike.py
```diff
@@ -382,39 +493,97 @@ def __init__(
...
+            Matrix with shape ``(robot.num_velocities, num_channels)``.
+        """
+        raw_matrix = robot._threadlike_actuation_matrix(
+            q, self.routing, friction=self.friction
+        )
         return raw_matrix * self.params.coordinate_scale[None, :]
 
-    def path_lengths(self, robot, q: Array) -> Array:
+    def path_lengths(self, robot: SoftRobot, q: Array) -> Array:
+        """Return unscaled physical routed-path lengths."""
```
mstoelzle: Please stick to the docstring style and format established in the package.

src/soromox/actuation/threadlike.py
```diff
@@ -382,39 +493,97 @@ def __init__(
...
         return robot._threadlike_path_lengths(q, self.routing)
 
-    def path_velocities(self, robot, q: Array, qd: Array) -> Array:
-        return robot._threadlike_moment_matrix(q, self.routing).T @ qd
+    def path_length_jacobian(self, robot: SoftRobot, q: Array) -> Array:
+        """Return the Jacobian of unscaled physical path lengths."""
```
mstoelzle: Please stick to the docstring style and format established in the package.

src/soromox/actuation/threadlike.py
```diff
@@ -382,39 +493,97 @@ def __init__(
...
+        return robot._threadlike_path_length_jacobian(q, self.routing)
 
-    def path_poses(self, robot, q: Array, s: Array) -> Array:
+    def path_velocities(self, robot: SoftRobot, q: Array, qd: Array) -> Array:
+        """Return unscaled physical path-length rates."""
```
mstoelzle: Please stick to the docstring style and format established in the package.

src/soromox/actuation/threadlike.py
```diff
@@ -561,7 +748,9 @@ def push_rods(
         routings: ThreadlikeRouting | BaseThreadlikeRoutingParams,
         *,
         labels: tuple[str, ...] | None = None,
+        friction: ThreadlikeFriction | None = None,
     ) -> ThreadlikeActuator:
+        """Construct compression push rods with lengthening-positive coordinates."""
```
mstoelzle: Please stick to the docstring style and format established in the package.

src/soromox/actuation/threadlike.py
```diff
@@ -598,7 +791,9 @@ def pressure_chambers(
         *,
         effective_areas: Array,
         labels: tuple[str, ...] | None = None,
+        friction: ThreadlikeFriction | None = None,
     ) -> ThreadlikeActuator:
+        """Construct pressure chambers scaled by their effective areas."""
```
mstoelzle: Please stick to the docstring style and format established in the package.

src/soromox/actuation/threadlike.py
```diff
@@ -735,51 +958,60 @@ def _from_params(
 
     @property
     def params(self) -> ThreadlikeImpedanceParams:
+        """Return the passive element parameters."""
         return self._params
 
     def with_params(self, params: BaseSystemParams) -> ThreadlikeImpedance:
+        """Return a copy carrying replacement impedance parameters."""
```
mstoelzle: Please stick to the docstring style and format established in the package.

src/soromox/coordinate_transformations/actuation_space.py
```diff
@@ -20,23 +20,31 @@ class ActuationSpaceDynamics(eqx.Module):
...
     defined by work-conjugate actuator coordinates and unactuated coordinates
-    that span the null space of the actuation matrix.
+    that complete the actuator-coordinate Jacobian.
 
-    The actuation space dynamics follow the formulation:
-        M_y @ ydd + eta_y @ yd + h_y = [tau, 0_{n-m}]
+    The actuation space dynamics follow the general formulation:
+        M_y @ ydd + eta_y @ yd + h_y = A_y @ e
+
+    with ``A_y = J(q)^(-T) @ A_q(q)``. For an ideal transmission this reduces
```
mstoelzle: I think it is easy to confuse the $J$ symbol here with the forward kinematics Jacobian. even though it refers to a different quantity. You could write $J_y$ or something similar

src/soromox/systems/gvs/core.py
```diff
@@ -5861,35 +5882,227 @@ def _threadlike_local_basis(
...
         strain = basis @ q_link + self.xi_ref_Xs[segment_index, point_index]
-        offset = jnp.append(routing.offset(path_params, s), 1.0)
-        derivative = jnp.append(routing.derivative(path_params, s), 1.0)
-        tangent_unnormalized = (derivative + se3.hat(strain) @ offset)[:-1]
-        tangent = safe_normalize(tangent_unnormalized, eps=self.global_eps)
-        local = jnp.hstack([so3.skew(offset[:-1]) @ tangent, tangent])
+        offset = routing.offset(path_params, s)
+        offset_derivative = routing.derivative(path_params, s)
+        tangent = threadlike_tangent(strain[:3], strain[3:], offset, offset_derivative)
+        return offset, safe_normalize(tangent, eps=self.global_eps), active
```
mstoelzle, line 5891: Please check the test coverage - not just this method that you added but also for the other methods and functions that you added - if both outputs and autograd gradients (forward and reverse) are well-defined at the singularities.

src/soromox/systems/gvs/core.py
```diff
@@ -5918,6 +6142,20 @@ def point_matrix(point_index: Array) -> Array:
         )
         return self.active_dof_map.T @ full
 
+    def _threadlike_path_length_jacobian(
+        self, q: Array, routing: ThreadlikeRouting
+    ) -> Array:
+        """Return the Jacobian of unscaled GVS routed-path lengths.
```
mstoelzle, line 6158: Please properly define the Jacobian here.

src/soromox/systems/gvs/core.py
```diff
@@ -5851,8 +5858,22 @@ def _threadlike_local_basis(
...
+    ) -> tuple[Array, Array, Array]:
+        """Return one GVS routed-path offset, unit tangent, and activity.
+
+        Args:
+            routing: Routed-path family.
+            path_params: Parameters of one path.
+            start_segment_index: First segment traversed by the path.
+            end_segment_index: Last segment traversed by the path.
+            q_link: Padded generalized link coordinates.
+            s: Physical backbone coordinate.
```
mstoelzle: `s` should be referred to as "backbone absicssa" everywhere.

src/soromox/systems/gvs/core.py
```diff
@@ -5861,35 +5882,227 @@ def _threadlike_local_basis(
...
+            length_gradient_density: Shape ``(6,)``.
+        """
+        local = jnp.hstack([so3.skew(offset) @ unit_tangent, unit_tangent])
         return active * local
 
-    def _threadlike_moment_matrix(self, q: Array, routing: ThreadlikeRouting) -> Array:
-        """Integrate raw routed-length moment arms in the native GVS basis."""
+    def _threadlike_strain_and_derivative(
+        self,
+        q_link: Array,
```
mstoelzle, line 5912: Is `q_link` an expression that we commonly use in the package?

src/soromox/systems/gvs/core.py
```diff
@@ -5861,35 +5882,227 @@ def _threadlike_local_basis(
...
+            basis = basis.at[:, :3, :].divide(length)
+            basis_derivative = basis_derivative.at[:, :3, :].divide(length)
+        reference = self.xi_ref_Xs[segment_index, 0]
+        strain = jnp.einsum("pij,j->pi", basis, q_link) + reference
+        strain_derivative = jnp.einsum("pij,j->pi", basis_derivative, q_link) / length
+        return strain, strain_derivative
+
+    def _threadlike_turn_rates(
+        self,
+        q_gathered: Array,
```
mstoelzle, line 5943: Is `q_gathered` an expression that we commonly use in the package?

src/soromox/systems/gvs/core.py
```diff
@@ -5861,35 +5882,227 @@ def _threadlike_local_basis(
...
+        segment_index: Array,
+        points: Array,
+    ) -> Array:
+        """Evaluate analytical GVS path-turn rates at physical points.
+
+        Args:
+            q_gathered: Per-segment padded generalized coordinates.
+            routing: Routed-path family.
+            segment_index: Segment containing every supplied point.
+            points: Physical backbone coordinates with shape ``(num_points,)``.
```
mstoelzle: Wouldn't `s_ps` better fit the convention established in the package?

src/soromox/systems/pcs/pcs.py
```diff
@@ -3103,27 +3110,201 @@ def _threadlike_local_basis(
...
+            length_gradient_density: Shape ``(6,)``.
+        """
+        basis = jnp.hstack([so3.skew(offset) @ unit_tangent, unit_tangent])
         return active * basis
 
-    def _threadlike_moment_matrix(self, q: Array, routing: ThreadlikeRouting) -> Array:
-        """Integrate raw routed-length moment arms in the PCS strain basis."""
+    def _threadlike_turn_rates(
+        self,
+        strains: Array,
```
mstoelzle: Do we usually refer to the strains as `strains` or as `xi`?

src/soromox/systems/pcs/pcs.py
```diff
@@ -3103,27 +3110,201 @@ def _threadlike_local_basis(
...
+        routing: ThreadlikeRouting,
+        segment_index: Array,
+        points: Array,
+    ) -> Array:
+        """Evaluate analytical spatial-path turn rates at physical points.
+
+        Args:
+            strains: Piecewise-constant PCS strains.
```
mstoelzle: Please document the array shapes for all args and return values.

src/soromox/systems/pcs/pcs.py
```diff
@@ -3103,27 +3110,201 @@ def _threadlike_local_basis(
...
+            One unsigned tangent angle per routed path.
+        """
+        s = self.L_cum[left_segment_index + 1]
+
+        def path_angle(path_params: BaseThreadlikeRoutingParams) -> Array:
+            """Return one routed path's tangent jump at the boundary."""
+            offset = routing.offset(path_params, s)
+            offset_derivative = routing.derivative(path_params, s)
+
+            def unit_tangent(strain: Array) -> Array:
```
mstoelzle, line 3222: Did you already specialize the `PlanarPCS` and `PCS` implementations to the constant-strain case where some of the expressions should simplify (as you mentioned yourself in the PR description)?

tests/coordinate_transformations/test_actuation_space_dynamics.py
```diff
@@ -307,18 +307,18 @@ def test_actuated_unactuated_coordinates(self, underactuated_pendulum):
...
 
         q = jnp.array([0.1, 0.2, 0.3])
         J_a = asd.jacobian_actuated(q)
-        A_at = robot.actuation_matrix(q)
+        expected = robot.actuator_coordinate_jacobian(q)
 
-        assert_allclose(J_a, A_at.T, rtol=1e-10)
+        assert_allclose(J_a, expected, rtol=1e-10)
 
     def test_jacobian_unactuated_equals_h_unactuated(self, underactuated_pendulum):
```
mstoelzle: I don't think it will be clear what is meant with "h_unactuated". Make the test name more self-explanatory.

docs/development/changelog.md
```diff
@@ -15,11 +15,26 @@ and include benchmark baseline and measurement context for performance claims.
...
+  analytical material-tangent derivative; custom routings provide an
+  analytical second derivative. Parameters remain differentiable leaves under
+  JAX transformations. See
+  [threadlike actuation](../api/actuation/threadlike.md#continuum-capstan-friction).
 
 ### Changed
 
 - Open3D and Viser swept backbones now preserve circular, elliptical, and
   rectangular cross-sections, including dimensions that vary with abscissa.
+- Corrected the planar routed-path offset sign, which used
```
mstoelzle: We had already a breaking-change entry in PR #202 . Therefore, I don't think we should add another entry.

src/soromox/actuation/threadlike.py
```diff
@@ -303,12 +352,23 @@ def linear(
...
     def num_paths(self) -> int:
+        """Return the number of paths represented by the routing family."""
         return self.params.num_paths
 
+    @property
+    def has_analytical_second_derivative(self) -> bool:
+        """Return whether the routing supplies the derivative needed for turn."""
+        return self.second_derivative_fn is not None or isinstance(
+            self.params, LinearThreadlikeRoutingParams
+        )
```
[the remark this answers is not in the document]
mstoelzle: Fixed. The implicit zero second derivative is now available only when both callbacks are the built-in linear routing functions (or an explicit second-derivative callback is supplied). Custom callbacks using `LinearThreadlikeRoutingParams` are rejected for friction unless they provide `second_derivative_fn`; a regression test covers this case.

tests/actuation/test_threadlike_friction.py
```diff
@@ -0,0 +1,492 @@
...
+def _sinusoidal_derivative(params: SinusoidalRoutingParams, s: Array) -> Array:
+    """Return the analytical first derivative of the sinusoidal offset."""
+    y = params.amplitude * params.frequency * jnp.cos(params.frequency * s)
+    return jnp.asarray([0.0, y, 0.0])
+
+
+def _sinusoidal_second_derivative(params: SinusoidalRoutingParams, s: Array) -> Array:
+    """Return the analytical second derivative of the sinusoidal offset."""
+    y = -(params.frequency**2) * params.amplitude * jnp.sin(params.frequency * s)
+    return jnp.asarray([0.0, y, 0.0])
```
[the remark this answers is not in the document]
mstoelzle: Fixed. The callbacks now use `jnp.stack(..., axis=-1)`, preserving the routing contract `(num_paths, 3)`. A two-path regression test checks all three callbacks.

docs/api/actuation/threadlike.md
mstoelzle: Updated. The Renda et al. reference is now presented in a dedicated citation callout with the full bibliographic entry and DOI.

docs/api/actuation/threadlike.md
mstoelzle: Updated. The routing section now starts with a dedicated “Linear routing” subsection, leaving the surrounding structure ready for additional routing geometries.

docs/api/actuation/threadlike.md
mstoelzle: I kept explicit first- and second-derivative APIs. The first derivative is part of the general routing/length contract, while the second derivative is the additional order specifically required by the friction turn-rate model. An arbitrary nth-order API would advertise unsupported higher orders, make static callback validation less explicit, and reduce discoverability without a current consumer. We can generalize if a concrete higher-order use case appears.

docs/api/actuation/threadlike.md
mstoelzle: Updated the example callback to the package’s typed Google-style docstring format, including `Args` and `Returns` with array shapes.

docs/api/actuation/threadlike.md
mstoelzle: Updated the example parameter property to the package’s typed Google-style docstring format, including its return type and `Returns` section.

docs/api/actuation/threadlike.md
mstoelzle: Removed the singular-versus-plural naming paragraph.

docs/api/actuation/threadlike.md
mstoelzle: Updated. The Feliu-Talegon et al. reference is now presented in a dedicated citation callout with the corrected DOI and full bibliographic entry.

docs/api/actuation/threadlike.md
mstoelzle: Updated the new documentation and implementation docstrings to call `s` the backbone abscissa.

docs/api/actuation/threadlike.md
mstoelzle: Removed this sentence.

src/soromox/actuation/__init__.py
mstoelzle: Collapsed these implementation details. The public tangent-derivative and unit-tangent-derivative helpers were removed; `threadlike_turn_rate` now forms the analytical tangent derivative and passes it to one private normalization/turn-rate helper. The material tangent remains separate because it is also used by path-length and boundary computations.

src/soromox/actuation/core.py
mstoelzle: Updated the identity actuation-matrix docstring to the package’s Google style with typed arguments, return symbol, and array shape.

src/soromox/actuation/core.py
mstoelzle: Updated the identity coordinate-Jacobian docstring to the package’s Google style with typed arguments, mapped velocities, return symbol, and array shape.

src/soromox/actuation/core.py
mstoelzle: Defined it as `J_a = dy_a/dq`, with shape `(num_channels, num_velocities)`, so actuator-coordinate velocities satisfy `yd_a = J_a(q) @ qd`. The actuation matrix keeps the dual orientation `tau = A(q) @ effort`; only an ideal transmission requires `A = J_a.T`.

src/soromox/actuation/friction.py
mstoelzle: Updated this and the other new occurrences to “backbone abscissa”; integration arguments also now use the established `s_ps` naming.

src/soromox/actuation/threadlike.py
mstoelzle: Updated with the package’s Google-style `Returns` section and the array shape.

src/soromox/actuation/threadlike.py
mstoelzle: Updated with the package’s Google-style `Returns` section and the array shape.

src/soromox/actuation/threadlike.py
mstoelzle: Updated to a Google-style docstring with `Args` and `Raises`, including the topology invariants being checked.

src/soromox/actuation/threadlike.py
mstoelzle: Documented the mapping explicitly: `yd_a = J_a(q) @ qd`, from generalized velocity to actuator-coordinate velocity, together with the matrix orientation and shape.

src/soromox/actuation/threadlike.py
mstoelzle: Updated to the package’s Google-style docstring format with argument and return shapes.

src/soromox/actuation/threadlike.py
mstoelzle: Updated to the package’s Google-style docstring format with argument and return shapes.

src/soromox/actuation/threadlike.py
mstoelzle: Updated to the package’s Google-style docstring format with argument and return shapes.

src/soromox/actuation/threadlike.py
mstoelzle: Updated the push-rod constructor to the package’s Google-style docstring format with `Args` and `Returns`.

src/soromox/actuation/threadlike.py
mstoelzle: Updated the pressure-chamber constructor to the package’s Google-style docstring format with `Args` and `Returns`, including the effective-area shape.

src/soromox/actuation/threadlike.py
mstoelzle: Updated `with_params` to the package’s Google-style docstring format with `Args`, `Returns`, and the relevant exceptions.

src/soromox/coordinate_transformations/actuation_space.py
mstoelzle: Updated the actuation-space notation to `J_y` throughout the affected formulas and prose, reserving plain forward-kinematics Jacobian notation from this coordinate-transform Jacobian.

src/soromox/systems/gvs/core.py
mstoelzle: Expanded singularity coverage across PCS, PlanarPCS, and GVS, with and without Capstan friction. The tests now check finite primal outputs plus `jax.jacfwd` and `jax.jacrev` derivatives for actuator coordinates, the actuator-coordinate Jacobian, and the actuation matrix. Separate helper-level tests cover collapsed turn rates and undefined boundary directions.

src/soromox/systems/gvs/core.py
mstoelzle: Defined the returned matrix as the unscaled routed-path length Jacobian `J_l`, including its shape and the velocity mapping `path_velocity = J_l @ qd`.

src/soromox/systems/gvs/core.py
mstoelzle: Updated all newly introduced GVS occurrences to call `s` the backbone abscissa.

src/soromox/systems/gvs/core.py
mstoelzle: Yes. `q_link` is already used extensively in the existing GVS kinematics and Jacobian implementation for a segment’s padded link coordinates, so I retained it and documented its exact shape here.

src/soromox/systems/gvs/core.py
mstoelzle: Yes. `q_gathered` is the established GVS name used throughout the existing forward-kinematics, Jacobian, and dynamics code for the padded per-segment joint/link coordinate array. I retained it and documented its exact shape.

src/soromox/systems/gvs/core.py
mstoelzle: Agreed and updated. The argument and related locals now use `s_ps` for batches of backbone abscissae.

src/soromox/systems/pcs/pcs.py
mstoelzle: The package predominantly uses `xi` for Cosserat strain arrays in these host implementations, so the new PCS and PlanarPCS friction paths now use `xi`/`xi_i` consistently.

src/soromox/systems/pcs/pcs.py
mstoelzle: Added shapes for every argument and return value in the new PCS routing helpers, including scalar segment indices/activity values and path-batch dimensions.

src/soromox/systems/pcs/pcs.py
mstoelzle: Now specialized explicitly. PCS and PlanarPCS call `threadlike_constant_strain_turn_rate`, which uses `omega' = v' = 0` and computes `a' = omega × r' + r''`; GVS retains the full varying-strain expression. A regression test verifies that the specialization equals the general formula with zero strain derivatives.

tests/coordinate_transformations/test_actuation_space_dynamics.py
mstoelzle: Renamed the test to `test_unactuated_coordinate_jacobian_matches_configured_mapping` and updated its docstring so the intent is explicit without relying on the internal `H_unactuated` symbol.

docs/development/changelog.md
mstoelzle: Removed PR #203’s duplicate planar sign/path-length changelog entries. The breaking-change note merged through PR #202 remains on `main`; this PR now adds only its own transmission-contract breaking change.

PR #204, opened 2026-09-03
Restore reproducible multi-start Section Vd control-gain optimization

Michele-Martini:
## Summary

This PR restores reproducible multi-start control-gain optimization for Section Vd and updates the committed results for the current one-segment PCS model.

- run six independent collocated and synergistic initializations through one shared optimization loop and generator
- use a shared dimensionless objective with retuned solver, controller, initialization, and Yogi settings
- store schema-v3 provenance, per-start histories, and trajectories that can be validated and rescored
- regenerate the comparison figure plus optimized-best and initial-median animations for both methods
- replace duplicated scripts with shared case, CLI, initialization, objective, result, and tuning modules

## Implementation

The optimization loop vmaps independent optimizer states, freezes non-finite starts, and keeps iteration-aligned loss, gain, gradient-norm, and update-norm histories. The result schema records the objective scales, solver step, material damping, saturation, initialization, optimizer metadata, and selected start.

The plotter selects each archive's best start, uses the canonical loss without controller-specific scaling, and safely handles non-positive or masked values on logarithmic axes. Explicit CPU and GPU selections are strict; automatic selection remains delegated to JAX.

The renderer defaults to `optimized-best`. The `initial-median` option selects the finite initial rollout closest to the median initial loss. Outputs use explicit `_optimized_best_tracking` and `_initial_median_tracking` names and one shared gravity-down camera.

## Validation

- Section Vd non-slow suite: `142 passed, 2 deselected`
- renderer and plot regressions: `47 passed`
- `git diff --check`: passed
- all regenerated MP4/GIF outputs decoded and were visually inspected


mstoelzle:
@Michele-Martini Before I review the PR in more detail, a few high-level points first:

- The actuation coordinates optimization (upper row) doesn't look very stable/well-behaved. What is happening here? Is the learning rate too high? Or does the changed optimization landscape exhibit many local minima and the various seeds are jumping around the solutions?
- Can you please expand on how you tuned the integral saturation threshold? What made you keep the `e_sat = 10 mm`? What was the motivation / decision criterion?
- The legacy script and data shouldn't remain in the repo. Please store and back-it-up locally in case we need to refer back to it in the future.
- I like the new normalized error formulations - should make it easier to consistenly select optimization hyperparameters! :) 

> Retain the 1:5 early/late weighting only if that scientific preference is intentional and documented. Dropped.

What is meant with that? Does it mean that the early/late weighting is dropped and we now have a uniform objective?

Michele-Martini:
> * The actuation coordinates optimization (upper row) doesn't look very stable/well-behaved. What is happening here? Is the learning rate too high? Or does the changed optimization landscape exhibit many local minima and the various seeds are jumping around the solutions?

I have decreased the learning step and changed the seed of initialization of the collocated PID gains to avoid overly "soft" controllers. The results is this:

plot_control_gain_opt

> * Can you please expand on how you tuned the integral saturation threshold? What made you keep the `e_sat = 10 mm`? What was the motivation / decision criterion?

`e_sat` is initialized at 10 mm as sensible physical value. The script then sweeps over a sensible range of `e_sat`, make some optimization iterations, and checks (i) how much time the saturation was active along the rollout, (ii) the impact of change of `e_sat` on the loss. Since there saturation is hardly active and the loss shows no sensitivity to the parameter for the cases presented, the initial value is fine.

> * The legacy script and data shouldn't remain in the repo. Please store and back-it-up locally in case we need to refer back to it in the future.

Stored and removed.
 
> Retain the 1:5 early/late weighting only if that scientific preference is intentional and documented. Dropped. What is meant with that? Does it mean that the early/late weighting is dropped and we now have a uniform objective?

Exactly, as you can see from the new loss, everything is plain as that and there is no transient/steady-state split anymore.

docs/development/changelog.md
```diff
@@ -13,139 +13,281 @@ and include benchmark baseline and measurement context for performance claims.
...
-  threadlike actuation.
-- Added native matrix-free Warp products for `J(q, s) @ qd` and
-  `sum_s J(q, s).T @ wrench_s`. These paths reuse caller-owned workspace and
-  floating-frame state instead of materializing a Jacobian at every spatial
-  sample.
-- Added fixed and floating-base configurations across all system families,
-  including planar and quaternion base poses, distinct configuration and
-  velocity dimensions, world-frame base velocities and wrenches, vectorized
-  JAX composition, and specialized Warp execution; see
+- Recovered and archived the generator behind the published Section Vd
```
mstoelzle: This changelog entry should be removed as you removed the previous/legacy results.

docs/development/changelog.md
```diff
@@ -13,139 +13,281 @@ and include benchmark baseline and measurement context for performance claims.
...
+  six-start results, together with scripts that verify it: the collocated
+  objective is re-derived from the archived trajectories to 6.1e-16 relative,
+  and re-running the generator reproduces the archived initial losses to 1.8e-9.
+  This answers [issue #154](https://github.com/tud-phi/soromox/issues/154) --
+  the six initial gain sets, `PRNGKey(15)`, the sampling procedure, the
+  vectorized multi-start execution, and how selection and aggregation worked.
+  The findings are recorded on the issue itself.
+- Restored multi-start Section Vd gain optimization: both optimizers now run six
```
mstoelzle: Combine these three changelog entries into one concise changelog entry on secVd results.

docs/development/changelog.md
```diff
@@ -13,139 +13,281 @@ and include benchmark baseline and measurement context for performance claims.
...
-- Reworked the execution layer around shared quaternion, `SO(2)`, `SO(3)`,
-  `SE(2)`, and `SE(3)` primitives and explicit operand/output contracts, so the
-  same public model methods can dispatch to optimized JAX or Warp paths without
-  changing the model definition.
-- Reduced GVS dynamics assembly to compact strain bases and bounded active-DOF
-  prefixes, and reused intermediate velocity and floating-root rotation terms.
-- Made effort laws declare which transmission states they need. `DirectEffort`
-  now skips unused actuator-coordinate and velocity evaluation, and each
-  actuator moment matrix is assembled only once.
+- Simplified the Section Vd case now that its tuning questions are answered.
```
mstoelzle: Waayyy too many details. Should be summarized in one concise changelog entry about secVd.

docs/development/changelog.md
```diff
@@ -173,8 +315,29 @@ and include benchmark baseline and measurement context for performance claims.
   standard `SoftRobot` systems automatically, while robot-defined hooks remain
   available for specialized visualization behavior.
 
+### Removed
+
+- Removed the recovered Section Vd generator and every trace of its conventions
```
mstoelzle: This is supposed to be a changelog entry that describes the delta with respect to v0.4.0. Therefore, this entry needs to be removed.

docs/development/changelog.md
```diff
@@ -173,8 +315,29 @@ and include benchmark baseline and measurement context for performance claims.
...
+  one job, answering [issue #154](https://github.com/tud-phi/soromox/issues/154),
+  where its findings are recorded; carrying a second sampler that belonged to a
+  different robot invited it into new work. Archives keep
+  their `init_scheme` field, now fixed at `log_uniform_v1` from
+  `secvd_init.SECVD_INIT_SCHEME`, so the committed data is unchanged and a
+  future sampler is still distinguishable.
+
 ### Fixed
 
+- Fixed the Section Vd generator and tuning study defaulting `--solver-dt` to the
```
mstoelzle: Revise or fully remove

docs/research.md
```diff
@@ -109,6 +109,15 @@ gain initializations are evaluated in parallel. For actuation-space and
 operational-space control, the optimized losses decrease by 62% and 57%,
 respectively, relative to their initial median values.
 
+!!! note "Provenance"
```
mstoelzle: Update the figures so we can remove the note.

pyproject.toml
```diff
@@ -352,6 +353,10 @@ exclude = [
     "build",
     "dist",
     "*.egg-info",
+    # Byte-exact archival copy of the recovered Section Vd generator. It targets
```
mstoelzle: This entry can be removed.

.gitignore
```diff
@@ -170,3 +170,7 @@ site/
 # camera params saved during rendering
 DepthCamera_*.json
 DepthCapture_*.png
+
+# Section Vd tuning-study stage summaries: regenerated by secvd_tuning_study.py,
+# never a published result. See paper_results/secVd_control_gain_optimization/README.md
+paper_results/secVd_control_gain_optimization/data/tuning/
```
mstoelzle: This should go into a (new) `.gitignore` in `paper_results/secVd_control_gain_optimization/data`

.gitattributes
```diff
@@ -10,3 +10,7 @@ docs/assets/paper/parallel-rl-trained-64-envs.gif filter=lfs diff=lfs merge=lfs
 paper_results/secVd_control_gain_optimization/data/*/optimization_results.npz filter=lfs diff=lfs merge=lfs -text
 paper_results/secVd_control_gain_optimization/outputs/*.mp4 filter=lfs diff=lfs merge=lfs -text
 paper_results/secVd_control_gain_optimization/outputs/*.gif filter=lfs diff=lfs merge=lfs -text
+# Archival copy of the recovered Section Vd generator: preserve its original
```
mstoelzle: This entry can be removed.

paper_results/secVd_control_gain_optimization/README.md
```diff
@@ -1,109 +1,133 @@
...
-back to CPU with a warning, while an explicit `--device gpu` request fails if
-CUDA is unavailable.
+```bash
+uv sync --extra paper_results
+```
 
-The collocated controller integrates tendon-length errors in metres and uses
-the unit-preserving saturation
+One program runs both optimizations, selected by `--method`, so the two cannot
+drift apart:
```
mstoelzle: > , so the two cannot drift apart:

Remove comment

paper_results/secVd_control_gain_optimization/README.md
```diff
@@ -1,109 +1,133 @@
...
 ```
 
-The default is `e_sat = 10 mm` (`gamma = 100 1/m`) and can be changed with
-`--integral-error-saturation-scale` in metres.
-
-## Plotting
+Defaults reproduce what ships. A run is saved only if every requested iteration
+completed and every start produced a finite iterate, so an interrupted run
+cannot replace an archive; `--result-dir` stages results elsewhere. `--device`
+takes `auto`, `cpu` or `gpu`, and pinning one makes the run fail rather than
```
mstoelzle: >  rather than silently start on another backend.

Remove such overly verbose comments.

paper_results/secVd_control_gain_optimization/README.md
```diff
@@ -1,109 +1,133 @@
...
+Redraw the four-panel comparison figure from the committed archives:
 
 ```bash
-uv run python paper_results/secVd_control_gain_optimization/code/plot_control_gain_optimization.py \
-  --force
+uv run python paper_results/secVd_control_gain_optimization/code/plot_control_gain_optimization.py
 ```
 
-## Robot rendering
+The plotter overwrites its outputs by default (`--no-force` refuses), since
```
mstoelzle: > , since redrawing costs seconds while a generator overwrite destroys an optimization

Remove such overly verbose comments

paper_results/secVd_control_gain_optimization/README.md
```diff
@@ -1,109 +1,133 @@
...
-in normal Git. Canonical optimization archives and robot-rendering MP4/GIF
-files are also versioned, but routed through Git LFS because full optimization
-and rendering runs can make them substantially larger. Directories matching
-`outputs/*_diagnostics/` are local scratch artifacts from the removed embedded
-diagnostic workflow and are intentionally ignored.
+## Archive format
+
+Each optimizer writes only `optimization_results.npz`, schema version 3. The
+batch axis of width `B` appears **only** on quantities that vary per start, so a
+batch axis in this archive always means something:
```
mstoelzle: > so a batch axis in this archive always means something:

Remove such overly verbose comments

tests/paper_results/secVd/conftest.py
```diff
@@ -0,0 +1,30 @@
+"""Run the Section Vd tests on the CPU.
```
mstoelzle: Remove the comment. It is enough to state that all tests are run on the CPU.

paper_results/secVd_control_gain_optimization/code/plot_control_gain_optimization.py
```diff
@@ -86,11 +160,40 @@ def _plot_three_channels(
     ylabel: str,
     panel: str,
     legend: bool,
+    ylim: tuple[float, float] | None = None,
 ) -> None:
+    """Draw target, the initial spread across starts, and the best start.
+
+    Args:
+        ax: Axes to draw on.
```
mstoelzle, line 191: Strange order arguments. Use a more logical order.

paper_results/secVd_control_gain_optimization/code/secvd_cli.py
```diff
@@ -0,0 +1,196 @@
...
+from pathlib import Path
+
+from secvd_init import (
+    GAIN_ORDER,
+    SECVD_BATCH_SIZE,
+    SECVD_INIT_SPREAD,
+)
+from secvd_results import METHODS, RESULTS_FILENAME
+
+DEVICE_CHOICES = ("auto", "cpu", "gpu")
```
mstoelzle, line 18: If the user explicitly requests a device, the code should not silently fall back if the device is not available. Please double check if that is the case.

paper_results/secVd_control_gain_optimization/code/secvd_case.py
```diff
@@ -1,19 +1,53 @@
...
+COMMITTED_ALPHA = 0.5
+COMMITTED_LR_RATIO = {"Kp": 1.0, "Ki": 0.5, "Kd": 0.1}
+TUNED_ALPHA = {"collocated": 1500.0, "synergistic": 1547.0}
+TUNED_INIT_SEED = {"collocated": 2, "synergistic": 0}
+MATERIAL_DAMPING = 7.2e1
+INIT_GAIN_SCALE = {
+    "collocated": {"Kp": 1.0, "Ki": 1.0, "Kd": 0.2},
+    "synergistic": {"Kp": 3.0, "Ki": 1.0, "Kd": 0.2},
+}
+COMMITTED_E_SAT = 1e-2
```
mstoelzle, line 47: Please double check that the integral saturation is active at any point during the closed-looop rollout.

mstoelzle (changes requested):
Some more cleaning-up and double checking is required.

docs/development/changelog.md
mstoelzle: Removed the legacy recovery changelog entry.

docs/development/changelog.md
mstoelzle: Removed this standalone fix entry; the tuned solver setting is covered by the concise Section Vd change summary.

docs/development/changelog.md
mstoelzle: Removed this entry because it described intermediate branch history rather than the delta from v0.4.0.

docs/development/changelog.md
mstoelzle: Combined the three additions into one concise Section Vd results entry.

pyproject.toml
mstoelzle: Removed the obsolete Ruff exclusion.

docs/research.md
mstoelzle: Regenerated the comparison figure, updated the documentation asset and current 47%/99% metrics, and removed the provenance note.

paper_results/secVd_control_gain_optimization/code/plot_control_gain_optimization.py
```diff
@@ -51,28 +77,76 @@ def _latex_is_usable() -> bool:
...
+def shared_loss_limits(*extents) -> tuple[float, float] | None:
+    """Combine per-panel extents into one padded limit for both loss panels."""
+    present = [e for e in extents if e is not None]
+    if not present:
+        return None
+    low = min(lo for lo, _ in present) / LOSS_YLIM_PAD
+    high = max(hi for _, hi in present) * LOSS_YLIM_PAD
+    if not low < high:
+        return None
+    return low, high
```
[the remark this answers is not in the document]
mstoelzle: Updated the log-loss path to derive extents only from positive finite values, fall back from shared limits when none exist, and clamp finite non-positive plotted values. Manual non-positive log limits are now rejected, with regression coverage for each case.

docs/development/changelog.md
mstoelzle: Collapsed the detailed Section Vd changes into one concise `Changed` entry.

paper_results/secVd_control_gain_optimization/code/secvd_loop.py
```diff
@@ -0,0 +1,288 @@
...
+        history.update_norm.append(iteration_update_norm)
+        history.time_iter.append(time.time() - iter_start)
+
+        frozen = int(batch_size - int(jnp.sum(alive)))
+        if not bool(jnp.any(alive)) and iteration < num_iters - 1:
+            history.stopped_early = True
+            history.stop_reason = (
+                f"all {batch_size} starts became non-finite by iteration "
+                f"{iteration}; recorded {len(history)} iterations"
+            )
```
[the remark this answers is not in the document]
mstoelzle: Updated the diagnostic to use `iteration + 1` and added a regression assertion for the first-iteration stop message.

paper_results/secVd_control_gain_optimization/README.md
mstoelzle: Removed this aside and rewrote the README more concisely.

paper_results/secVd_control_gain_optimization/README.md
mstoelzle: Trimmed this to: `--device auto` defers to JAX; explicit CPU and GPU selections are strict.

.gitattributes
mstoelzle: Removed the obsolete legacy-file attribute.

tests/paper_results/secVd/conftest.py
mstoelzle: Replaced the module comment with the single sentence: `Run all Section Vd tests on the CPU.`

paper_results/secVd_control_gain_optimization/README.md
mstoelzle: Removed the rationale and kept only the plotting command and limit options.

.gitignore
mstoelzle: Moved the rule into `paper_results/secVd_control_gain_optimization/data/.gitignore` as `tuning/`.

paper_results/secVd_control_gain_optimization/README.md
mstoelzle: Removed this aside and reduced the archive section to the schema contract and field shapes.

paper_results/secVd_control_gain_optimization/code/plot_control_gain_optimization.py
mstoelzle: Reordered the helper arguments to `target`, `initial`, `best`, matching the plot and legend order.

paper_results/secVd_control_gain_optimization/code/secvd_case.py
mstoelzle: Verified against both committed collocated rollouts. The control error peaks at 14.51 mm, above the 10 mm saturation scale, for both initial and optimized trajectories. Added a regression test that requires the threshold to be crossed.

paper_results/secVd_control_gain_optimization/code/secvd_cli.py
mstoelzle: Verified. Explicit GPU selection sets `JAX_PLATFORMS=cuda`; with CUDA devices hidden, JAX initialization fails instead of falling back to CPU. The existing strict-pin and backend-mismatch tests also pass.

mstoelzle:
Addressed the review in fddd6d5.

- reduced the Section Vd changelog and README to concise, current-delta documentation
- regenerated the comparison outputs and updated the documentation figure and metrics
- made log-loss limits and plotted values safe for non-positive data
- moved the tuning ignore rule into the data directory and removed obsolete legacy-file configuration
- made the early-stop message 1-based and reordered trajectory inputs
- verified strict explicit-device selection and confirmed the 10 mm integral saturation is active in both committed collocated rollouts (14.51 mm peak), with regression coverage

Verification: 133 non-slow Section Vd tests passed, 6 focused review regressions passed, Ruff passed, the Zensical documentation build passed, and git diff --check passed.

mstoelzle:
Addressed in `c9a278a`:

- renamed best-render outputs to `_optimized_best_tracking`
- applied one fixed, gravity-down camera to every Section Vd animation and removed the frame-zero camera jump
- regenerated optimized-best and initial-median MP4/GIF outputs for both methods
- rebased onto current `origin/main`

Validation: 142 passed, 2 slow tests deselected; Ruff, formatting, docs, media decoding, and visual checks pass. All review threads are resolved.

mstoelzle (approved):
Seems to work well now.

PR #206, opened 2026-09-04
Restrict semi-implicit Euler to soft robots

mstoelzle:
## Summary

- type `SemiImplicitEuler.system` as `SoftRobot`, which defines the state helpers used by the integrator
- clarify that `DynamicalSystem.forward_dynamics` does not assume a particular state decomposition or system order
- describe `DynamicalSystem.project_state` in terms of the generic state manifold instead of a configuration manifold

## Validation

- `ruff check src/soromox/simulation/integrators.py src/soromox/systems/dynamical_system.py`
- `ruff format --check src/soromox/simulation/integrators.py src/soromox/systems/dynamical_system.py`
- `JAX_PLATFORMS=cpu JAX_ENABLE_X64=1 pytest -q tests/simulation` (7 passed)

### DIFF
--- a/docs/development/changelog.md
+++ b/docs/development/changelog.md
@@ -36,6 +36,8 @@ and include benchmark baseline and measurement context for performance claims.
 
 ### Fixed
 
+- Fixed `DynamicalSystem` API documentation to avoid assuming second-order
+  `[q, qd]` states or configuration-only projection.
 - Made Viser discrete cross-section markers consistent with Open3D: circular,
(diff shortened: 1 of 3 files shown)


Johannap1 (commented):
This fixes the issues of some methods in integrators.py only being available for SoftRobot. It does restrict its use to SoftRobot class, cannot be generalized to dynamical system. Something to keep in mind, but otherwise valid fix.

PR #210, opened 2026-09-09
Unify renderer configuration and add studio presets

mstoelzle:
## Summary

- introduce `RendererConfig` and organize public settings under `rendering/config/`, with scene, camera, colors, output and composed renderer modules
- add technical, neutral/bright/dark studio, flat and clay presets with photometric light units, world/base ground alignment and documented backend approximations
- add four shared mount shapes (disk, beveled disk, truncated cone and flared collar), with flared collars as the default
- smooth modern Open3D link joins by sharing side normals and removing buried caps at matching contours; preserve actual cross-section steps
- use modern Open3D rendering for static viewing and image/video export while retaining efficient legacy updates for animated previews with the same camera projection; simulation and control examples export trajectory videos by default
- add reproducible macOS Metal and Linux EGL/GLX source builds, restore tone-mapper selection for neutral studio colors and white technical backgrounds, and add Ubuntu native integration tests and an upstream-revision check
- migrate renderer constructors, specialized renderers, examples and research callers; consolidate the tentacle example and include twelve inspected preset images
- keep Matplotlib plots on a white background with standard axes and simple filled base disks, ignoring scene appearance while applying camera, colors, geometry and output settings
- keep both OpenCV backends on a white canvas, ignoring scene backgrounds and ground scenery while preserving robot colors and BGR output
- keep documentation builds independent of native rendering dependencies and make Linux build warning flags portable across Clang versions
- document the API and default changes relative to `main` in the changelog

## Breaking changes

- renderer constructors take appearance, camera, colors, geometry and output defaults through `config=RendererConfig(...)`; removed constructor keywords must be migrated
- direct camera/color configuration imports move into `soromox.rendering.config`; `VideoEncodingConfig` belongs to `config.output`. Public types remain available from `soromox.rendering`
- defaults become 800 × 600 pixels, 80 backbone samples and 48 cross-section samples; automatic camera distance uses a factor of 1.5 instead of 10.0, with a world-aligned technical grid (+Z spatial, +Y planar); base-aligned filled planes must be selected explicitly
- Open3D and Viser use flared collar mounts by default; select `config.geometry.base_plate_style="disk"` for a cylindrical mount
- per-call camera, color and video overrides replace complete sections; OpenCV sequence options are keyword-only and require `record_path`
- Open3D recording exports synchronously without interactive playback; `record_every_n` reduces frame count and FPS, and viewer controls no longer affect exports

## Migration

Before:

```python
renderer = Open3DRenderer(
    robot, width=1920, height=1200, num_points=80,
    cross_section_resolution=20, show_ground_plane=True,
    ground_plane_size=1.2,
)
```

After:

```python
from soromox.rendering import (
    GeometryConfig, GroundPlaneConfig, Open3DRenderer,
    RendererConfig, RenderOutputConfig, SceneConfig,
)

renderer = Open3DRenderer(
    robot,
    config=RendererConfig(
        scene=SceneConfig.technical(
            ground=GroundPlaneConfig(
                visible=True, surface=True, alignment="base", size=1.2,
            ),
        ),
        geometry=GeometryConfig(num_points=80, cross_section_resolution=20),
        output=RenderOutputConfig(width=1920, height=1200),
    ),
)
```

This preserves the explicit dimensions, sampling and base alignment; the modern lighting can change the appearance. The [shared configuration documentation](https://github.com/tud-phi/soromox/blob/codex/open3d-studio-soft-robots/docs/api/rendering/configuration.md) describes the current configuration API.

## Upstream follow-up

- [Open3D #7553](https://github.com/isl-org/Open3D/issues/7553): the official unpatched macOS development wheel reproduces the modern-to-legacy GUI segmentation fault; the legacy-only control succeeds. The issue includes the script and full error log.
- [Validation feedback on Open3D PR #7550](https://github.com/isl-org/Open3D/pull/7550#issuecomment-5595190113): the official unpatched wheel reproduces Metal's unsupported RGB readback abort, with the minimal script and full log.
- [Open3D #7557](https://github.com/isl-org/Open3D/issues/7557): source-level tone-mapper selection regression. Its macOS minimal example encounters the Metal readback abort before returning RGB means; Linux comparison is pending.
- [Open3D #7554](https://github.com/isl-org/Open3D/issues/7554), [#7555](https://github.com/isl-org/Open3D/issues/7555), [#7556](https://github.com/isl-org/Open3D/issues/7556), and [Filament #10397](https://github.com/google/filament/issues/10397): source findings concerning surfaceless contexts, archive byproducts, static linking and worker-thread EGL API selection. Isolated Ubuntu reproductions with full logs are pending; the reports explicitly identify the required evidence.

## Preset gallery

[Full preset gallery and comparison notes](https://github.com/tud-phi/soromox/blob/codex/open3d-studio-soft-robots/docs/api/rendering/presets.md). Each pair uses identical tentacle geometry, placement and camera. Studio and clay scenes use a low, eased backdrop (`radius=0.40`, `vertical_radius=0.23`, `wall_offset=0.02`, `curvature_easing=0.80`). Bright and dark have their own lighting; clay uses neutral studio lighting with an extra cool rim light. Neutral studio is also used in the refreshed core renderer documentation gallery; the renderer-wide default remains technical.

| Preset | Open3D | Viser |
| --- | --- | --- |
| Technical | open3d technical | viser technical |
| Neutral | open3d neutral | viser neutral |
| Bright | open3d bright | viser bright |
| Dark | open3d dark | viser dark |
| Flat | open3d flat | viser flat |
| Clay | open3d clay | viser clay |

### Core renderer gallery

Open3D and Viser use neutral studio settings. Matplotlib uses white plotting axes and a simple filled disk; OpenCV shows the planar counterpart on a white canvas, ignoring scene appearance.

| Open3D | Viser |
| --- | --- |
| Open3D overview | Viser overview |

| Matplotlib | OpenCV |
| --- | --- |
| Matplotlib overview | OpenCV overview |

## Validation

- 248 rendering tests passed on macOS/Python 3.12, with two platform-gated skips. An additional 51 focused configuration and UMArm example tests passed after rebasing onto current main.
- Regression coverage includes mount dimensions/orientation and backend registration, continuous versus discontinuous link contours, Linux display-failure diagnostics, configured Matplotlib video dimensions at 72, 100 and 300 global DPI, migrated example constructors, Open3D light conversion and exposure, and native actuator materials.
- The complete tendon GVS example exported 201 frames at 800 × 600 and 100 FPS. Native static viewing, image/video export and a fresh legacy viewer were exercised; the legacy camera matched the modern camera position.
- Twelve Open3D/Viser preset images and all four core renderer images were refreshed and visually inspected. The six updated bright/dark/clay captures are pixel-identical to the approved trials; regression tests cover the shared backdrop, retained lighting and Matplotlib disk updates. Eight additional captures compare the four mount shapes across backends. Native Metal checks verified neutral grey grading and distinct linear/ACES output.
- A clean Zensical documentation build, focused Ruff checks and `git diff --check` passed.
- Previous CI checks passed on `179c0c99` (the new commit awaits CI): Python 3.11–3.14, minimum numerical dependencies, native Ubuntu rendering and the documentation build ([test run](https://github.com/tud-phi/soromox/actions/runs/34315903897), [docs run](https://github.com/tud-phi/soromox/actions/runs/34315903993)). Documentation deployment is skipped for pull requests.


examples/simulation/pendulum/simulate_pendulum.py
```diff
@@ -156,10 +161,11 @@
     output_path.parent.mkdir(parents=True, exist_ok=True)
     renderer = OpenCVPlanarRenderer(
         robot,
-        width=video_width,
-        height=video_height,
         backbone_color=(0, 0, 0),
         length_scale=2.5,
+        config=RendererConfig(
```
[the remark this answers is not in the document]
mstoelzle: Fixed in 2431adbb: the pendulum example now supplies its black backbone color through RendererConfig.colors. Added a regression test that executes the actual example renderer constructor and verifies the resolved black colors. I also checked 37 example and research renderer constructor calls for removed keyword arguments; none remain. The combined focused suite passes with 326 tests.

src/soromox/rendering/base.py
```diff
@@ -141,6 +142,147 @@ def __init__(
...
+            else np.asarray(base_axes)
+        )
+        normals = normals / np.linalg.norm(normals, axis=-1, keepdims=True)
+        centers = np.pad(curves[:, 0], ((0, 0), (0, 3 - curves.shape[-1])))
+        centers = (
+            centers + (cfg.height - self.config.geometry.base_plate_thickness) * normals
+        )
+        return list(zip(centers, normals, np.full(count, size)))
+
     def _resolve_ground_plane_size(self, *minimum_sizes: float) -> float:
```
Johannap1: I think this is dead code now

mstoelzle: Confirmed: `_resolve_ground_plane_size()` had no production callers; Open3D, Viser and OpenCV use `_resolve_ground_planes()`. Removed the unused helper, cached `ground_plane_size` attribute and redundant validation in 452bf472. GroundPlaneConfig still validates the configured size. Replaced the obsolete helper assertions with tests for the active path: explicit size, scene-scaled automatic size and the minimum size. Validation: 197 rendering tests passed (2 platform-gated skips), clean documentation build, Ruff and git diff --check passed.

src/soromox/rendering/open3d_renderer.py
```diff
@@ -1595,6 +1999,201 @@ def show(
...
+
+        Returns:
+            Tuple of the GUI application, window and scene widget. The caller
+            runs the event loop and owns window cleanup. Call on the main thread.
+        """
+        self._require_modern_capture_support()
+        gui = o3d.visualization.gui
+        app = gui.Application.instance
+        app.initialize()
+        window = app.create_window("Robot (Open3D)", self.width, self.height)
```
Johannap1, line 2193: Unlike previous implementation this does not check for failure before being used so when no usable window surface is available (e.g. GLFW can't create a Wayland window), this segfaults instead of raising.

mstoelzle: Thanks, this is a real gap. Upstream's GLFWWindowSystem::CreateOSWindow() calls glfwSetWindowUserPointer() immediately after glfwCreateWindow(), so a null window can abort inside app.create_window(), before Python can check its result.

403eec3c adds a Linux GUI probe in a separate process before initializing the parent GUI. It exercises native window creation, renderer access and one event-loop tick; a nonzero exit, signal or timeout raises RuntimeError with the captured native stdout/stderr. A returned None is also checked before accessing window.renderer. The probe runs for each show() with the current display environment. It adds startup time and can briefly show a test window; this tradeoff is documented. As with any preflight check, a display failure after the successful probe can still occur in the native viewer.

Regression coverage checks failed exits, SIGABRT/SIGSEGV return codes, timeout logs, platform selection and a missing window. An Ubuntu native test now uses an invalid display connection and requires a Python exception containing the native log, alongside the existing successful Xvfb GUI test.

Local validation: 205 rendering tests passed, 2 platform-gated skips; Ruff, the documentation build and git diff --check passed. The real probe creation/cleanup path passed on macOS. The Linux display failure test is queued in CI; this workstation does not reproduce the Wayland failure itself.

docs/assets/rendering/presets/open3d_bright.png
Johannap1, line 1: Open3D shows a seam Viser doesn't. Viser merges vertices at link-segment boundaries before upload, so normals stay smooth across the joint. Open3D never merges its per-segment meshes, so the same boundary renders as a visible seam (compare open3d_bright.png vs viser_bright.png)

Johannap1 (approved):
small comments, overall nice update

docs/assets/rendering/presets/open3d_bright.png
mstoelzle: Fixed in 34b84928. Modern Open3D now shares area-weighted side normals at matching adjacent contours, snaps the tiny float32 link-boundary sampling offset to a common ring, and removes the buried caps at that join. Real shape/size discontinuities retain their separate contours and caps. Material colors remain independent. This applies to static show(), image export and modern video export.

Simply merging exact duplicate vertices was insufficient: the link-boundary samples are separated by float32 ULPs, and the buried caps also contributed to the visible ring. The final correction removes the artificial seam in a native bright-studio rerender. All six Open3D gallery images have been refreshed; the Viser images were also refreshed for the new shared mount shapes.

Regression tests check equal normals and coincident rings at a continuous boundary, retained caps at a radius step, and preservation of colors. The full rendering suite passes (220 tests, 2 platform-gated skips), as do Ruff, the documentation build and git diff --check. The updated bright rendering is in docs/assets/rendering/presets/open3d_bright.png.

src/soromox/rendering/open3d_renderer.py
mstoelzle: The native Ubuntu job passed on 403eec3c, including the invalid-display test with captured native logs and the successful Xvfb viewer checks: https://github.com/tud-phi/soromox/actions/runs/34367270602/job/102519138048.

PR #211, opened 2026-09-09
Add UMArm operational-space impedance tracking example

mstoelzle:
## Summary
- add a position-only UMArm operational-space example using OperationalSpaceImpedanceControlTracker
- expose a square 12-by-12 virtual actuation matrix through balanced antagonistic pressure inputs around a nominal preload
- track a base-parallel circle with a two-second period and a diameter comparable to the robot length
- provide tracking plots and specialized UMArm Viser rendering with the reference circle and target position
- rename both UMArm control examples to identify their controllers and add a concise changelog entry

## Validation
- 17 focused tests passed
- Ruff formatting and lint checks passed
- both renamed CLI entry points load successfully
- git diff --check passed

### DIFF
--- a/docs/development/changelog.md
+++ b/docs/development/changelog.md
@@ -13,8 +13,9 @@ and include benchmark baseline and measurement context for performance claims.
 
 ### Added
 
-- Added position-only circular operational-space impedance tracking for the
-  McKibben-actuated UMArm using balanced antagonistic pressure inputs.
+- Added position-only, base-parallel circular operational-space impedance
+  tracking for the McKibben-actuated UMArm using balanced antagonistic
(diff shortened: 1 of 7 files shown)


zuorunze (approved):
Reviewed the pressure-coordinate reduction, nominal preload compensation, circle reference, initialization, pressure bounds, and renamed entry points. All 17 focused UMArm tests pass locally on Python 3.13, and Ruff lint/format plus diff whitespace checks pass. No blocking issues found. The rollout is an exact-model demonstration initialized with the desired task velocity; it should not be interpreted as evidence of hardware tracking or global stability.

Runze

